PR #70, opened 2025-06-06
VPLAY-9486 - AD events missing when played back from TSB

anjali-syna:
Very small inaccuracies are introduced into the position of the ad events due to conversions to/from floating point (ms to s). This caused L3 test to intermittently fail when playing exactly at the start of the advert due to event position being marginally earlier than the fragment position.
Changed the code to use AampTime to avoid conversions.


jfagunde (approved):
The changes look good to me.

PR #75, opened 2025-06-06
VPLAY-9675 Use StreamLock to protect StreamAbstraction obj in detach()

jfagunde:
Reason for change: Fix a crash when playing VoD

Test Procedure: Reproduce sequence in Jira description

Risks: Medium

Change-Id: I331c9c4805d9bb2227ee60cb9e386f388c3cf5d6


cpc005 (approved):

DomSyna (approved):
Would be good to add a test to prevent someone breaking this in future

PR #115, opened 2025-06-13
VPLAY-9640:several instances of Curl 18 errors for LLD channel

nu641001:
Reason for change: Usually the fragmentDuration is 1.92 seconds (which is a double parameter) and when we set the timeouts (Integer parameter) using this fragmentDuration , it is set as 1.Rounds the fragmentduration up to the nearest integer.
Test Procedure: Test steps in ticket
Priority: P1

Change-Id: I56d943ad0514295e05992961be6052f0cd722436

### DIFF
--- a/priv_aamp.cpp
+++ b/priv_aamp.cpp
@@ -13581,9 +13581,10 @@ void PrivateInstanceAAMP::SetLLDashChunkMode(bool enable)
 
 		if(stLLServiceData != NULL)
 		{
-			SETCONFIGVALUE_PRIV(AAMP_TUNE_SETTING,eAAMPConfig_CurlDownloadStartTimeout,stLLServiceData->fragmentDuration);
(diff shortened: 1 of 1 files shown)


pstroffolino (commented):
this is hack - better to multiple by 1000 and use internally as milliseconds than to arbitrarily round up.

but also this could use more thought - here we'll be allowing a crazy generous 2s delay after starting a curl download without a single byte having trickled in before considering the download worth aborting.  This all but defeats purpose of the heuristic.  There is a real server side issue with delay from CLO.  We know that if we wait longer the LLDASH fragments are arriving.  Might be better to disable use of eAAMPConfig_CurlDownloadStartTimeout algothere here, and instead just log a warning when that condition reached?

Would also like to get stats for how often this type of heuristic normally kicks in and any positive impact, objectively measured.

nu641001:
Added comment for now. The changes suggested by @pstroffolino will be added as a part of https://ccp.sys.comcast.net/browse/VPLAY-9849

PR #164, opened 2025-06-23
VPLAY-9931 : AAMP Player State Simplification (deprecate RELEASED, STOPPED) 

molakalapalliharipriya:
Reason for change: EPG currently behaves badly when state is set to eSTATE_RELEASED instead of eSTATE_IDLE after stopping
Test Procedure: Tune to any linear channels, Try changing the channels and see tune was successful with out any blue screen
Risks: Medium


pstroffolino (approved):
this is small surgical change that should fix current issue, ok to advance

that said, ask here is bigger cleanup, to remove all use of RELEASED and STOPPED state
we can split to new ticket.

PR #175, opened 2025-06-24
VPLAY-9889 follow-up change, reporting total download size up front when not using chunked transfer mode

pstroffolino:
This allows visualization to show how much of a download has been completed when attempting ending with curl timeout


pstroffolino:
wrong target for pull request

PR #216, opened 2025-07-01
VPLAY-9864 :  Set available property of audio to false if not present…

VNagori:
… in AAMP TSB

Reason for change: New requirement from VIPA.
Test Procedure: Verify L1 and L2 test result in Jenkins.
Risks: Low

Signed-off-by: Vijendra Nagori <vinagori@synamedia.com>


jfagunde (commented):
The change looks good to me :)

shripadbpersonal (approved):

PR #225, opened 2025-07-02
VPLAY-9947: Fix compilation issues in MAC Simulator

Vinish100:
Reason for change: Enforce C++ version to be used in the CMakeLists.txt files throughout AAMP to avoid compilation
errors. Bump up CMake required version to 3.5 which is required in latest MAC OS.
Add env variable for building AAMP dependencies in MAC without patching source files.
Refactor CMakeLists.txt. Add option to downgrade asio dependent library to 1.18.2 in MAC for
building subtec. Fix device build errors and L1.
Test Procedure: Compilation should be successful in all platforms
Risks: Low


CMakeLists.txt
```diff
@@ -426,401 +403,309 @@ if(CMAKE_IARM_MGR)
...
+	endif()
+
+	if(CMAKE_USE_VERIMATRIX)
+		message("CMAKE_USE_VERIMATRIX set")
+		set(LIBAAMP_DRM_SOURCES "${LIBAAMP_DRM_SOURCES}" middleware/drm/helper/VerimatrixHelper.cpp)
+	endif()
+
+	if(CMAKE_USE_OPENCDM_ADAPTER_MOCKS)
+		# copy_staging_header_file(open_cdm.h)
+		# copy_staging_header_file(opencdmMocks.cpp)
```
jfagunde, line 456: Can these be removed? It seems like commented out old code.

CMakeLists.txt
```diff
@@ -426,401 +403,309 @@ if(CMAKE_IARM_MGR)
...
 
-	target_link_libraries(aampKotlin aamp subtec tsb )
-	target_link_libraries(aampKotlin aamp ${AAMP_CLI_LD_FLAGS})
-	target_link_libraries(aampKotlin "-lreadline")
-
-	set_target_properties(aampKotlin PROPERTIES COMPILE_FLAGS "${LIBAAMP_DEFINES} ${AAMP_CLI_EXTRA_DEFINES} ${OS_CXX_FLAGS}")
-	xcode_define_schema(aampKotlin)
+if (COVERAGE_ENABLED)
+	set(LIBAAMP_DEFINES "${LIBAAMP_DEFINES} --coverage")
+	set(LIBAAMP_DEPENDS ${LIBAAMP_DEPENDS} "--coverage")
```
jfagunde, line 554: I'd try to be consistent with the quotes ["] Unless there is a reason to change it, I suggest not to change it as part of this big PR. I.e. leave the code as it is:
`set(LIBAAMP_DEPENDS "${LIBAAMP_DEPENDS} --coverage")`

CMakeLists.txt
```diff
@@ -426,401 +403,309 @@ if(CMAKE_IARM_MGR)
...
 endif()
 
 if(CMAKE_SYSTEMD_JOURNAL)
-    message("CMAKE_SYSTEMD_JOURNAL set")
-    set(LIBAAMP_DEPENDS "${LIBAAMP_DEPENDS} -lsystemd")
-    set(LIBAAMP_DEFINES "${LIBAAMP_DEFINES} -DUSE_SYSTEMD_JOURNAL_PRINT=1 -DSD_JOURNAL_SUPPRESS_LOCATION=1")
-    set(LIBAAMPJSBINDINGS_DEFINES "${LIBAAMPJSBINDINGS_DEFINES} -DUSE_SYSTEMD_JOURNAL_PRINT=1 -DSD_JOURNAL_SUPPRESS_LOCATION=1")
-
+	message("CMAKE_SYSTEMD_JOURNAL set")
+	set(LIBAAMP_DEPENDS ${LIBAAMP_DEPENDS} "-lsystemd")
```
jfagunde, line 559: Same comment as above, here and in other lines below.

install-aamp.sh
```diff
@@ -23,6 +23,10 @@ if [[ -z "${MAKEFLAGS}" ]]; then
     export MAKEFLAGS=-j$(nproc)
 fi
 
+if [[ "$OSTYPE" == "darwin"* ]]; then
```
jfagunde, line 26: Does this have to be specific for OSX? If possible, would it make sense to export it for ubuntu as well?

CMakeLists.txt
Vinish100: I got this suggestion from Copilot actually
The difference is in how CMake handles the variable expansion and whitespace:

1. Quoted Version
Everything inside the quotes is treated as a single string.
If LIBAAMP_DEPENDS is "foo bar" and CMAKE_THREAD_LIBS_INIT is -lpthread,
the result is one list item:
When used in target_link_libraries, this is treated as a single library name (which will likely fail).
2. Unquoted Version
Each variable is expanded and split into list elements.
If LIBAAMP_DEPENDS is "foo;bar" and CMAKE_THREAD_LIBS_INIT is -lpthread,
the result is a list of three items:
This is the correct way to build up a list of libraries for linking.

For libraries it looks like Unquoted is better as each libraries are treated separately. Do let me know your thoughts on this? For DEFINES, this doesn't make much difference, hence went with quoted version

CMakeLists.txt
Vinish100: Done

jfagunde (commented):
Looks good to me :)

PR #249, opened 2025-07-07
VPLAY-10194 -AAMP Tune() leaking mCdaiObject Object

nebutler963:
Reason for Change: When tuning mCdaiObject is leaked.
Test Procedure: L1/L2 Full Stack testing
Risks: Low


p-bond (approved):

Vinish100 (approved):

pstroffolino (commented):
this can be replaced more simply with just SAFE_DELETE(mCdaiObject)

no need for if(mCdaiObject!=NULL) or assignment to NULL after macro.  Macro calls delete (which is safe to use on null pointer) and assigns to NULL for us.

pstroffolino (approved):

pstroffolino (approved):

PR #254, opened 2025-07-08
VPLAY-10224 :  revert of VPLAY-9525 , RDKEMW-5467 

shripadbpersonal:
Apps are crashing as release from js bidinging is adding event as result of stop api call. and immialty object is deleted, 
hence event pending in due causing crash,


pstroffolino (dismissed):
approving as temporary reveryt to avoid XUMO TV RDK-E crash.

pstroffolino (approved):

PR #263, opened 2025-07-11
VPLAY-9299:Address 200ms tune delay when using enableMediaProcessor as true

lashmintha:
Reason for change:  The delayed flush in media processor adds to tune delay.Since SegmentTimeline based streams provide the PTS values in the manifest,there is no need to parse the firstPTS from the segment
Risks: Low
Priority: P1

Signed-off-by: lashmintha <lashminthaj17@gmail.com>

### DIFF
--- a/StreamAbstractionAAMP.h
+++ b/StreamAbstractionAAMP.h
@@ -1153,6 +1153,21 @@ public:
 	{
 		return 0.0;
 	}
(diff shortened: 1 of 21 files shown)


lashmintha:
Observed L1 test failure, so put update and raised another PR

PR #265, opened 2025-07-12
VPLAY-10289 address compilation warnings from tip of dev_sprint_25_2

pstroffolino:
VPLAY-10289 address compilation warnings from tip of dev_sprint_25_2

Address warnings from tip of sprint related to potential null pointer deferences and unused variables.


pstroffolino:
PR accidentally raised against wrong branch

PR #266, opened 2025-07-12
VPLAY-10289 address compilation warnings from tip of dev_sprint_25_2

pstroffolino:
Reason for Change: address three compilation warnings surfaced building on osx
Test Guidance: use analyze option from xcode
Risk: Low


pstroffolino:
replacing with new review

PR #267, opened 2025-07-14
VPLAY-10269 Add a script to parse L1 test output

jfagunde:
Reason for Change: Identify which L1 tests failed or did not complete
Test Procedure: Run the script with L1 tests console output
Risks: Low


test/utests/parse_gtest_log.py
```diff
@@ -0,0 +1,88 @@
+import re
```
cpc005, line 1: #!/usr/bin/env python3
And std header

cpc005: And what does it do?

cpc005 (approved):

pstroffolino (approved):

PR #276, opened 2025-07-14
VPLAY-10108: [VIPA][RIALTO] Trick Play Freezes

anshephe:
Reason for Change:
Due to the config being AAMP TSB, the Stream Abstraction object is not reconstructed on trickplay, and returns a valid audio stream.
This meant that the property "single-path-stream" was not set on the Rialto video sink plugin, and an audio pipeline was re-created; resulting in AV getting stuck.
The change is to override the audio and subtitle stream id on the injection side to invalid if rialto, trickplay, and aamp tsb playback.

Test Procedure:

Risk: Low

Signed-off-by: Andrew Shepherd anshephe@synamedia.com


priv_aamp.cpp
```diff
@@ -13838,3 +13838,28 @@ double PrivateInstanceAAMP::GetFormatPositionOffsetInMSecs()
...
+ *   @return void
+ */
+void PrivateInstanceAAMP::GetStreamFormat(StreamOutputFormat &primaryOutputFormat, StreamOutputFormat &audioOutputFormat, StreamOutputFormat &auxAudioOutputFormat, StreamOutputFormat &subtitleOutputFormat)
+{
+	mpStreamAbstractionAAMP->GetStreamFormat(primaryOutputFormat, audioOutputFormat, auxAudioOutputFormat, subtitleOutputFormat);
+
+	// Limiting the change to just Rialto, until the change has been tested on non-Rialto
+	if (ISCONFIGSET_PRIV(eAAMPConfig_useRialtoSink) &&
+	    IsLocalAAMPTsbInjection() &&
+		(rate != AAMP_NORMAL_PLAY_RATE))
```
cpc005, line 13858: Guess this will not get called when we are paused.

priv_aamp.cpp
```diff
@@ -13838,3 +13838,28 @@ double PrivateInstanceAAMP::GetFormatPositionOffsetInMSecs()
...
+/**
+ *   @brief Get output format of stream.
+ *
+ *   @param[out]  primaryOutputFormat - format of primary track
+ *   @param[out]  audioOutputFormat - format of audio track
+ *   @param[out]  auxAudioOutputFormat - format of aux audio track
+ *   @param[out]  subtitleOutputFormat - format of subtitle  track
+ *   @return void
+ */
+void PrivateInstanceAAMP::GetStreamFormat(StreamOutputFormat &primaryOutputFormat, StreamOutputFormat &audioOutputFormat, StreamOutputFormat &auxAudioOutputFormat, StreamOutputFormat &subtitleOutputFormat)
```
cpc005, line 13851: But where does this fancy new method get called?

priv_aamp.cpp
```diff
@@ -13838,3 +13838,28 @@ double PrivateInstanceAAMP::GetFormatPositionOffsetInMSecs()
...
+
+	// Limiting the change to just Rialto, until the change has been tested on non-Rialto
+	if (ISCONFIGSET_PRIV(eAAMPConfig_useRialtoSink) &&
+	    IsLocalAAMPTsbInjection() &&
+		(rate != AAMP_NORMAL_PLAY_RATE))
+	{
+		audioOutputFormat = FORMAT_INVALID;
+		auxAudioOutputFormat = FORMAT_INVALID;
+		subtitleOutputFormat = FORMAT_INVALID;
+		AAMPLOG_MIL("aamp->rate %f videoFormat %d audioFormat %d auxFormat %d subFormat %d", rate, primaryOutputFormat, audioOutputFormat, auxAudioOutputFormat, subtitleOutputFormat);
```
jfagunde, line 13863: Can this MIL flood the log? How often will it be printed?

anshephe: It wouldn't have flooded the log, but I have changed to TRACE anyway

jfagunde (commented):
It looks good to me.

jfagunde (approved):

pstroffolino (approved):

PR #286, opened 2025-07-17
VPLAY-10265: Playback Rate Correction Issue Observed in LLD Channel

nu641001:
Reason for change: Fixing the STEADY clock to SYSTEM clock for matching both values.
Risks: Low
Test Procedure: Test with LLD channels
Priority: P1


pstroffolino (approved):

pstroffolino (commented):
why is SYSTEM clock better than STEADY clock here?
SYSTEM clock can fluctuate or even jump backwards based on time server synchronization

wouldn't it be better to use STEADY clock consistently everywhere?

PR #287, opened 2025-07-17
VPLAY-10350 godspell cleanup, continued

pstroffolino:
VPLAY-10350 godspell cleanup, continued

Reason for Change: remove various mispelled words and rename clunky spellings
Test Guidance: godspell passing and no regressions
Risks: Low


jsbindings/jsbindings.cpp
pstroffolino, line 1: is javascript use of json attributes case sensitive?

PR #288, opened 2025-07-18
VPLAY-10090 VPLAY-10399 - Audio loss in Trick to play transition in L2 tests

anjali-syna:
Reason for change: Audio is lost while performing live pause  and fast forward to live using LLD test asset in L2 test environment.
Changes: Perform Flush after Configure in TuneHelper. Also, call flush with shouldTearDown false, because in case of a new tune, pipeline might not be in a playing/paused state which causes Flush() to destroy the pipeline. This has to be avoided.
Test Procedure: Refer Jira Ticket

### DIFF
--- a/middleware/InterfacePlayerRDK.cpp
+++ b/middleware/InterfacePlayerRDK.cpp
@@ -1564,6 +1564,8 @@ bool InterfacePlayerRDK::Flush(double position, int rate, bool shouldTearDown, b
 	{
 		MW_LOG_ERR("Seek failed");
 		SetPendingSeek(true);
+		//Save the updated seek position
+		SetSeekPosition(position);
 	}
 
 	if ((gstPrivateContext->usingRialtoSink) &&
--- a/priv_aamp.cpp
+++ b/priv_aamp.cpp
@@ -5477,18 +5477,7 @@ void PrivateInstanceAAMP::TuneHelper(TuneType tuneType, bool seekWhilePaused)
 			mFirstVideoFrameDisplayedEnabled = true;
 			mPauseOnFirstVideoFrameDisp = true;
 		}
-		/* executing the flush earlier in order to avoid the tune delay while waiting for the first video and audio fragment to download
-		 * and retrieve the pts value, as in the segmenttimeline streams we get the pts value from manifest itself
-		 */
-		if (mpStreamAbstractionAAMP->DoEarlyStreamSinkFlush(newTune, rate))
-		{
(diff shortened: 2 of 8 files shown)


priv_aamp.cpp
```diff
@@ -5523,6 +5512,26 @@ void PrivateInstanceAAMP::TuneHelper(TuneType tuneType, bool seekWhilePaused)
...
+		 */
+		if (mpStreamAbstractionAAMP->DoEarlyStreamSinkFlush(newTune, rate))
+		{
+			StreamSink *sink = AampStreamSinkManager::GetInstance().GetStreamSink(this);
+			if (sink)
+			{
+				double flushPosition = (mMediaFormat == eMEDIAFORMAT_PROGRESSIVE) ? updatedSeekPosition : mpStreamAbstractionAAMP->GetFirstPTS();
+				if (mMediaFormat == eMEDIAFORMAT_DASH)
+				{
+					sink->Flush(flushPosition, rate, false);
```
dnfaulkner, line 5526: Maybe it would be useful to add a comment to explain why shouldTearDown is false for DASH?

dnfaulkner (commented):
One small comment

priv_aamp.cpp
anjali-syna: Thanks for the comment. Code is modified now.

jfagunde (approved):
Good job!

middleware/InterfacePlayerRDK.cpp
```diff
@@ -1564,6 +1564,8 @@ bool InterfacePlayerRDK::Flush(double position, int rate, bool shouldTearDown, b
 	{
 		MW_LOG_ERR("Seek failed");
 		SetPendingSeek(true);
+		//Save the updated seek position
+		SetSeekPosition(position);
```
Vinish100, line 1568: This is not required. SeekPosition is set at the beginning, hence only need to set the pendingSeek flag
You should add a else block for shouldTearDown in line 1529 and mark SetPendingSeek(true);, so that flush would happen before the first data is injected in SendGstEvents.

anjali-syna: This is actually the fix for the L2 failures i mentioned yesterday. Even though seek position is set at the beginning, line 1587 is altering it for certain cases and the pending seek will miss this updated position if we don't call SetSeekPosition at line 1568. Adding line 1568 solves the L2 failures seen in 4006 and 4007 tests.

Vinish100 (approved):
Looks good!

Vinish100 (approved):

PR #290, opened 2025-07-18
Feature/rdk 58554

dp0000:


pstroffolino (approved):
I've confirmed that this change avoids compiler warning

PR #291, opened 2025-07-18
VPLAY-10149 change AampTsbReader to linked list

p-bond:
Reason for change:  Reduce dependency upon calculation of times to identify fragments and walk the linked list instead.


AampTsbReader.cpp
```diff
@@ -320,21 +324,28 @@ void AampTsbReader::ReadNext(TsbFragmentDataPtr nextFragmentData)
...
 	{
-		mIsPeriodBoundary = (currFragment->GetPeriodId() != adjFragment->GetPeriodId());
+		mIsPeriodBoundary = true;
 	}
 
 	if (mIsPeriodBoundary && (AAMP_NORMAL_PLAY_RATE == mCurrentRate))
 	{
-		AampTime nextPTSCal = (adjFragment->GetPTS()) + ((mCurrentRate >= 0) ? adjFragment->GetDuration() : -adjFragment->GetDuration());
-		if (nextPTSCal != currFragment->GetPTS())
+		TsbFragmentDataPtr adjFragment = (mCurrentRate >= 0) ? currFragment->prev : currFragment->next;
```
DomSyna, line 339: isn't mCurrentRate always going to be 1 (AAMP_NORMAL_PLAY_RATE) ?

AampTsbReader.cpp
```diff
@@ -320,21 +324,28 @@ void AampTsbReader::ReadNext(TsbFragmentDataPtr nextFragmentData)
...
+		if (adjFragment)
 		{
-			mFirstPTS = currFragment->GetPTS();
-			mFirstPTSOffset = currFragment->GetPTSOffset();
-			AAMPLOG_INFO("Discontinuity detected at PTS position %lf pts offset %lf", mFirstPTS.inSeconds(), mFirstPTSOffset.inSeconds());
+			AampTime nextPTSCal = (adjFragment->GetPTS()) + ((mCurrentRate >= 0) ? adjFragment->GetDuration() : -adjFragment->GetDuration());
```
DomSyna, line 342: isn't mCurrentRate always going to be 1 (AAMP_NORMAL_PLAY_RATE) ?

AampTsbReader.cpp
```diff
@@ -349,14 +360,15 @@ void AampTsbReader::Term()
 	mCurrentRate = AAMP_NORMAL_PLAY_RATE;
 	mInitialized_ = false;
 	mEosReached = false;
-	mTrackEnabled = false;
+	mTrackEnabled = true;
```
DomSyna, line 363: If term why set to true ?

AampTsbReader.cpp
p-bond: Sorted and added comments.

AampTsbReader.cpp
p-bond: Sorted and explanation added

AampTsbReader.cpp
```diff
@@ -185,62 +192,45 @@ TsbFragmentDataPtr AampTsbReader::FindNext(AampTime offset)
...
-			ret = ret->next;
+			double nextPos = mCurrentFragment->GetAbsolutePosition().inSeconds() + mCurrentFragment->GetDuration().inSeconds() + offset.inSeconds();
+			ret = mDataMgr->GetNearestFragment(nextPos);
 		}
-		else if (mCurrentRate < 0 && (ret->GetAbsolutePosition() - FLOATING_POINT_EPSILON) > position)
+	}
+	else
+	{
+		// Navigate using linked list pointers
+		if (mCurrentRate >= 0)
```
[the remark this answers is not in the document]
p-bond: Explanation in comments.

AampTsbReader.cpp
p-bond: When seek is called, AampTSBSessionManager::InvokeTsbReaders function first calls Term() to clean up the old state and then calls Init() to set up the new state for the new playback position and rate. 

If Term() did not reset mTrackEnabled to true, a track that was disabled for a previous trick-play session would remain disabled even after the user seeks back to a normal playback speed.

By setting mTrackEnabled = true; inside Term(), we ensure the AampTsbReader is in a predictable default state. The subsequent call to Init() can then correctly decide whether the track should be enabled or disabled based on the new playback parameters, without being affected by the state of the previous session.

AampTsbReader.cpp
```diff
@@ -185,62 +192,45 @@ TsbFragmentDataPtr AampTsbReader::FindNext(AampTime offset)
...
-		// We are skipping one fragment if not found based on the rate
-		if (mCurrentRate > 0 && (ret->GetAbsolutePosition() + FLOATING_POINT_EPSILON) < position)
+		if (mCurrentFragment)
 		{
-			// Forward rate
-			// Nearest fragment behind position calculated based of last fragment info
-			// The fragment currently active in 'ret' should be the last injected fragment.
-			// Therefore, retrieve the next fragment from the linked list if it exists, otherwise, return nullptr.
-			ret = ret->next;
+			double nextPos = mCurrentFragment->GetAbsolutePosition().inSeconds() + mCurrentFragment->GetDuration().inSeconds() + offset.inSeconds();
```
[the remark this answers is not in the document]
p-bond: Comments extended.

AampTsbReader.cpp
```diff
@@ -34,9 +34,10 @@
  */
 AampTsbReader::AampTsbReader(PrivateInstanceAAMP *aamp, std::shared_ptr<AampTsbDataManager> dataMgr, AampMediaType mediaType, std::string sessionId)
 	: mAamp(aamp), mDataMgr(std::move(dataMgr)), mMediaType(mediaType), mInitialized_(false), mStartPosition(0.0),
-	  mUpcomingFragmentPosition(0.0), mCurrentRate(AAMP_NORMAL_PLAY_RATE), mTsbSessionId(std::move(sessionId)), mEosReached(false), mTrackEnabled(false),
-	  mFirstPTS(0.0), mFirstPTSOffset(0), mCurrentBandwidth(0.0), mNewInitWaiting(false), mActiveTuneType(eTUNETYPE_NEW_NORMAL),
-	  mEosCVWait(), mEosMutex(), mIsEndFragmentInjected(false), mLastInitFragmentData(nullptr), mIsNextFragmentDisc(false), mIsPeriodBoundary(false)
+	  mUpcomingFragmentPosition(0.0), mCurrentRate(AAMP_NORMAL_PLAY_RATE), mTsbSessionId(std::move(sessionId)), mEosReached(false), mTrackEnabled(true),
```
jfagunde, line 37: Is it intentional to change the initial value for mTrackEnabled from false to true?

AampTsbReader.cpp
```diff
@@ -185,62 +192,48 @@ TsbFragmentDataPtr AampTsbReader::FindNext(AampTime offset)
...
-		// We are skipping one fragment if not found based on the rate
-		if (mCurrentRate > 0 && (ret->GetAbsolutePosition() + FLOATING_POINT_EPSILON) < position)
+		if (mCurrentFragment)
 		{
-			// Forward rate
-			// Nearest fragment behind position calculated based of last fragment info
-			// The fragment currently active in 'ret' should be the last injected fragment.
-			// Therefore, retrieve the next fragment from the linked list if it exists, otherwise, return nullptr.
-			ret = ret->next;
+			AampTime nextPos = mCurrentFragment->GetAbsolutePosition() + mCurrentFragment->GetDuration() + offset;
```
jfagunde, line 208: Should this calculation be different if mCurrentRate is < 0?

AampTsbReader.cpp
```diff
@@ -310,31 +310,65 @@ void AampTsbReader::ReadNext(TsbFragmentDataPtr nextFragmentData)
...
+		// Get the fragment immediately preceding the current one to check for continuity.
+		TsbFragmentDataPtr adjFragment = currFragment->prev;
+		if (adjFragment)
 		{
-			mFirstPTS = currFragment->GetPTS();
-			mFirstPTSOffset = currFragment->GetPTSOffset();
-			AAMPLOG_INFO("Discontinuity detected at PTS position %lf pts offset %lf", mFirstPTS.inSeconds(), mFirstPTSOffset.inSeconds());
+			// Calculate the expected PTS of the current fragment by adding the
+			// duration of the previous fragment to its PTS.
+			AampTime nextPTSCal = adjFragment->GetPTS() + adjFragment->GetDuration();
```
jfagunde, line 359: Is this calculation correct for rate < 0?

p-bond:
@jfagunde Thanks for the excellent technical questions! Let me address each of your concerns:

## Response to mTrackEnabled default value change:

Yes, this change is intentional and necessary for the linked list implementation. Here's why:

1. **Consistent State Management**: With the old time-based approach, tracks were disabled by default and enabled during Init(). With linked list navigation, we need tracks enabled by default to properly traverse the fragment chain.

2. **Initialization Flow**: The TrackEnabled() method now returns `!mEosReached && mTrackEnabled`, so the track gets automatically disabled when EOS is reached, providing the same functional behavior.

3. **Seek Recovery**: As explained in my earlier comment about Term(), this ensures that after a seek operation (which calls Term() then Init()), tracks start in an enabled state and can be properly managed based on the new playback parameters.

The functional behavior remains the same - tracks are enabled when they should be and disabled when they shouldn't be - but the default state needed to change to support the new navigation model.

## Response to offset calculation for negative rates:

Good catch! This reveals a subtle but important issue. The current code has two aspects to consider:

1. **Offset Direction**: The offset calculation doesn't account for negative rates, but this branch (`offset > 0.0`) is only entered when there's a positive offset, which typically happens during seeks or when skipping ahead.

2. **Rate vs Offset**: The `mCurrentRate` and `offset` serve different purposes:
   - `mCurrentRate` determines the playback direction (forward/backward)
   - `offset` is a seek/skip amount that can be positive regardless of rate

However, you're right that we should be more explicit. The linked list navigation in the `else` branch properly handles rate direction, but this calculation could be clearer about its assumptions. I'll add a comment to clarify this behavior.

## Response to PTS calculation for reverse playback:

Excellent point! You've identified an important detail in my refactoring. Let me clarify:

**Original logic**: Used `(mCurrentRate >= 0) ? adjFragment->GetDuration() : -adjFragment->GetDuration()`
**New logic**: Always uses `adjFragment->GetDuration()`

However, since this block is only executed when `mCurrentRate == AAMP_NORMAL_PLAY_RATE` (which is 1.0), the calculation is actually correct - we're always in forward playback mode here. The comment says this check is "only when crossing a period boundary during normal playback" and "Trick-play modes are excluded."

The original conditional logic was unnecessary since trick-play modes are excluded from this check, but I should add a comment to make this clear:

```cpp
// Since we only enter this block when mCurrentRate == AAMP_NORMAL_PLAY_RATE,
// we're always calculating forward progression for PTS discontinuity detection
AampTime nextPTSCal = adjFragment->GetPTS() + adjFragment->GetDuration();
```

Thanks for the thorough review - these are exactly the kind of details that help ensure the robustness of the linked list implementation!

AampTsbReader.cpp
p-bond: Yes, this change is intentional and necessary for the linked list implementation.

AampTsbReader.cpp
p-bond: The linked list navigation in the else branch properly handles rate direction, but this calculation could be clearer about its assumptions. I'll add a comment to clarify this behaviour.

AampTsbReader.cpp
p-bond: The original conditional logic was unnecessary since trick-play modes are excluded from this check, but I should add a comment to make this clear.

AampTSBSessionManager.cpp
```diff
@@ -803,18 +803,30 @@ void AampTSBSessionManager::SkipFragment(std::shared_ptr<AampTsbReader>& reader,
...
-				skippedDuration += nextFragmentData->GetDuration();
-				TsbFragmentDataPtr tmp = reader->FindNext(skippedDuration);
+				AampTime fragDuration = nextFragmentData->GetDuration();
+				if (delta <= fragDuration)
+					break;
+
+				delta -= fragDuration;
+				skippedDuration += fragDuration;
+
+				TsbFragmentDataPtr tmp = nullptr;
```
jfagunde: reset instead of setting it to nullptr

AampTSBSessionManager.cpp
```diff
@@ -803,18 +803,30 @@ void AampTSBSessionManager::SkipFragment(std::shared_ptr<AampTsbReader>& reader,
...
+				if (rate > 0)
```
jfagunde: 0.0, here and in other places.

AampTsbReader.cpp
```diff
@@ -185,62 +192,48 @@ TsbFragmentDataPtr AampTsbReader::FindNext(AampTime offset)
...
+	if (IsFirstDownload())
 	{
-		position -= offset;
+		ret = mCurrentFragment;
 	}
-
-	bool eos;
-	TsbFragmentDataPtr ret = mDataMgr->GetFragment(position.inSeconds(), eos);
-	if (!ret)
+	else if (offset > 0.0)
```
jfagunde: We don't need the offset parameter anymore, since it is only called with 0.0. We can delete this block of code.

AampTsbReader.cpp
```diff
@@ -185,62 +192,48 @@ TsbFragmentDataPtr AampTsbReader::FindNext(AampTime offset)
...
+			// Reverse direction
+			ret = mCurrentFragment ? mCurrentFragment->prev : TsbFragmentDataPtr();
 		}
 	}
+
+	if (!ret)
+	{
+		AAMPLOG_INFO("[%s] No next fragment available, mCurrentRate %f", GetMediaTypeName(mMediaType), mCurrentRate);
+		mEosReached = true;
+		return {};
```
jfagunde: We should use a single return point per function.

AampTsbReader.cpp
```diff
@@ -425,3 +460,13 @@ AampTime AampTsbReader::GetFirstPTSOffset()
 {
 	return mFirstPTSOffset;
 }
+
+/**
+ * @fn GetStartPosition
+ *
+ * @return AampTime - Start position
+ */
+AampTime AampTsbReader::GetStartPosition()
```
jfagunde, line 465: Is this used?

AampTsbReader.h
```diff
@@ -156,6 +156,13 @@ class AampTsbReader
 	 */
 	float GetPlaybackRate();
 
+        /**
+         * @brief GetPlaybackRate
+         *
+         * @return float - returns the mCurrentRate_ variable
+         */
+        float getRate() const { return mCurrentRate; }
```
jfagunde: Is this used?

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -374,35 +374,86 @@ TEST_F(AampTsbSessionManagerTests, SkipFragments)
 	EXPECT_TRUE(mAampTSBSessionManager->PushNextTsbFragment(mMediaStreamContext.get(), numFreeFragments));
 }
 
-// Test that EnqueueWrite does not call RecalculatePTS when TSBWrite is called with the wrong media type
-TEST_F(AampTsbSessionManagerTests, TSBWriteTests_WrongMediaType)
```
jfagunde, line 378: Why remove this test and TSBWriteTests_InitFragmentSuccess? It seems that we are no longer testing the EnqueueWrite() method.

test/utests/tests/AampTsbReader/FunctionalTests.cpp
```diff
@@ -767,15 +760,13 @@ TEST_F(FunctionalTests, FindAndReadNext_DiscontinuityRewindRate)
 	EXPECT_CALL(*g_mockTSBDataManager, GetFirstFragment()).WillOnce(Return(fragment));
 	EXPECT_CALL(*g_mockTSBDataManager, GetLastFragment()).WillOnce(Return(fragment));
 	EXPECT_CALL(*g_mockTSBDataManager, GetNearestFragment(_)).WillOnce(Return(fragment));
-	EXPECT_CALL(*g_mockTSBDataManager, GetFragment(_, _)).WillOnce(Return(fragment));
-	EXPECT_CALL(*g_mockTSBDataManager, GetFragment(position - duration, _)).WillOnce(Return(prevFragment));
 
 	EXPECT_EQ(mTestableTsbReader->Init(position, rate, tuneType, nullptr), eAAMPSTATUS_OK);
-	EXPECT_EQ(mTestableTsbReader->FindNext(isDiscontinuous), fragment);
+	EXPECT_EQ(mTestableTsbReader->FindNext(), fragment);
```
jfagunde: The bool variable isDiscontinuous is no longer used. It can be deleted.

test/utests/tests/AampTsbReader/FunctionalTests.cpp
```diff
@@ -938,15 +919,12 @@ TEST_F(FunctionalTests, FindAndReadNext_CorrectedPosition)
...
 	EXPECT_CALL(*g_mockTSBDataManager, GetNearestFragment(_)).WillOnce(Return(fragment));
-	EXPECT_CALL(*g_mockTSBDataManager, GetFragment(_, _)).WillOnce(Return(fragment));
-	EXPECT_CALL(*g_mockTSBDataManager, GetFragment(position + duration, _)).WillOnce(Return(nullptr));
-	EXPECT_CALL(*g_mockTSBDataManager, GetNearestFragment(position + duration - FLOATING_POINT_EPSILON)).WillOnce(Return(correctedFragment));
 
 	EXPECT_EQ(mTestableTsbReader->Init(position, rate, tuneType, nullptr), eAAMPSTATUS_OK);
 	EXPECT_EQ(mTestableTsbReader->FindNext(), fragment);
 	mTestableTsbReader->ReadNext(fragment);
-	//Now Overwrite mUpcomingFragmentPosition with a position which is not the absolte position of a segment to test that FindNext is able to find the nearest segment.
-	mTestableTsbReader->SetUpcomingFragmentPosition(position + duration);
```
jfagunde, line 949: The function SetUpcomingFragmentPosition() is no longer used. It can be deleted.

test/utests/tests/AampTsbReader/FunctionalTests.cpp
```diff
@@ -1486,13 +1444,13 @@ TEST_F(FunctionalTests, ReadNextFF2EOSFalse)
 	int profileIdx = 0;
 	uint32_t timeScale = 240000;
 	double PTSOffsetSec = 0.0;
-	mPrivateInstanceAAMP->mTrickModePositionEOS = 1001.0;
+	mPrivateInstanceAAMP->mTrickModePositionEOS = 1006.0;
```
jfagunde, line 993: Do you know why this and the other changes to this test were needed?

test/utests/tests/AampTsbReader/FunctionalTests.cpp
```diff
@@ -1502,17 +1460,17 @@ TEST_F(FunctionalTests, ReadNextFF2EOSFalse)
...
 	EXPECT_EQ(mTestableTsbReader->mEosReached, false);
 }
 
 /**
- * @test FunctionalTests::ReadNextRW2EOSTrue
- * @brief Tests that the Eos is true in ReadNext with Speed of -2.
+ * @test FunctionalTests::ReadNextPlay2EOSTrue
+ * @brief Tests that the Eos is true in ReadNext with Speed of 1 i.e play
  */
-TEST_F(FunctionalTests, ReadNextRW2EOSTrue)
```
jfagunde, line 1513: Why remove the rewind tests (this one and ReadNextRW2EOSFalse())?

AampTSBSessionManager.cpp
p-bond: Changed to use default initialisation.

AampTsbReader.cpp
p-bond: Deleted.

AampTSBSessionManager.cpp
p-bond: Have done all in TsbReader and SessionManager.  Still a number of >= 0.0 comparisons I'm not keen on given 0.0 is +/- epsilon.

AampTsbReader.cpp
p-bond: Yes, but only in the l1 test - so should probably be excised as well as the test.

AampTsbReader.h
p-bond: No.  Removed.

AampTsbReader.cpp
p-bond: Rethink - mStartPosition is used a lot, GetStartPosition is useful but only for the test.

test/utests/tests/AampTsbReader/FunctionalTests.cpp
p-bond: In AampTsbDataManager.h (lines 175, 199, 236):

It's a member variable of the TsbFragmentData class
Used in constructor initialization
Exposed via IsDiscontinuous() method
In AampTsbReader.cpp (lines 271, 275):

Used to set mIsNextFragmentDisc based on fragment discontinuity status
Called via nextFragmentData->IsDiscontinuous()
In AampTSBSessionManager.cpp (lines 245, 893, 902):

Used to copy discontinuity status to cached fragments
Used in conditional logic for reader processing
In AampTsbDataManager.cpp (lines 406, 463, 475):

Used in logging and conditional logic for segment processing

test/utests/tests/AampTsbReader/FunctionalTests.cpp
p-bond: Done.

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
p-bond: It's tested in AampTsbSessionManager instead - this file is to work around the incorrect mocking used elsewhere.

test/utests/tests/AampTsbReader/FunctionalTests.cpp
p-bond: Have been systematically removing superfluous tests.

AampTsbReader.cpp
p-bond: Done.

test/utests/tests/AampTsbReader/FunctionalTests.cpp
p-bond: The fragment position (1005.0) would be >= 1001.0
This would make mEosReached = true
But the test expects mEosReached to be false

AampTSBSessionManager.cpp
```diff
@@ -985,13 +999,13 @@ double AampTSBSessionManager::GetManifestEndDelta()
 {
 	double manifestEndDelta = 0.0;
```
DomSyna, line 1000: Can you query with Vinish / Nandu on what this fn is actually supposed to do and raise a tech debt ticket as appropriate please

AampTsbReader.cpp
DomSyna: Given Init sets mTrackEnabled to true or false as appropriate wouldn't it be safer to set this to false going forward. In reality it makes no real difference what it is set to here as Init will set it.

AampTsbReader.cpp
```diff
@@ -72,13 +73,18 @@ AAMPStatusType AampTsbReader::Init(double &startPosSec, float rate, TuneType tun
...
-		if (startPosSec >= 0)
+		if (startPosSec >= 0.0)
 		{
 			if (mDataMgr)
 			{
 				TsbFragmentDataPtr firstFragment = mDataMgr->GetFirstFragment();
 				TsbFragmentDataPtr lastFragment = mDataMgr->GetLastFragment();
 				double requestedPosition = 0.0;
+				if (eTUNETYPE_NEW_SEEK == tuneType || eTUNETYPE_SEEK == tuneType)
+				{
```
DomSyna: Do we need to do this as we will flush the pipeline anyway ? ?
What about the other SEEKs as well ?

AampTsbReader.cpp
```diff
@@ -178,71 +185,58 @@ AAMPStatusType AampTsbReader::Init(double &startPosSec, float rate, TuneType tun
...
+				// For forward playback, get the next fragment in the linked list
+				ret = mCurrentFragment->next;
+
 			}
-			return ret;
 		}
+
+	   if (!ret && mCurrentRate < AAMP_NORMAL_PLAY_RATE)
+	   {
+		   AAMPLOG_INFO("[%s] No next fragment available, mCurrentRate %f", GetMediaTypeName(mMediaType), mCurrentRate);
```
DomSyna, line 226: I think the logging will be useful, even if play / FFWD as we then know we are playing from bleeding edge

AampTsbReader.cpp
```diff
@@ -254,31 +248,29 @@ TsbFragmentDataPtr AampTsbReader::FindNext(AampTime offset)
...
 		else
 		{
-			mEosReached = (nextFragmentData->next == nullptr);
+			mEosReached = !nextFragmentData->next;
 		}
 
-		// Compliment this state with last init header push status
-		if (mActiveTuneType == eTUNETYPE_SEEKTOLIVE)
-		{
-			mEosReached &= !mNewInitWaiting;
```
DomSyna, line 275: Why don't we need to do this anymore ?

AampTsbReader.cpp
```diff
@@ -310,31 +306,66 @@ void AampTsbReader::ReadNext(TsbFragmentDataPtr nextFragmentData)
 			GetMediaTypeName(mMediaType), nextFragmentData->GetAbsolutePosition().inSeconds(), mUpcomingFragmentPosition.inSeconds(), mEosReached, mNewInitWaiting, mIsNextFragmentDisc,
 			mIsPeriodBoundary, mAamp->mTrickModePositionEOS, mCurrentRate);
 	}
+	else
+	{
+		// Handle null fragment case - this indicates we've reached the end of available data
+		AAMPLOG_INFO("[%s] Null fragment read, setting EOS.", GetMediaTypeName(mMediaType));
+		mEosReached = true;
```
DomSyna, line 312: Do we really want this here ? what if we are in FFWD and haven't actually reached the EOS position as calculated ?

AampTsbReader.cpp
DomSyna: Given its Init job sets mTrackEnabled to true or false as appropriate wouldn't it be safer to set this to false going forward. In reality it probably makes no real difference what it is set to here as Init will set it. But leaving it as true here may introduce a race condition at some point if term is called whilst another thread is processing and checks the enabled flag to see if it should continue as the track is being left enabled

AampTsbReader.cpp
p-bond: Setting mTrackEnabled inappropriately results in the TSBReader thread getting into a joinable state and being terminated prematurely.

AampTsbReader.cpp
p-bond: Setting this to false here results in AampTsbReader being terminated prematurely.

AampTsbReader.cpp
p-bond: Moved so debug is displayed even if we're not going backwards.

AampTsbReader.cpp
p-bond: Added all the seek types in.

AampTsbReader.cpp
p-bond: Believe race condition this covered is no longer an issue with simplified calculation.

AampTsbReader.cpp
```diff
@@ -254,31 +260,29 @@ TsbFragmentDataPtr AampTsbReader::FindNext(AampTime offset)
...
+			mEosReached = !nextFragmentData->next;
 		}
 
-		// Compliment this state with last init header push status
-		if (mActiveTuneType == eTUNETYPE_SEEKTOLIVE)
-		{
-			mEosReached &= !mNewInitWaiting;
-		}
-		// Determine if the next fragment is discontinuous.
+		// Complement this state with last init header push status/ Determine if the next fragment is discontinuous.
```
DomSyna: No longer relevant ;)

AampTsbReader.cpp
DomSyna: OK should never get here as caller won't call this with null ptr

AampTsbReader.cpp
p-bond: Then excised after discussion & concluding this was no longer relevant.

AampTsbReader.cpp
p-bond: This appears to have been added to fix a race condition due to calculation issues.  Should not arise with list navigation.

AampTsbReader.cpp
p-bond: Removed after discussion.

DomSyna (approved):

pstroffolino (approved):

PR #296, opened 2025-07-19
Revert "VPLAY-9299:Address 200ms tune delay when using enableMediaProcessor as true"

pstroffolino:
VPLAY-9299: Reverts rdkcentral/aamp#280 "Address 200ms tune delay when using enableMediaProcessor as true"

Reason for Change: this optimization is breaking L2 test 8008; video freezes after playing cdai ads shorter than source period

### DIFF
--- a/StreamAbstractionAAMP.h
+++ b/StreamAbstractionAAMP.h
@@ -1153,21 +1153,6 @@ public:
 	{
 		return 0.0;
 	}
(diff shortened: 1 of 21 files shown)


pstroffolino (commented):
temporary revert to avoid regression with CDAI (L2 test 8008)

PR #329, opened 2025-07-29
VPLAY-10150 - Remove floating point comparisons in AampTsbReader

psiva01:
Additional logging for playback rate

move rate init

VPLAY-10150 - Remove floating point comparisons in AampTsbReader


jfagunde (commented):
I have verified that this cherry-pick is the same that was merged to dev_sprint_25_2 in https://github.com/rdkcentral/aamp/pull/258

PR #340, opened 2025-08-04
VPLAY-10421 - CC is not working in VOD with pre-roll ads

anjali-syna:
CC is not working in VOD with pre-roll ads. Logs are lost and difficult to debug the issue. So moving to printf temporarily for debugging


anjali-syna:
wrong PR

PR #343, opened 2025-08-04
VPLAY-10516: Fix L2 5003 failures in sprint

Vinish100:
Reason for change: SetLocalTSBInjection before querying GetFirstPTS to get the accurate PTS value from TSBReader Test Procedure: All L2 5003 should pass
Risks: Low


fragmentcollector_mpd.cpp
```diff
@@ -3496,6 +3496,17 @@ AAMPStatusType StreamAbstractionAAMP_MPD::InitTsbReader(TuneType tuneType)
 		{
 			retVal = eAAMPSTATUS_SEEK_RANGE_ERROR;
 		}
+
+		if (retVal == eAAMPSTATUS_OK && aamp->IsLocalAAMPTsbInjection())
```
jfagunde: How about moving this block of code to inside `if(eAAMPSTATUS_OK == retVal)`, after the MIL in line 3493?

Vinish100: Fixed 👍

jfagunde (approved):

PR #347, opened 2025-08-05
VPLAY-10148 Add latency correction related log lines

jfagunde:
Reason for Change: Help follow the code flow
Test Procedure: Verify new log lines added and check that there is no log flooding
Priority: P3
Risks: Low

Signed-off-by: Jose Fagundez <jfagunde@synamedia.com>


fragmentcollector_mpd.cpp
```diff
@@ -3842,6 +3842,7 @@ AAMPStatusType StreamAbstractionAAMP_MPD::Init(TuneType tuneType)
 					{
 						if (mLowLatencyMode)
 						{
+							AAMPLOG_INFO("Set LLDashAdjustSpeed to false for tuneType %d", tuneType);
```
cpc005, line 3880: So will get this log line and the one added to SetLLDashAdjustSpeed().
How about add the tuneType to this log line which is earlier in the func
AAMPLOG_MIL("StreamAbstractionAAMP_MPD: MPD duration val %" PRIu64 " seconds", durationMs/1000);

priv_aamp.cpp
```diff
@@ -1246,6 +1246,7 @@ PrivateInstanceAAMP::PrivateInstanceAAMP(AampConfig *config) : mReportProgressPo
 	, prevFirstPeriodStartTime(0)
 	, mIsFlushOperationInProgress(false)
 {
+	AAMPLOG_MIL("Create Private Player %d", mPlayerId);
```
cpc005, line 1320: ("mPlayerId %d", mPlayerId)

priv_aamp.h
```diff
@@ -3467,7 +3467,11 @@ class PrivateInstanceAAMP : public DrmCallbacks, public std::enable_shared_from_
 	 *   @param[in] state - true or false
 	 *   @return void
 	 */
-	void SetLLDashAdjustSpeed(bool state) { bLLDashAdjustPlayerSpeed = state; }
+	void SetLLDashAdjustSpeed(bool state)
+	{
+		AAMPLOG_INFO("Set LLDash adjust speed to %d", state);
```
cpc005, line 3475: AAMPLOG_INFO("bLLDashAdjustPlayerSpeed %d", state);

fragmentcollector_mpd.cpp
jfagunde: I added this log line to identify where was SetLLDashAdjustSpeed() called from. It is difficult with only the MIL printed earlier and the log line in that function.

priv_aamp.cpp
jfagunde: I think "Create Private Player 0" is more descriptive of what is happening. It would be easier to find in the log when Player instances are created.
Also, I don't like using variable names in log lines. It means that if the variable changes its name we also have to change log lines.

priv_aamp.h
jfagunde: See the comments above.

Vinish100:
@jfagunde  Changes looks good. But L1 is failing... On hold

jfagunde:
> @jfagunde Changes looks good. But L1 is failing... On hold

L1 test failure fixed.

Vinish100 (approved):

PR #348, opened 2025-08-05
AAMP TSB - CDAI Ad Presentation freezes on Channel Four

cpc005:
When a multi AD break is played then the start position of ADs following the first is not being calculated correctly. This leads to the AD length check failing and the AD being terminated early with the log line.
[AdvanceTrack][9078][CDAI] Track[1] Adbreak ended early. Terminating Ad playback. fragmentTime[1753948141.197000] periodStartOffset[1753948118.557001]

The solution is to calculate the ad start position from absoluteAdBreakStartTime not mLivePeriodCulledSeconds which is wrong on CH4


fragmentcollector_mpd.cpp
```diff
@@ -7967,14 +7967,19 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
-						pMediaStreamContext->fragmentTime += (double)((double)mCdaiObject->mCurAds->at(mCdaiObject->mCurAdIdx).basePeriodOffset / 1000.0);
+						//Set the period start back to the begining of the base period and then add basePeriodOffset
+						//to get the start for this AD
+						double absoluteAdBreakStartTime = mCdaiObject->mAdBreaks[mBasePeriodId].mAbsoluteAdBreakStartTime.inSeconds();
+						pMediaStreamContext->fragmentTime = absoluteAdBreakStartTime;
+						// convert to seconds, standard implicit conversion
+						pMediaStreamContext->fragmentTime += mCdaiObject->mCurAds->at(mCdaiObject->mCurAdIdx).basePeriodOffset / 1000.0;
 						if((aamp->mMPDPeriodsInfo.at(0).periodId) == mBasePeriodId)
 						{
 							pMediaStreamContext->fragmentTime -= mLivePeriodCulledSeconds;
```
Vinish100: @cpc005 I looked into this code and it seems like it was added to fix the same scenario that you are seeing
https://code.rdkcentral.com/r/plugins/gitiles/rdk/components/generic/aamp/+/4fbc61c7a6ae61221b6b7a3fa129f4778a003f38
So if the current period is getting culled. ie mLivePeriodCulledSeconds > 0, then it will further offset your fragmentTime backwards, which is not expected. Could you please look into why mLivePeriodCulledSeconds is not working in your case

cpc005: The mLivePeriodCulledSeconds is not working because the manifest is including an empty period preceding the basePeriod that is being used for substituting ads. 
See
https://cim.slack.com/archives/C0414S5F7GF/p1754565150535879?thread_ts=1754045329.205279&cid=C0414S5F7GF

cpc005: It would seem a better fix with the change I had using
mCdaiObject->mAdBreaks[mBasePeriodId].mAbsoluteAdBreakStartTime.inSeconds() for better cohesion in CDAI
Remove the variable mLivePeriodCulledSeconds; which leaves the problem of no tests for GetCulledSeconds() and it does not work on Ch4 ad periods

Vinish100 (approved):
Looks good to me. Lets add a scenario for this in L2, 8004 might be a good candidate

fragmentcollector_mpd.cpp
cpc005: Vinish K B
  [Yesterday at 5:57 PM](https://cim.slack.com/archives/CE4JSGZK5/p1755017873541139?thread_ts=1754988476.329899&cid=CE4JSGZK5)
[@Patrick](https://comcast.enterprise.slack.com/team/U04M1DCS44R) Looks good to me. Especially removing a variable and reusing existing ones.
[@Nandu](https://comcast.enterprise.slack.com/team/WHDNWS1BM) I believe we understood why the empty period was not culled properly. Would you like to take up VPLAY-10639?

PR #349, opened 2025-08-05
RDKEMW-6727 : Move Plugin-Tests from aamp to middleware utest

narenr94:
reason for change : plugin will be part of middleware-player-interface hence related tests also need to be migrated
test procedure : run L1 
signed-off : Naren Ramesh <naren_ramesh@comcast.com>


Vinish100 (approved):
Looks good. Waiting for L1 to pass

pstroffolino (approved):

PR #396, opened 2025-08-20
Testing Jira ticket validation from feature to target v1

Abhi-jith-S:
test


Abhi-jith-S:
Closing the PR as it is working as expected. it is failing in feature to target PR without proper Jira ticket number.

PR #402, opened 2025-08-20
VPLAY-10403 Fix IsVideoSink() when using Rialto

jfagunde:
Reason for Change: Mute video when PIN is required
Test Procedure: Verify that video is not displayed when PIN is required
Priority: P1
Risks: Low

Signed-off-by: Jose Fagundez [jfagunde@synamedia.com](mailto:jfagunde@synamedia.com)

### DIFF
--- a/middleware/InterfacePlayerRDK.cpp
+++ b/middleware/InterfacePlayerRDK.cpp
@@ -4491,7 +4491,7 @@ void InterfacePlayerRDK::SetVideoMute(bool muted)
 	}
 	else
 	{
-		MW_LOG_WARN("InterfacePlayerRDK not setting video mute");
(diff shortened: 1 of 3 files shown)


DomSyna (approved):
Maybe add L1 test ?
I assume the only possible sinks are westeros and rialto ?

jfagunde:
> Maybe add L1 test ? I assume the only possible sinks are westeros and rialto ?

It would be very easy to test this fix in an L1 test. The problem is that there are no L1 tests for the middleware at the moment. It needs to be decided how those tests will be implemented, and it is better not to do that as part of this Jira.

pstroffolino (approved):

PR #453, opened 2025-09-02

lashmintha:
Risks: Low
Priority: P1

### DIFF
--- a/fragmentcollector_mpd.cpp
+++ b/fragmentcollector_mpd.cpp
@@ -161,6 +161,8 @@ StreamAbstractionAAMP_MPD::StreamAbstractionAAMP_MPD(class PrivateInstanceAAMP *
 	,mIsSegmentTimelineEnabled(false)
 	,mSeekedInPeriod(false)
 	,mIsFinalFirstPTS(false)
+	,mCachedBandwidth(0)
+	,mCachedRepresentationIndex(-1)
 {
 	this->aamp = aamp;
 	if (aamp->mDRMLicenseManager)
@@ -2612,6 +2614,7 @@ double StreamAbstractionAAMP_MPD::SkipFragments( MediaStreamContext *pMediaStrea
 								if(ISCONFIGSET(eAAMPConfig_MidFragmentSeek))
 								{
 									mFirstPTS += mVideoPosRemainder;
+
 									if(mVideoPosRemainder > fragmentDuration/2)
 									{
 										if(aamp->GetInitialBufferDuration() == 0)
@@ -7653,13 +7656,61 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW, 
 							// chosenAdaptationIdxs.insert(0);
 						}
 					}
+					// Check if representation should be reset for period change
+					// Use adaptation set 0 (video) to check for representation changes
+					bool shouldResetRepresentationIndex = false;
+					if (periodChanged && i == eMEDIATYPE_VIDEO)
+					{
+						// For video period changes, check if cached bandwidth exists at the same representation index
+						bool representationChanged = true; // Default to changed
+						if (mCachedBandwidth > 0 && mCachedRepresentationIndex >= 0 && adaptationSets.size() > 0)
+						{
+							// Use adaptation set 0 for video only
+							IAdaptationSet* adaptationSet = adaptationSets.at(0);
+							const auto &representations = adaptationSet->GetRepresentation();
+							for (int reprIdx = 0; reprIdx < representations.size(); reprIdx++)
+							{
+								if (reprIdx == mCachedRepresentationIndex)
+								{
+									IRepresentation *representation = representations.at(reprIdx);
+									uint32_t currentBandwidth = representation->GetBandwidth();
+									if (currentBandwidth == mCachedBandwidth)
+									{
+										representationChanged = false;
+										AAMPLOG_INFO("Representation unchanged - Bandwidth %u exists at same index %d in adaptationSet 0", 
+													 mCachedBandwidth, mCachedRepresentationIndex);
+									}
+									else
+									{
+										AAMPLOG_INFO("Representation changed - Bandwidth changed from %u to %u at index %d in adaptationSet 0", 
+													 mCachedBandwidth, currentBandwidth, mCachedRepresentationIndex);
+									}
+									break;
+								}
+							}
+							if (representationChanged)
+							{
+								AAMPLOG_INFO("Representation changed - cached bandwidth %u at index %d not found in current period adaptationSet 0", 
+											 mCachedBandwidth, mCachedRepresentationIndex);
+							}
+						}
+						else
+						{
+							AAMPLOG_INFO("Representation changed - no cached bandwidth, invalid index, or no adaptation sets");
+						}
+						shouldResetRepresentationIndex = representationChanged;
+					}
+					if (shouldResetRepresentationIndex)
+					{
+						// reset representationIndex to -1 to allow updating the currentProfileIndex for period change.
+						pMediaStreamContext->representationIndex = -1;
+						AAMPLOG_INFO("Resetting representationIndex for track %d", i);
+					}
 					if ((representationCount != GetProfileCount()) && mStreamInfo)
 					{
 						SAFE_DELETE_ARRAY(mStreamInfo);
-
-						// reset representationIndex to -1 to allow updating the currentProfileIndex for period change.
-						pMediaStreamContext->representationIndex = -1;
 						AAMPLOG_WARN("representationIndex set to (-1) to find currentProfileIndex");
+						pMediaStreamContext->representationIndex = -1;
 					}
 					if (!mStreamInfo)
 					{
@@ -7941,7 +7992,11 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW, 
 							}
 							UpdateIframeTracks();
 						}
-						currentProfileIndex = GetDesiredProfile(false);
+						//currentProfileIndex = GetDesiredProfile(false);
+						/* Updating profile index based on the available bandwidth */
+						UpdateProfileBasedOnFragmentCache();
+						AAMPLOG_INFO("Profile updated based on fragment cache [%d]",currentProfileIndex);
+
 						// Adaptation Set Index corresponding to a particular profile
 						pMediaStreamContext->adaptationSetIdx = mProfileMaps[currentProfileIndex].adaptationSetIndex;
 						// Representation Index within a particular Adaptation Set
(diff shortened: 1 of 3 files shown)


fragmentcollector_mpd.cpp
```diff
@@ -520,6 +520,8 @@ static int GetDesiredVideoCodecIndex(IAdaptationSet *adaptationSet)
 		else if(adapCodecs.size())
 			codecValue = adapCodecs.at(0);
 		selectedRepIdx = representationIndex;
+
+		AAMPLOG_INFO("StreamAbstractionAAMP_MPD: Video SelectedRepIndex : %d , Codec : %s", selectedRepIdx, codecValue.c_str());
```
Vinish100: This log is required?
Its inside a loop. will it cause flooding?

fragmentcollector_mpd.cpp
```diff
@@ -7639,6 +7645,12 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 							// chosenAdaptationIdxs.insert(0);
 						}
 					}
+				   /* Resetting the profile, due to in case if the profile order got changed in the Multiperiod stream,
+					* desired profile selection may not happen based on the old profile index.
+					* So resetting the profile index to -1, to avoid the wrong profile selection
+					*/
+					AAMPLOG_INFO("Resetting rerepresentationIndex from %d to -1", pMediaStreamContext->representationIndex);
```
Vinish100: rerepresentationIndex is a typo. How did it get past godspell?

fragmentcollector_mpd.cpp
```diff
@@ -7960,6 +7972,7 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 			if(pMediaStreamContext->representationIndex < pMediaStreamContext->adaptationSet->GetRepresentation().size())
 			{
 				pMediaStreamContext->representation = pMediaStreamContext->adaptationSet->GetRepresentation().at(pMediaStreamContext->representationIndex);
+				AAMPLOG_INFO("Track %d selected AdaptationSetID %d Representation index %d", pMediaStreamContext->type, pMediaStreamContext->adaptationSetId, pMediaStreamContext->representationIndex);
```
Vinish100: Why so many logs have been added. Are they really required or just debug logs?

fragmentcollector_mpd.cpp
```diff
@@ -7639,6 +7645,12 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 							// chosenAdaptationIdxs.insert(0);
 						}
 					}
+				   /* Resetting the profile, due to in case if the profile order got changed in the Multiperiod stream,
+					* desired profile selection may not happen based on the old profile index.
+					* So resetting the profile index to -1, to avoid the wrong profile selection
+					*/
+					AAMPLOG_INFO("Resetting rerepresentationIndex from %d to -1", pMediaStreamContext->representationIndex);
+					pMediaStreamContext->representationIndex = -1;
```
Vinish100: Should we be doing this for FogTSB? As per 
https://github.com/rdkcentral/aamp/pull/453/files#diff-9412049d3685e68289ccd0b5e0a718c64d26b9a783882ba14700abd57e136be3L7954 it looks like unexpected

lashmintha: The representation indez updated under the below condition, so there won't be in any conflict when isFogTsb is enabled
else if(!isFogTsb && mUpdateStreamInfo)
{
}

fragmentcollector_mpd.cpp
```diff
@@ -7639,6 +7640,12 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 							// chosenAdaptationIdxs.insert(0);
 						}
 					}
+				   /* Resetting the profile, due to in case if the profile order got changed in the Multiperiod stream,
+					* desired profile selection may not happen based on the old profile index.
+					* So resetting the profile index to -1, to avoid the wrong profile selection
+					*/
+					AAMPLOG_INFO("Resetting representationIndex from %d to -1", pMediaStreamContext->representationIndex);
+					pMediaStreamContext->representationIndex = -1;
```
Vinish100: Remove lines at https://github.com/rdkcentral/aamp/pull/453/files#diff-9412049d3685e68289ccd0b5e0a718c64d26b9a783882ba14700abd57e136be3R7654, it doesn't have any meaning anymore as representationIndex is reset already here ( rephrase the log line at line 7655)

lashmintha: As the representationCount and GetProfileCount() are equal so the below condition doesn't met..!
if ((representationCount != GetProfileCount()) && mStreamInfo)
{
}

fragmentcollector_mpd.cpp
```diff
@@ -7668,6 +7669,31 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 						for (int reprIdx = 0; reprIdx < representations.size(); reprIdx++)
 						{
 							IRepresentation *representation = representations.at(reprIdx);
+							if(reprIdx == 0)
```
Vinish100: Basically you are comparing the firstId of representation between periods and see if they differ or not. If they differ invalidate the representationIndex. This can be easily broken..
Period 1 -> rep1, rep2, rep3, rep4
Period 2 -> rep1, rep3, rep4, rep2
Can't we just invalidate the representationIdx and pick the best one based on available bandwidth?

lashmintha:
will create new PR

fragmentcollector_mpd.cpp
```diff
@@ -7653,13 +7656,61 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
+							// Use adaptation set 0 for video only
+							IAdaptationSet* adaptationSet = adaptationSets.at(0);
+							const auto &representations = adaptationSet->GetRepresentation();
+							for (int reprIdx = 0; reprIdx < representations.size(); reprIdx++)
+							{
+								if (reprIdx == mCachedRepresentationIndex)
+								{
+									IRepresentation *representation = representations.at(reprIdx);
+									std::string currentRepID = representation->GetId();
+									if (currentRepID == mCachedRepresentationID)
```
nu641001: Representation IDs are not guaranteed to be unique, they can repeat across different periods. Would it make more sense to compare based on bandwidth instead? If the bandwidth remains unchanged from the previous index, perhaps we can continue without switching. Could you please consider reviewing this approach?

fragmentcollector_mpd.cpp
```diff
@@ -7653,13 +7656,61 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
-
 						// reset representationIndex to -1 to allow updating the currentProfileIndex for period change.
 						pMediaStreamContext->representationIndex = -1;
+						AAMPLOG_INFO("Resetting representationIndex for track %d", i);
+					}
+					if ((representationCount != GetProfileCount()) && mStreamInfo)
+					{
+						SAFE_DELETE_ARRAY(mStreamInfo);
 						AAMPLOG_WARN("representationIndex set to (-1) to find currentProfileIndex");
+						pMediaStreamContext->representationIndex = -1;
```
nu641001, line 7713: Also, if there aren’t any L1 tests covering this scenario, it would be good to add one. I don’t think there are many tests for UpdateTrackInfo currently.

lashmintha: will include L1 nandhu

fragmentcollector_mpd.cpp
```diff
@@ -161,6 +161,8 @@ StreamAbstractionAAMP_MPD::StreamAbstractionAAMP_MPD(class PrivateInstanceAAMP *
 	,mIsSegmentTimelineEnabled(false)
 	,mSeekedInPeriod(false)
 	,mIsFinalFirstPTS(false)
+	,mCachedBandwidth(0)
```
Vinish100, line 164: Don't think these needs to be class members

fragmentcollector_mpd.cpp
```diff
@@ -7653,13 +7656,61 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 							// chosenAdaptationIdxs.insert(0);
 						}
 					}
-					if ((representationCount != GetProfileCount()) && mStreamInfo)
+					// Check if representation should be reset for period change
+					// Use adaptation set 0 (video) to check for representation changes
+					bool shouldResetRepresentationIndex = false;
```
Vinish100, line 7661: pMediaStreamContext->representationIndex is already available and thus makes mCachedRepresentationIndex moot. 
the mCachedBandwidth can be retrieved by called ABRManager apis to get the bandwidth of the currentProfileIndex before its reset.

fragmentcollector_mpd.cpp
```diff
@@ -7653,13 +7656,61 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
+					if (periodChanged && i == eMEDIATYPE_VIDEO)
+					{
+						// For video period changes, check if cached bandwidth exists at the same representation index
+						bool representationChanged = true; // Default to changed
+						if (mCachedBandwidth > 0 && mCachedRepresentationIndex >= 0 && adaptationSets.size() > 0)
+						{
+							// Use adaptation set 0 for video only
+							IAdaptationSet* adaptationSet = adaptationSets.at(0);
+							const auto &representations = adaptationSet->GetRepresentation();
+							for (int reprIdx = 0; reprIdx < representations.size(); reprIdx++)
```
Vinish100, line 7671: Whole loop is not required. 
reprIdx = mCachedRepresentationIndex and make sure its within the representations.size()

fragmentcollector_mpd.cpp
```diff
@@ -7653,13 +7656,61 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
+						}
+						shouldResetRepresentationIndex = representationChanged;
+					}
+					if (shouldResetRepresentationIndex)
 					{
-						SAFE_DELETE_ARRAY(mStreamInfo);
-
 						// reset representationIndex to -1 to allow updating the currentProfileIndex for period change.
 						pMediaStreamContext->representationIndex = -1;
+						AAMPLOG_INFO("Resetting representationIndex for track %d", i);
```
Vinish100, line 7707: representationIndex = -1 should be done only in one place.
representationCount != GetProfileCount() should set shouldResetRepresentationIndex = true

PR #454, opened 2025-09-02
VPLAY-10764 CC Visible when parental control at tune time

DomSyna:
Reason for change: When the video is muted the subtitles get reenabled if the aamp interface is called to unmute the subtitles. This can result in PIN protected content having the AV muted but subtitles present.

Risks: Low
Test Procedure: Refer Jira ticket

### DIFF
--- a/priv_aamp.cpp
+++ b/priv_aamp.cpp
@@ -7153,7 +7153,7 @@ void PrivateInstanceAAMP::SetSubtitleMute(bool muted)
 	StreamSink *sink = AampStreamSinkManager::GetInstance().GetStreamSink(this);
 	if (sink)
 	{
-		sink->SetSubtitleMute(video_muted || muted);
+		sink->SetSubtitleMute(video_muted | muted);
 	}
 }
 
--- a/priv_aamp.cpp
+++ b/priv_aamp.cpp
@@ -7153,7 +7153,7 @@ void PrivateInstanceAAMP::SetSubtitleMute(bool muted)
 	StreamSink *sink = AampStreamSinkManager::GetInstance().GetStreamSink(this);
 	if (sink)
 	{
-		sink->SetSubtitleMute(muted);
+		sink->SetSubtitleMute((video_muted) ? video_muted : muted);
 	}
 }
 
--- a/priv_aamp.cpp
+++ b/priv_aamp.cpp
@@ -7153,7 +7153,7 @@ void PrivateInstanceAAMP::SetSubtitleMute(bool muted)
 	StreamSink *sink = AampStreamSinkManager::GetInstance().GetStreamSink(this);
 	if (sink)
 	{
-		sink->SetSubtitleMute((video_muted) ? video_muted : muted);
+		sink->SetSubtitleMute(video_muted | muted);
 	}
 }
 
--- a/priv_aamp.cpp
+++ b/priv_aamp.cpp
@@ -7153,7 +7153,7 @@ void PrivateInstanceAAMP::SetSubtitleMute(bool muted)
 	StreamSink *sink = AampStreamSinkManager::GetInstance().GetStreamSink(this);
 	if (sink)
 	{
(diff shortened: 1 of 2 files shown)


priv_aamp.cpp
```diff
@@ -7153,7 +7153,7 @@ void PrivateInstanceAAMP::SetSubtitleMute(bool muted)
 	StreamSink *sink = AampStreamSinkManager::GetInstance().GetStreamSink(this);
 	if (sink)
 	{
-		sink->SetSubtitleMute(muted);
+		sink->SetSubtitleMute(video_muted | muted);
```
dnfaulkner: Shouldn't that be || instead of | (bit-wise OR)?

test/utests/tests/PrivateInstanceAAMP/SubtitleMuteTests.cpp
```diff
@@ -42,79 +42,104 @@ class SubtitleMuteTests : public testing::TestWithParam< std::pair<bool, bool> >
...
+	auto param = GetParam();
+	bool currState = param.first;
+	bool inputState = param.second;
+	CacheAndMuteSubtitles(currState, inputState);
+}
+
+TEST_P(SubtitleMuteTests, SetSubtitleMuteTests)
+{
+	auto param = GetParam();
+	bool videoMuted = param.first;
```
jfagunde: The comments in TestSubtitleMute are the other way around (subtile muted first)

priv_aamp.cpp
DomSyna: No :)

test/utests/tests/PrivateInstanceAAMP/SubtitleMuteTests.cpp
DomSyna: You are right

priv_aamp.cpp
dnfaulkner: OK :)

jfagunde: It would be good to explain why not so the reasoning gets recorded.
I agree with Dave. I think either works but logical OR makes more sense than bitwise OR.

test/utests/tests/PrivateInstanceAAMP/SubtitleMuteTests.cpp
```diff
@@ -42,79 +42,104 @@ class SubtitleMuteTests : public testing::TestWithParam< std::pair<bool, bool> >
...
+
+TEST_P(SubtitleMuteTests, SetSubtitleMuteTests)
+{
+	auto param {GetParam()};
+	bool subtitleMuted {param.first};
+	bool videoMuted {param.second};
+	TestSetSubtitleMute(videoMuted, subtitleMuted);
 }
 
 // Run PlaybackSpeedTests tests at various speeds
```
jfagunde: [Existing code] This comment seems wrong. Probably copied from some other test.

jfagunde (approved):

priv_aamp.cpp
dnfaulkner: My explanation why was originally because of my C background, where bool type is just an enum and you can't assume that the bool enums have specific numerical values that can be ored together.
When Dom highlighted that bool is a natural type in C++ and the compiler knows to automatically cast the bools to numerical values where true =1 and false=0 and hence can be bitwise ored together and then it will automatically cast back to a bool result, I was happy that a single | is valid (plus it results in more optimal machine code because there's no branches in the evaluation).
However, the consensus is that || makes more sense than |. If compiler optimisation is turned on, it produces identical / optimal machine code with no branches (try it at [https://godbolt.org](url)).
Interesting discussion here: [https://stackoverflow.com/questions/24542/using-bitwise-operators-for-booleans-in-c](url)

dnfaulkner (approved):

pstroffolino (approved):

PR #463, opened 2025-09-04
VPLAY-10900 Intermittent no Ad events in L3 2014

cpc005:
With Cloud TSB. When a rew stops exactly at the start of a period where an Ad has previously been played then  the Ad event code will set immediate=True. This results in   AAMP_EVENT_AD_PLACEMENT_START, AAMP_EVENT_AD_PLACEMENT_END getting sent but not AAMP_EVENT_AD_PLACEMENT_PROGRESS.
Fix so that AAMP_EVENT_AD_PLACEMENT_PROGRESS is also generated


cpc005:
Not going to deliver

PR #497, opened 2025-09-16
VPLAY-10931 Mute SubTec when video is muted

jfagunde:
Reason for change: Refactoring to improve efficiency and readability

Summary of changes:
* Call SubTec with (video_muted || subtitles_muted)
* Move SetVideoMute and SetSubtitleMute logic from PlayerInstanceAAMP to PrivateInstanceAAMP
* Replace CacheAndApplySubtitleMute() with calls to SetCCStatus()
* Update existing L1 tests related to subtitles

Risks: Low

Test Procedure: Turn subtitles on and check regressions when enabling/disabling video or subtitles (e.g. due to a PIN request)


priv_aamp.cpp
```diff
@@ -6179,8 +6179,8 @@ void PrivateInstanceAAMP::Tune(const char *mainManifestUrl,
 		if (mpStreamAbstractionAAMP)
 		{
 			//There two fns are being called in PlayerInstanceAAMP::SetVideoMute
-			SetVideoMute(video_muted);
-			CacheAndApplySubtitleMute(video_muted);
+			SetVideoMuteInternal(video_muted);
+			SetCCStatus(!subtitles_muted);
```
cpc005: If video muted do we need to ensure we set subtitles the same 
SetCCStatus(!video_muted);

priv_aamp.cpp
```diff
@@ -7122,6 +7122,36 @@ void PrivateInstanceAAMP::SetVideoZoom(VideoZoomMode zoom)
  *   @brief Enable/ Disable Video.
  */
 void PrivateInstanceAAMP::SetVideoMute(bool muted)
+{
+	AAMPLOG_INFO("mute == %s subtitles_muted == %s", muted?"true":"false", subtitles_muted?"true":"false");
```
cpc005: Make a shorter log line 
'==' really!

main_aamp.cpp
```diff
@@ -1451,33 +1451,8 @@ void PlayerInstanceAAMP::SetVideoZoom(VideoZoomMode zoom)
...
-			}
-			aamp->ReleaseStreamLock();
-		}
-		else
-		{
-			AAMPLOG_WARN("StreamLock is not available, value has been cached");
-			aamp->mApplyCachedVideoMute = true;
-		}
-	}
+	AAMPLOG_MIL(" mute == %s", muted?"true":"false");
```
cpc005: How about
("mute %d",muted)
there is no precedent for having '==' in a log line and very little for outputting bool as "true"

priv_aamp.cpp
jfagunde: It is done inside PrivateInstanceAAMP::SetCCStatus(). It is calling the lower layers with (video_muted || subtitles_muted)

priv_aamp.cpp
```diff
@@ -10979,13 +11029,13 @@ void PrivateInstanceAAMP::SetCCStatus(bool enabled)
 	subtitles_muted = !enabled;
 	if (mpStreamAbstractionAAMP)
 	{
-		mpStreamAbstractionAAMP->MuteSubtitles(subtitles_muted);
+		mpStreamAbstractionAAMP->MuteSubtitles(video_muted || subtitles_muted);
```
dnfaulkner: Why not change line 11029 to:
subtitles_muted = !enabled | video_muted;
Then it's clearer what setting is being applied and avoid multiple duplicate evaluations for mpStreamAbstractionAAMP->MuteSubtitles(), mpStreamAbstractionAAMP->MuteSidecarSubtitles and SetSubtitleMuteInternal

priv_aamp.cpp
```diff
@@ -10979,13 +11029,13 @@ void PrivateInstanceAAMP::SetCCStatus(bool enabled)
...
 		if (HasSidecarData())
 		{ // has sidecar data
-			mpStreamAbstractionAAMP->MuteSidecarSubtitles(subtitles_muted);
+			mpStreamAbstractionAAMP->MuteSidecarSubtitles(video_muted || subtitles_muted);
```
dnfaulkner: See line 11032

priv_aamp.cpp
```diff
@@ -10979,13 +11029,13 @@ void PrivateInstanceAAMP::SetCCStatus(bool enabled)
...
 		}
 	}
-	SetSubtitleMute(subtitles_muted);
+	SetSubtitleMuteInternal(video_muted || subtitles_muted);
```
dnfaulkner: See line 11032

priv_aamp.cpp
dnfaulkner: Yeah, give logging less work to do - use %d instead (and uses less code space). I think we all understand that 1=true and 0=false

main_aamp.cpp
dnfaulkner: Yeah, give logging less work to do - use %d instead (and uses less code space). I think we all understand that 1=true and 0=false

priv_aamp.cpp
jfagunde: I'll use a local variable to have the logic in a single place; but subtitles_muted is what the user requested, i.e. if the user enables subtitles it will be true, even if subtitles are muted because he has to enter the PIN.

main_aamp.cpp
jfagunde: Done

priv_aamp.cpp

main_aamp.cpp
```diff
@@ -1487,22 +1462,8 @@ void PlayerInstanceAAMP::SetVideoMute(bool muted)
...
-		{
-			aamp->SetSubtitleMute(muted);
-		}
-		else
-		{
-			AAMPLOG_WARN("Player is in state eSTATE_IDLE, value has been cached");
-		}
-		aamp->ReleaseStreamLock();
-	}
+	AAMPLOG_MIL(" mute == %s", muted?"true":"false");
```
dnfaulkner: See line 1454

priv_aamp.cpp
dnfaulkner: I don't understand cpc005's comment here. SetCCStatus() takes an enabled flag not a muted flag, so it's the other way up.

priv_aamp.cpp

priv_aamp.cpp

main_aamp.cpp
jfagunde: Done

priv_aamp.cpp
dnfaulkner: Ah, yes. Sorry. I forgot that subtitles_muted was a class member!

dnfaulkner (approved):

priv_aamp.cpp
cpc005: An unsubstantiated concern is that in these lines of code we are setting video to a know value but just toggling the state of the subtitles. If subtitles get out of step then they are going to stay that way if they only ever get toggled.

jfagunde: The state of subtitles is not being toggled. But I see this code is confusing. I am doing a bit of refactoring to try to make it clearer.

cpc005: Apologies, it is not a toggle

SetCCStatus(!subtitles_muted);
...
void PrivateInstanceAAMP::SetCCStatus(bool enabled)
{
	PlayerCCManager::GetInstance()->SetStatus(enabled);
	AcquireStreamLock();
	subtitles_muted = !enabled;

cpc005 (approved):

pstroffolino (approved):

PR #489, opened 2025-09-12
VPLAY-11040 Reduce logging from PushNextTsbFragment()

jfagunde:
Reason for Change: Log output from PushNextTsbFragment() during trick
                   modes is very noisy

Summary of Changes:
* Do not call PushNextTsbFragment() for tracks that are not enabled
* During FF "Failed to read next fragment" is expected so moved to TRACE
* Move noisy INFO "No next fragment available" to TRACE
* Move implementation of two AampTsbReader methods to source file
* Improve PushNextTsbFragment() method comments
* Add L1 test verifying only fragmentcollector_mpd.cpp
* Test change in L1 test
* Add necessary fakes and mocks

Test Procedure: Check log lines from PushNextTsbFragment() when using trick modes

Risk: Low

PR #502, opened 2025-09-18
VPLAY-11125 [AAMP-TSB] Fragments not skipped when ff to live at high rate

jfagunde:
Reason for Change: Too many i-frames injected when ff to live at high rate

Summary of Changes:
* SkipFragment() output parameter nextFragmentData is null if the next
  fragment to be presented is not available
* Test new behaviour in AampTsbSessionManagerTests L1 test
* Recover old AampTsbSessionManagerTests L1 tests that had been deleted
* Delete L1 SkipFragment_TrickplayRates that was not testing AAMP code

Test Procedure: With >10min AAMP TSB, FF BoS to live and check i-frames displayed

Risk: Low


AampTSBSessionManager.cpp
```diff
@@ -824,14 +824,19 @@ void AampTSBSessionManager::SkipFragment(std::shared_ptr<AampTsbReader>& reader,
 				{
 					tmp = nextFragmentData->prev;
 				}
+				nextFragmentData = tmp;
```
DomSyna: Consider *8 with 1.92 sec segments.
delta will be 2 initially so we will skip every other segment as delta will be greater than 1.92 and hence not the the break on the first pass.

with a 1.92 sec segment we should drop the odd segment, but we will be dropping 1 in 2. It would be better to tell skip fragment what the time of the next displayed time should be and skip segments skips until it has a segment that covers time t

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -58,12 +57,10 @@ class AampTsbSessionManagerTests : public ::testing::Test {
 
 		// Create mocks for the AAMP objects
 		g_mockPrivateInstanceAAMP = new NiceMock<MockPrivateInstanceAAMP>();
```
DomSyna, line 60: good to use smart pointers instead of new for the other mocks ?

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -119,29 +112,16 @@ class AampTsbSessionManagerTests : public ::testing::Test {
 	constexpr static double kDefaultBandwidth{1000000}; // 1Mbps
 };
 
-// Test calling PushNextTsbFragment() for a track that is disabled
-TEST_F(AampTsbSessionManagerTests, TrackDisabled)
-{
```
DomSyna, line 124: doesn't this test push next fails if the track is disabled ?

jfagunde: The code has been modified in https://github.com/rdkcentral/aamp/pull/489/files to not call PushNextTsbFragment() for a track that is disabled. The check that the track is enabled inside PushNextTsbFragment() was removed. I.e. PushNextTsbFragment() is not expected to be called for a track that is disabled. If it was and the next fragment was present in the TSB, it would be injected. So this test does no longer make sense (it was not a good test anyway because it didn't fail after the check for track enabled was removed from the code).

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
jfagunde: I rather not refactor these existing tests like that as part of this Jira. I have already probably done too many changes to these tests.

AampTSBSessionManager.cpp
jfagunde: Are you suggesting to change the algorithm used to skip fragments? I rather not do that as part of this Jira.
Your description of how skipping works at the moment is correct. At a rate x8 and 4fps, the delta is 2s, so i-frames will be presented at a rate not shorter than 2s. With 1.92s duration segments, one every two segments will be dropped, so the rate at which i-frames would be presented is 1.92*2=3.84s. We could maybe drop fewer segments, but that is not the purpose of this Jira.

DomSyna: Yes, I was thinking the method should be provided the time t of the next segment to be presented and do something like

if direction is forward
   while(nextFragmentData  && t > (nextFragmentData->start + duration))
       nextFragmentData = nextFragmentData->next
 else
     do something similar
  
  return nextFragmentData

AampTSBSessionManager.cpp
```diff
@@ -824,14 +824,19 @@ void AampTSBSessionManager::SkipFragment(std::shared_ptr<AampTsbReader>& reader,
 				{
 					tmp = nextFragmentData->prev;
```
DomSyna: does having tmp serve any purpose still ? I'd be tempted to get rid of it and just 
nextFragmentData = nextFragmentData->prev;

jfagunde: Done

AampTSBSessionManager.cpp
```diff
@@ -806,36 +806,52 @@ void AampTSBSessionManager::SkipFragment(std::shared_ptr<AampTsbReader>& reader,
 				delta = static_cast<AampTime>(std::abs(static_cast<double>(rate))) / static_cast<double>(vodTrickplayFPS);
 			}
 			
+			bool skipFragments = true;
```
DomSyna: Using break is better as the variable doesn't need testing on the last iteration of the loop. 

Can line 797 ^ be deleted, no point in setting delta to 0 again.

Is it worth adding && rate to the first if at 785 given no point in doing any of this code if paused ?

AampTSBSessionManager.cpp
```diff
@@ -806,36 +806,52 @@ void AampTSBSessionManager::SkipFragment(std::shared_ptr<AampTsbReader>& reader,
...
 				{
-					tmp = nextFragmentData->prev;
+					skipFragments = false;
 				}
-				if (!tmp)
+				else
 				{
-					break;
+					delta -= fragDuration;
+					skippedDuration += fragDuration;
```
DomSyna: Only seems to be used in the logging, and is only output if a fragment is returned. 
Is this useful given some scenarios it won't be logged ?

AampTSBSessionManager.cpp
```diff
@@ -806,36 +806,52 @@ void AampTSBSessionManager::SkipFragment(std::shared_ptr<AampTsbReader>& reader,
 				delta = static_cast<AampTime>(std::abs(static_cast<double>(rate))) / static_cast<double>(vodTrickplayFPS);
 			}
 			
+			bool skipFragments = true;
 			// Only skip fragments when delta is larger than fragment duration
-			while (delta > 0.0)
+			while (nextFragmentData && (delta > AAMP_RATE_PAUSE) && skipFragments)
```
DomSyna: Don't think this makes sense ? comparing a delta with a speed
delta > AAMP_RATE_PAUSE

AampTSBSessionManager.cpp
```diff
@@ -806,36 +806,52 @@ void AampTSBSessionManager::SkipFragment(std::shared_ptr<AampTsbReader>& reader,
...
+					if ((rate > AAMP_NORMAL_PLAY_RATE) && (nextFragmentData->GetAbsolutePosition().inSeconds() >= mAamp->mTrickModePositionEOS))
+					{
+						AAMPLOG_INFO("Reached live play position during fast forward");
+						skipFragments = false;
+					}
+					else if (rate > AAMP_RATE_PAUSE)
+					{
+						nextFragmentData = nextFragmentData->next;
+					}
+					else if (rate < AAMP_RATE_PAUSE)
```
DomSyna: Can we just have else given for PAUSE delta will be 0 and we won't get in the while ?

AampTSBSessionManager.cpp
```diff
@@ -806,36 +806,52 @@ void AampTSBSessionManager::SkipFragment(std::shared_ptr<AampTsbReader>& reader,
...
+					{
+						if (nextFragmentData->prev)
+						{
+							nextFragmentData = nextFragmentData->prev;
+						}
+						else
+						{
+							AAMPLOG_INFO("Reached beginning of TSB during rewind");
```
DomSyna: Maybe wants a comment explain why we don't want to return null when rewind

AampTSBSessionManager.cpp
jfagunde: Line 797 deleted.
I replaced the calls to break with this boolean to avoid having multiple exit points in the loop.
The logic to check the rate is in PushNextTsbFragment() where this function is called. Would it be cleaner if it is moved inside this function? I would not want to duplicate it.

AampTSBSessionManager.cpp
jfagunde: You are right. Undone ;)

AampTSBSessionManager.cpp
jfagunde: I think is useful. It gives information about how much duration was skipped:
`Line 15018: 2025-09-24T12:26:13.630Z com.sky.as.apps_com.comcast.viper_ipa[3984]:  [AAMP-PLAYER][-1][INFO][b27ba648][SkipFragment][848]Skipped frames [rate=-30.00] from 1758716531.64 to 1758716525.88 delta = 1.74, total duration = 5.76`

AampTSBSessionManager.cpp
jfagunde: Another argument to move the logic to check the rate inside this function?

AampTSBSessionManager.cpp
jfagunde: Done.

AampTSBSessionManager.cpp
jfagunde: I reverted to use break instead of the boolean and moved the rate check inside SkipFragment() function.

AampTSBSessionManager.cpp
jfagunde: Done. I have moved the rate checks to inside this function.

AampTSBSessionManager.cpp
```diff
@@ -777,65 +777,125 @@ AAMPStatusType AampTSBSessionManager::InvokeTsbReaders(double &startPosSec, floa
...
+ */
+AampTime AampTSBSessionManager::CalculateSkipDelta(float rate, int vodTrickplayFPS)
+{
+	AampTime delta = 0.0;
+
+	if (mAamp->playerStartedWithTrickPlay)
+	{
+		AAMPLOG_WARN("Played switched in trickplay, delta set to zero");
+		mAamp->playerStartedWithTrickPlay = false;
+		delta = 0.0;
```
DomSyna: pointless ?

AampTSBSessionManager.cpp
```diff
@@ -777,65 +777,125 @@ AAMPStatusType AampTSBSessionManager::InvokeTsbReaders(double &startPosSec, floa
...
+	}
+	else if (vodTrickplayFPS == 0)
+	{
+		AAMPLOG_WARN("vodTrickplayFPS is zero, delta set to zero");
+		delta = 0.0;
```
DomSyna: pointless ?

AampTSBSessionManager.cpp
```diff
@@ -777,65 +777,125 @@ AAMPStatusType AampTSBSessionManager::InvokeTsbReaders(double &startPosSec, floa
...
+	return success;
+}
+
+/**
+ * @brief Check if fragment skipping should be performed
+ */
+bool AampTSBSessionManager::ShouldSkipFragments(std::shared_ptr<AampTsbReader>& reader, float rate)
+{
+	return (eMEDIATYPE_VIDEO == reader->GetMediaType() &&
+			((AAMP_NORMAL_PLAY_RATE < rate) || (AAMP_RATE_PAUSE > rate)));
```
DomSyna, line 851: (AAMP_NORMAL_PLAY_RATE < rate) || (AAMP_RATE_PAUSE > rate)
Would it be more readable as

(AAMP_NORMAL_PLAY_RATE != rate) && (rate != AAMP_RATE_PAUSE)
which will shortcut more often that then other options since vast majority of time we will be at normal play rate. It may be better to check rate before media type as AAMP_NORMAL_PLAY_RATE != rate will be the norm

AampTSBSessionManager.cpp
```diff
@@ -777,65 +777,125 @@ AAMPStatusType AampTSBSessionManager::InvokeTsbReaders(double &startPosSec, floa
...
 void AampTSBSessionManager::SkipFragment(std::shared_ptr<AampTsbReader>& reader, TsbFragmentDataPtr& nextFragmentData)
 {
+	// Early validation check
 	if (nextFragmentData && reader && !reader->IsEos())
 	{
-		AampTime skippedDuration{};
-		if(eMEDIATYPE_VIDEO == reader->GetMediaType())
+		float rate = reader->GetPlaybackRate();
+
+		if (ShouldSkipFragments(reader, rate))
```
DomSyna, line 864: Could make this first, pass in rate by reference and include the early validation checks ? Not sure

AampTSBSessionManager.cpp
```diff
@@ -777,65 +777,125 @@ AAMPStatusType AampTSBSessionManager::InvokeTsbReaders(double &startPosSec, floa
...
-				{
-					tmp = nextFragmentData->next;
-				}
-				else if (rate < 0.0)
-				{
-					tmp = nextFragmentData->prev;
-				}
-				if (!tmp)
+
+				if (!NavigateToNextFragment(nextFragmentData, rate))
```
DomSyna, line 883: don't like the name, but maybe me

AampTSBSessionManager.cpp
jfagunde: Removed

AampTSBSessionManager.cpp
jfagunde: Deleted

AampTSBSessionManager.cpp
jfagunde: The original check was:
(AAMP_NORMAL_PLAY_RATE != rate && AAMP_RATE_PAUSE != rate && AAMP_SLOWMOTION_RATE != rate)
But I think this is safer, in case they come up with some other slow motion rate.

AampTSBSessionManager.cpp
jfagunde: It was Copilot idea :)
Do you have any suggestion?

AampTSBSessionManager.cpp
jfagunde: I don't think is worth changing it.

pstroffolino (approved):

PR #509, opened 2025-09-23
VPLAY-11042: [VIPA][AAMP TSB] No ad progress reports when resuming normal rate 1x playback after seeking or trick-playing into an ad break

anshephe:
Reason for Change:  If seek (or resume from trickplay) to within an advert reservation period, raise the RESERVATION_START event. This event is required by JSPP to report advert progress.

Test Guidance: Test Ad Progress Reports are raised for complete CDAI Adverts following trickplay

Risk: Low


AampTSBSessionManager.cpp
```diff
@@ -1224,54 +1235,139 @@ void AampTSBSessionManager::ShiftFutureAdEvents()
...
+		std::list<std::shared_ptr<AampTsbAdMetaData>> reservationList = mMetaDataManager.GetMetaDataByType<AampTsbAdMetaData>(
+			AampTsbMetaData::Type::AD_RESERVATION_METADATA_TYPE, reservationRangeStart, reservationRangeEnd);
+
+		// Get AD_PLACEMENT_METADATA_TYPE entries in range
+		std::list<std::shared_ptr<AampTsbAdMetaData>> placementList = mMetaDataManager.GetMetaDataByType<AampTsbAdMetaData>(
+			AampTsbMetaData::Type::AD_PLACEMENT_METADATA_TYPE, placementRangeStart, placementRangeEnd);
+
+		// Remove entries up to and including the last processed for each list
+		// If no last processed, remove from beginning until a START event is found
+		auto skipUpToLast = [](std::list<std::shared_ptr<AampTsbAdMetaData>>& list, const std::shared_ptr<AampTsbMetaData>& lastProcessed)
```
jfagunde, line 1325: Why not use a normal helper function instead of a lambda function?

AampTSBSessionManager.cpp
```diff
@@ -1224,54 +1235,139 @@ void AampTSBSessionManager::ShiftFutureAdEvents()
...
-
-		// Get all ad metadata within the fragment's time range
-		auto adMetadataItems = mMetaDataManager.GetMetaDataByType<AampTsbAdMetaData>(
-			AampTsbMetaData::Type::AD_METADATA_TYPE, rangeStart, rangeEnd);
-
-		// Process metadata items in chronological order
-		bool skip = static_cast<bool>(mLastAdMetaDataProcessed);
-		for (const auto& adMetadata : adMetadataItems)
+		// Sort merged list
+		std::sort(merged.begin(), merged.end(), [](const MetaWithType& a, const MetaWithType& b)
```
jfagunde: Why not use a normal helper function to implement the sort logic? It would simplify AampTSBSessionManager::ProcessAdMetadata()

AampTSBSessionManager.cpp
```diff
@@ -1224,54 +1235,139 @@ void AampTSBSessionManager::ShiftFutureAdEvents()
...
 		{
-			// Skip until we have reached the last processed metadata
-			if (skip)
+			auto apos = a.meta->GetPosition().milliseconds();
+			auto bpos = b.meta->GetPosition().milliseconds();
+
+			// Different positions, sort by position
+			if (apos != bpos) return apos < bpos;
```
jfagunde: I would write the return statement in its own line, with open and closing braces. If I remember correctly, it is part of AAMP coding guidelines. I.e.
```
if (apos != bpos)
{
	return apos < bpos;
}
```
The same comment applies to similar lines below.

AampTSBSessionManager.cpp
```diff
@@ -1224,54 +1235,139 @@ void AampTSBSessionManager::ShiftFutureAdEvents()
...
+
+			// Same position: apply rules
+			auto aType = a.meta->GetEventType();
+			auto bType = b.meta->GetEventType();
+			// Assumption is that reservations and placements don't have 0 duration,
+			// i.e START and END events are not at the same position for the same type
+			if (a.isReservation && !b.isReservation)
+			{
+				// Reservation vs Placement
+				if (aType == AampTsbAdMetaData::EventType::START && bType == AampTsbAdMetaData::EventType::START) return true; 	// Reservation START before Placement START
```
jfagunde: I'd prefer a single exit point for this function. I.e. use a flag and a single return.

jfagunde (commented):
Are there L1 tests for reservation and placement ad events with the same positions?

AampTSBSessionManager.cpp
anshephe: I think it is clearer to have the code in place

AampTSBSessionManager.cpp
anshephe: Moved the merge and sort out to a new function

AampTSBSessionManager.cpp
anshephe: Early returns removed

AampTSBSessionManager.cpp
anshephe: Early returns removed

jfagunde (approved):

pstroffolino (approved):

PR #519, opened 2025-09-26
feature/VPLAY-11187 GetAvailableTextTracks() mark in-band CC tracks as available 

DomSyna:
Currently if AAMP local TSB is enabled all text tracks other than the current text track are marked as not available. With this change the CC tracks are also marked as available.

Test Procedure: L1 tests added to confirm the behaviour of GetAvailableTextTracks()

Risk: Low


priv_aamp.cpp
```diff
@@ -10353,8 +10358,10 @@ std::string PrivateInstanceAAMP::GetAvailableTextTracks(bool allTrack)
 					{
 						cJSON_AddStringToObject(item, "codec", iter->codec.c_str());
 					}
+					// When local TSB is present, availability refers to availability in the TSB
 					bool isAvailable = iter->isAvailable;
-					if (IsLocalAAMPTsb())
+					// Closed Captions are carried in the video stream and hence always available
```
jfagunde, line 10359: Are Closed Captions always in-line and only Closed Captions are in-line?
I.e. is the result of isCC() the same to are subtitles in-line?
Probably a question for Dave.

test/utests/fakes/FakeCJSON.cpp
```diff
@@ -74,50 +84,69 @@ CJSON_PUBLIC(const char *) cJSON_GetErrorPtr(void)
...
-	return nullptr;
+	return g_mockCJsonManager ? g_mockCJsonManager->AddArrayToObject(object, name) : nullptr;
 }
 
 CJSON_PUBLIC(char *) cJSON_Print(const cJSON *item)
 {
-    return (char *)"";
+	if (g_mockCJsonManager) {
+		std::string result = g_mockCJsonManager->Print(item);
+		// Allocate memory with malloc to match real cJSON behavior
```
jfagunde, line 129: Is allocating this memory needed?
Who frees this memory?
The same questions apply to the next function.

test/utests/mocks/MockCJsonManager.h
```diff
@@ -0,0 +1,74 @@
...
+	MOCK_METHOD(std::string, Print, (const cJSON *item));
+	MOCK_METHOD(std::string, PrintUnformatted, (const cJSON *item));
+};
+
+// Global mock instance that fake cJSON functions will delegate to
+extern MockCJsonManager* g_mockCJsonManager;
+
+// Functions to control the mock from tests
+void SetMockCJsonManager(MockCJsonManager* manager);
+MockCJsonManager* GetMockCJsonManager();
```
jfagunde: These functions don't seem to be used or implemented. I'd delete them.

test/utests/tests/PrivateInstanceAAMP/ClosedCaptionTests.cpp
```diff
@@ -0,0 +1,395 @@
...
+		// Set the stream abstraction in the PrivateInstanceAAMP object
+		mPrivateInstanceAAMP->mpStreamAbstractionAAMP = g_mockStreamAbstractionAAMP;
+
+		// Create and set up cJSON mock manager
+		mMockCJsonManager = new MockCJsonManager();
+		g_mockCJsonManager = mMockCJsonManager;
+
+		// Setup the mock data
+		setupMockTextTracks();
+	}	void TearDown() override
```
jfagunde: Please move TearDown() function to a new line.

test/utests/tests/PrivateInstanceAAMP/ClosedCaptionTests.cpp
```diff
@@ -0,0 +1,395 @@
...
+		g_mockStreamAbstractionAAMP = new MockStreamAbstractionAAMP(mPrivateInstanceAAMP);
+		g_mockAampGstPlayer = new MockAAMPGstPlayer(mPrivateInstanceAAMP);
+		g_mockAampStreamSinkManager = new MockAampStreamSinkManager();
+
+		// Set the stream abstraction in the PrivateInstanceAAMP object
+		mPrivateInstanceAAMP->mpStreamAbstractionAAMP = g_mockStreamAbstractionAAMP;
+
+		// Create and set up cJSON mock manager
+		mMockCJsonManager = new MockCJsonManager();
+		g_mockCJsonManager = mMockCJsonManager;
```
jfagunde: Why not use g_mockCJsonManager directly?

priv_aamp.cpp
```diff
@@ -10358,8 +10358,10 @@ std::string PrivateInstanceAAMP::GetAvailableTextTracks(bool allTrack)
...
+					if ((IsLocalAAMPTsb()) && (!iter->isCC))
```
dnfaulkner, line 10360: So did you determine that iter->isAvailable is always true for in-band CC streams?

DomSyna: Yes I think so, we will find out with higher level tests and fix it if not

test/utests/tests/PrivateInstanceAAMP/ClosedCaptionTests.cpp
```diff
@@ -0,0 +1,395 @@
...
+	EXPECT_CALL(*g_mockStreamAbstractionAAMP, GetAvailableTextTracks(false))
+		.WillOnce(ReturnRef(mockTextTracks));
+
+	std::string result = mPrivateInstanceAAMP->GetAvailableTextTracks(false);
+
+	// Assert - The REAL test verification is in the EXPECT_CALL expectations above,
+	// which ensure the correct cJSON structure is built with the right field names and values.
+
+	// We only do basic sanity checks since the mock Print returns minimal JSON
+	EXPECT_FALSE(result.empty()) << "Should return non-empty result when tracks are available";
```
jfagunde, line 275: You could do more checks on the result string, for example that the different languages are in it ("fr", "de", etc). However, the new tests are testing much more than we had before, so happy if you leave this as is if you don't want to spend more time.

priv_aamp.cpp
dnfaulkner: You mean in-band?
Yes, I'm pretty sure that that's what that flag indicates (famous last words!)
I think that the use of it in this code confirms it:
[https://github.com/rdkcentral/aamp/blob/2f0d7fd419b17897daece8e869bf0eee24320b41/priv_aamp.cpp#L10879](url)

test/utests/fakes/FakeCJSON.cpp
DomSyna: Yes as it is deleted in the main code

test/utests/tests/PrivateInstanceAAMP/ClosedCaptionTests.cpp
DomSyna: Um but we would be testing the json mock then, not the real code

priv_aamp.cpp
jfagunde: Great. Yes, I meant in-band :)

test/utests/tests/PrivateInstanceAAMP/ClosedCaptionTests.cpp
jfagunde: You are right. GetAvailableTextTracks() calls cJSON_Print() (which calls the mock Print()) to convert the JSON object to a string, and the test is making that mock return `"[]"` so the string does not contain the actual JSON document. All you could do is set the mock to return a test string and check that the string returned by GetAvailableTextTracks(). Something like:
```
const char kJsonString[] = "Test JSON string";

void setupArrayFinalizationExpectations()
{
    EXPECT_CALL(*g_mockCJsonManager, Print(mockArray))
			.WillOnce(Return(kJsonString));
}

EXPECT_STR_EQ(result.c_str(), kJsonString);
```
But the test is already very thorough as it is.

jfagunde (approved):

cpc005 (approved):

pstroffolino (approved):

PR #523, opened 2025-09-26
VPLAY-10693 Create files with AAMP logging

jfagunde:
Reason for Change: AAMP log lines are being lost when running with VIPA

Test Procedure: Check the AAMP new log files generated

Risk: Low


tsb/src/TsbStore.cpp
```diff
@@ -27,6 +27,7 @@
 #include "TsbSem.h"
 #include "TsbLocationLock.h"
 #include "TsbFs.h"
+#include "FileLogger.h"
```
jfagunde, line 30: This creates a dependency between the TSB library and AAMP :(

middleware/playerLogManager/PlayerLogManager.cpp
```diff
@@ -30,12 +30,16 @@
...
+#include "FileLogger.h"
 
 #ifdef USE_ETHAN_LOG
 #include <ethanlog.h>
 #else
 // stubs for use if USE_ETHAN_LOG not defined
-static void vethanlog(int level, const char *filename, const char *function, int line, const char *format, va_list ap){}
+static void vethanlog(int level, const char *filename, const char *function, int line, const char *format, va_list ap){
+	// Original stub behavior (no-op) plus redirect to our file logger for dual logging
+	FileLogger::getInstance().writeLog(format, ap);
```
jfagunde, line 41: This seems wrong. writeLog() is called after vethanlog(). If USE_ETHAN_LOG is not defined, writeLog() would be called twice for every log line.

middleware/playerLogManager/PlayerLogManager.cpp
```diff
@@ -150,9 +154,11 @@ void logprintf(MW_LogLevel logLevelIndex, const char* file, int line, const char
 					    break;
 			    }
 			    vethanlog(ethanLogLevel,NULL,NULL,-1,format_ptr, args);
+			    FileLogger::getInstance().writeLog(format_ptr, args);
 		    }
 		    else
 		    {
+				FileLogger::getInstance().writeLog(format_ptr, args);
```
jfagunde, line 161: Would it not be better to call this function only once outside the if?
Is there a reason to call it before vethanlog() but after sd_journal_printv()?
Same comments apply to aamplogging.cpp

tsb/src/TsbStore.cpp
```diff
@@ -285,6 +286,8 @@ StoreImpl::StoreImpl(const Store::Config& config, LogFunction logger, int logger
 		throw std::invalid_argument("Failed to create flushDir");
 	}
 
+	FileLogger::setCustomFilename(mLocation);
```
jfagunde, line 289: Would it be better to do this from AAMP so it is done only once and not everytime a TSB session is created?

PR #525, opened 2025-09-26
VPLAY-11189 Map relevant CP Thunder Plugin Error Codes in aamp

varatharajan568:
Reason for Change: Mapped the relevant CP Thunder Plugin Error Codes in AAMP
Test Guidance: Need to check the CP failure scenario
Risk: Low


middleware/externals/contentsecuritymanager/IFirebolt/ContentProtectionFirebolt.cpp
```diff
@@ -30,10 +30,57 @@
...
+#include "PlayerSecInterface.h"
+
 std::condition_variable mConnectionCV;
 std::mutex mConnectionMutex;
 using namespace Firebolt;
-uint64_t ContentProtectionFirebolt::mSubscriptionId = 0; 
+uint64_t ContentProtectionFirebolt::mSubscriptionId = 0;
+
+//Lookup table to convert CPS error to secmanager error
+std::map<const int32_t, std::pair<const int32_t, const int32_t>> ContentProtectionSeManagerErrorLookUp =
```
pstroffolino: Typo?

Should be:
ContentProtectionSecManagerErrorLookUp

Also, can this lookup table me made const static?

pstroffolino (changes requested):
godspell failures

middleware/externals/contentsecuritymanager/IFirebolt/ContentProtectionFirebolt.cpp
varatharajan568: Resolved the typo errors & modified the Lookup table variable as an const static.

PR #532, opened 2025-09-29
VPLAY-11242 [VIPA] Refactor SetPreferredTextLanguages

cpc005:
Refactor PrivateInstanceAAMP::SetPreferredTextLanguages() with no intentional functional changes as a precursor to further changes

New functions:
SavePreferredTextLanguages() 
CheckPreferredTextLanguages()
SetCCFromTextTrack()
FindClosedCaptionTrackIndex()


priv_aamp.h
```diff
@@ -800,6 +812,15 @@ class PrivateInstanceAAMP : public DrmCallbacks, public std::enable_shared_from_
 	 * This function is invoked continuously when ever there is an update in manifest
 	 */
 	void updateManifest(const char *manifestData);
+	/**
+	 * @fn CheckPreferredTextLanguages
+	 * @param[in] trackInfo - Text track information
+	 * @param[out] isSelectionChange true if preferences now select a different track to the current selection
+ 	 * @param[out] isAvailableInManifest true if new selection is available in the manifest
+	 * @param[out] closedCaptionTrackIdx - closed caption track index
+	 * @return void
```
jfagunde: I'd remove this line.

priv_aamp.cpp
```diff
@@ -10978,6 +10949,43 @@ void PrivateInstanceAAMP::SetTextTrack(int trackId, char *data)
 	}
 }
 
+/**
+ * @brief Set closed caption track with appropriate format from passed text track
+ */
+void PrivateInstanceAAMP::SetCCFromTextTrack(TextTrackInfo &track)
+{
+	CCFormat format = eCLOSEDCAPTION_FORMAT_DEFAULT;
```
DomSyna: Wouldn't this be better in the scope of the else as if the track is empty it serves no purpose

priv_aamp.cpp
```diff
@@ -12232,148 +12236,196 @@ void PrivateInstanceAAMP::SetPreferredTextLanguages(const char *param )
...
+void PrivateInstanceAAMP::CheckPreferredTextLanguages(const std::vector<TextTrackInfo> &trackInfo, bool &isAvailableInManifest, bool &isSelectionChange, int &closedCaptionTrackIdx)
+{
 
-			bool languageAvailabilityInManifest = false;
-			bool renditionAvailabilityInManifest = false;
-			bool accessibilityAvailabilityInManifest = false;
-			bool labelAvailabilityInManifest = false;
-			bool nameAvailabilityInManifest = false;
-			bool trackNotEnabled = false;
+	int trackIndex = GetTextTrack();
```
jfagunde: How about renaming this to currentTextTrackIndex or something more meaningful so we don't get confused with trackIdx?

priv_aamp.cpp
```diff
@@ -12232,148 +12236,196 @@ void PrivateInstanceAAMP::SetPreferredTextLanguages(const char *param )
...
+						isAvailableInManifest = true;
 					}
+				}
+				if ((trackLanguage == firstLanguage) && track.isCC && preferredRenditionString != "subtitle")
+				{
+					closedCaptionLanguageTrackIdx = trackIdx;
+				}
 
-					if (preferredTextLanguagesList.size() > 1)
+				// do not understand the following
```
jfagunde: I would remove this comment.

priv_aamp.cpp
```diff
@@ -12232,148 +12236,196 @@ void PrivateInstanceAAMP::SetPreferredTextLanguages(const char *param )
...
-			bool renditionPresent = false;
-			bool accessibilityTypePresent = false;
-			bool labelPresent = false;
-			bool instreamIdPresent = false;
-			int trackIndex = GetTextTrack();
-			bool namePresent = false;
+/**
+ *  @brief Compare text track preferences vs manifest contents vs current selection
+ */
+void PrivateInstanceAAMP::CheckPreferredTextLanguages(const std::vector<TextTrackInfo> &trackInfo, bool &isAvailableInManifest, bool &isSelectionChange, int &closedCaptionTrackIdx)
```
jfagunde, line 12294: Would it be safer to initialise isSelectionChange and isAvailableInManifest to false at the beginning of the function rather than relying on the caller to initialise them before calling this method?

DomSyna, line 12294: Would it be worth grouping the out vars into a struct and return those for readability.....
struct TextLanguageSelectionResult {
    bool isSelectionChange = false;
    bool isAvailableInManifest = false;
    int closedCaptionTrackIdx = -1;
};
 and then 
 TextLanguageSelectionResult CheckPreferredTextLanguages(const std::vector<TextTrackInfo>& trackInfo);

test/utests/tests/PreferredLanguages/SetPreferredTextLanguagesTests.cpp
```diff
@@ -690,7 +700,7 @@ TEST_F(SetPreferredTextLanguagesTests, SetTsbSessionManagerNull)
 	EXPECT_EQ(g_mockTSBSessionManager, nullptr);
 
 	EXPECT_CALL(*g_mockStreamAbstractionAAMP, GetAvailableTextTracks(_))
-		.Times(2).WillRepeatedly(ReturnRef(tracks));
+		.Times(1).WillRepeatedly(ReturnRef(tracks));
```
jfagunde, line 707: You can use .WillOnce()

priv_aamp.cpp
```diff
@@ -12232,148 +12236,196 @@ void PrivateInstanceAAMP::SetPreferredTextLanguages(const char *param )
...
+		char *currentPrefName = const_cast<char *>(trackInfo[trackIndex].name.c_str());
+		Accessibility currentAccessibilityNode = trackInfo[trackIndex].accessibilityItem;
+
+		for (const auto &track : trackInfo)
+		{
+			trackIdx++;
+			// Logic to check whether the given language is present in the available tracks,
+			// if available, it should not match with current preferredLanguagesString, then call tune to reflect the language change.
+			// if not available, then avoid calling tune.
+			if (preferredTextLanguagesList.size() > 0)
```
DomSyna, line 12314: Looks like each of these 5 if blocks could be a helper function to aid readability

priv_aamp.cpp
```diff
@@ -12232,148 +12236,196 @@ void PrivateInstanceAAMP::SetPreferredTextLanguages(const char *param )
...
 					}
 				}
-
 			}
-			else
+		}
+	}
+	else
+	{
+		isSelectionChange = true;
```
DomSyna, line 12390: Maybe this could be a couple of helper functions? 
        closedCaptionTrackIdx = FindClosedCaptionTrackIndexByInstreamId(trackInfo, preferredInstreamIdString);
        if (closedCaptionTrackIdx < 0) {
            closedCaptionTrackIdx = FindClosedCaptionTrackIndexByLanguage(trackInfo, preferredTextLanguagesList[0], preferredRenditionString);
        }

priv_aamp.cpp
cpc005: done

priv_aamp.cpp
cpc005: done

priv_aamp.cpp
cpc005: done

priv_aamp.cpp
cpc005: done

test/utests/tests/PreferredLanguages/SetPreferredTextLanguagesTests.cpp
cpc005: no change

priv_aamp.h
cpc005: done

priv_aamp.cpp
cpc005: done

priv_aamp.cpp
cpc005: no action

priv_aamp.cpp
cpc005: no action

priv_aamp.cpp
```diff
@@ -12074,13 +12083,13 @@ void PrivateInstanceAAMP::SanitizeLanguageList(std::vector<std::string>& languag
...
 /**
- *  @brief Set Preferred Text Language
+ *  @brief Process json object or language string and save the preferred selection to AampConfig
  */
-void PrivateInstanceAAMP::SetPreferredTextLanguages(const char *param )
+void PrivateInstanceAAMP::SavePreferredTextLanguages(const char *param )
 {
 	/**< First argument is Json data then parse it and and assign the variables properly*/
 	AampJsonObject* jsObject = nullptr;
-	bool accessibilityPresent = false;
```
dnfaulkner, line 12084: Should accessibilityPresent really have been deleted? There's mentions of it in UVE-JS API documentation

priv_aamp.cpp
```diff
@@ -12239,148 +12244,200 @@ void PrivateInstanceAAMP::SetPreferredTextLanguages(const char *param )
...
 	SETCONFIGVALUE_PRIV(AAMP_APPLICATION_SETTING,eAAMPConfig_PreferredTextLanguage,preferredTextLanguagesString);
+}
 
-	AAMPPlayerState state = GetState();
-	if (state != eSTATE_IDLE && state != eSTATE_RELEASED && state != eSTATE_ERROR )
-	{ // active playback session; apply immediately
-		if (mpStreamAbstractionAAMP)
+/**
+ *  @brief Find closed caption track index if any
+ */
```
dnfaulkner, line 12257: Need to doxygen the returned int - presumably it's the index into the list of available tracks that matches the preferences or -1 if not found?

priv_aamp.cpp
```diff
@@ -12239,148 +12244,200 @@ void PrivateInstanceAAMP::SetPreferredTextLanguages(const char *param )
...
+		closedCaptionTrackIdx = closedCaptionInstreamIdTrackIdx;
+	}
+	else
+	{
+		closedCaptionTrackIdx = closedCaptionLanguageTrackIdx;
+	}
+	return closedCaptionTrackIdx;
+}
+/**
+ *  @brief Compare text track preferences vs manifest contents vs current selection
```
dnfaulkner, line 12292: Oxygen the parameters and returned information

priv_aamp.h
```diff
@@ -800,6 +818,14 @@ class PrivateInstanceAAMP : public DrmCallbacks, public std::enable_shared_from_
 	 * This function is invoked continuously when ever there is an update in manifest
 	 */
 	void updateManifest(const char *manifestData);
+	/**
```
dnfaulkner, line 823: Maybe I've misread but it looks like this is declared as a public function - isn't it only used internally?

priv_aamp.cpp
```diff
@@ -12429,49 +12481,11 @@ void PrivateInstanceAAMP::SetPreferredTextLanguages(const char *param )
...
-							}
-							else if (track.rendition.find("708") != std::string::npos)
-							{
-								format = eCLOSEDCAPTION_FORMAT_708;
-							}
-						}
-						PlayerCCManager::GetInstance()->SetTrack(track.instreamId, format);
-					}
+					TextTrackInfo track = trackInfo[closedCaptionTrackId];
+					SetCCFromTextTrack(track);
```
dnfaulkner: Interesting that there's no setting of "mIsInbandCC = true" here, whereas there is in SetTextTrack(). I wonder why (it could be done in SetCCFromTextTrack() instead if we thought that it should be set in both cases).

priv_aamp.cpp
```diff
@@ -12239,148 +12244,200 @@ void PrivateInstanceAAMP::SetPreferredTextLanguages(const char *param )
...
+		}
+	}
+	else
+	{
+		isSelectionChange = true;
+		// no track is currently selected but need to find closedCaptionTrackIdx if there is one
+	}
 
-			if((mMediaFormat == eMEDIAFORMAT_HDMI) || (mMediaFormat == eMEDIAFORMAT_COMPOSITE) || (mMediaFormat == eMEDIAFORMAT_OTA) || \
+	closedCaptionTrackIdx = FindClosedCaptionTrackIndex(trackInfo);
```
dnfaulkner, line 12394: Don't you only need to do this if isSelectionChange==true (but then need to set closedCaptionTrackIdx=-1 if false)? I guess that the original code didn't.

priv_aamp.cpp
cpc005: It was moved to CheckPreferredTextLanguages() i could not see a reason for it being handled differently to the other attributes

priv_aamp.cpp
cpc005: Added to the declaration.

priv_aamp.cpp
cpc005: In the declaration .h file

priv_aamp.cpp
cpc005: There should not be a functional change for closed captions at this point. I will make the change later and hopefully it will be trivial.

priv_aamp.cpp
cpc005: Lack a clear understanding about how the 'EPG' should call the following:
PrivateInstanceAAMP::SetTextTrack(...)
PrivateInstanceAAMP::SetPreferredTextLanguages(...)

On skyxione then it looks like SetPreferredTextLanguages() is immediately followed by a call to SetTextTrack() so SetCCFromTextTrack() will get called twice and mIsInbandCC = true does occur

priv_aamp.h
cpc005: Correct. I wanted to L1 test it and this is the easiest way to do it, although some will not like this approach.

jfagunde (approved):
Looks good to me :)

priv_aamp.cpp
dnfaulkner: Hmmm. I don't see why EPG (or AS, or JSPP) is calling SetTextTrack() after SetPreferredTextLanguages() -maybe a workaround to set mIsInbandCC = true!

priv_aamp.h
```diff
@@ -541,13 +541,32 @@ class PrivateInstanceAAMP : public DrmCallbacks, public std::enable_shared_from_
...
+
+	/**
+ 	* @brief Set closed caption track with appropriate format from passed text track
+ 	* @param[in] track - Text track information
+ 	*/
+	void SetCCFromTextTrack(TextTrackInfo &track);
+
+	/**
+	 * @brief  Find closed caption track index if any
+	 * @param[in] track - Text track information
```
dnfaulkner: trackInfo - list of text track informations

priv_aamp.cpp
```diff
@@ -10985,6 +10956,44 @@ void PrivateInstanceAAMP::SetTextTrack(int trackId, char *data)
...
+		}
+
+		// preferredCEA708 overrides whatever we infer from track. USE WITH CAUTION
+		int overrideCfg = GETCONFIGVALUE_PRIV(eAAMPConfig_CEAPreferred);
+		if (overrideCfg != -1)
+		{
+			format = (CCFormat)(overrideCfg & 1);
+			AAMPLOG_WARN("PrivateInstanceAAMP: CC format override present, override format to: %d", format);
+		}
+		AAMPLOG_INFO("patrick instreamId %s format %d", track.instreamId.c_str(), format);
```
dnfaulkner: "patrick"?

priv_aamp.cpp
dnfaulkner: I'm a bit concerned that the new code starting at line 12372 ([https://github.com/rdkcentral/aamp/pull/532/files#diff-f5d793f0f6446b6571ffc1a4e8698a79659b0f6970fc88d4056e4a2c08cf782dR12372](url))
is different to the original code starting at line 12186 ([https://github.com/rdkcentral/aamp/pull/532/files#diff-f5d793f0f6446b6571ffc1a4e8698a79659b0f6970fc88d4056e4a2c08cf782dL12186](url))

priv_aamp.cpp
cpc005: oops, removed

priv_aamp.cpp
cpc005: Previously it was:
SetPreferredTextLanguages( accessibility)
{
    if accessibility != lastAccessibility
           retune =true
    lastAccessibility =  accessibility 
 }
 
 Now it is  
 SetPreferredTextLanguages( accessibility)
{
    if accessibility != accessibility_of_currently_selected_track
           retune =true
 }
 I changed it by mistake. The new approach matches all the other attributes. I have looked back in the history and cannot find a reason for accessibility being different. I will have to change code to deal with accessibility as pre refactor

priv_aamp.h
cpc005: It is not a list, just a single entry for this func. no action

priv_aamp.cpp
cpc005: done

priv_aamp.h
dnfaulkner: Are you sure? It looks like a list to me:
`const std::vector<TextTrackInfo> &trackInfo`
And the function name suggests that it's looking through a list to find a CC track
And in the function itself:
`for (const auto &track : trackInfo)`

priv_aamp.cpp
cpc005: added L1 tests for accessibility

priv_aamp.h
cpc005: FindClosedCaptionTrackIndex() not SetCCFromTextTrack(). Comment updated.

priv_aamp.h
```diff
@@ -541,13 +541,33 @@ class PrivateInstanceAAMP : public DrmCallbacks, public std::enable_shared_from_
...
+
+	/**
+ 	* @brief Set closed caption track with appropriate format from passed text track
+ 	* @param[in] track - Text track information
+ 	*/
+	void SetCCFromTextTrack(TextTrackInfo &track);
+
+	/**
+	 * @brief  Find closed caption track index in list of text tracks
+	 * @param[in] track - Text track information vector
```
dnfaulkner: Sorry to be pedantic:
* @param[in] **trackInfo** - Text track information vector

dnfaulkner (commented):
One last comment :)

priv_aamp.h
cpc005: done

test/utests/fakes/FakeFragmentCollector_MPD.cpp
```diff
@@ -33,7 +33,12 @@ StreamAbstractionAAMP_MPD::~StreamAbstractionAAMP_MPD()
 
 Accessibility StreamAbstractionAAMP_MPD::getAccessibilityNode(AampJsonObject &accessNode)
 {
-   	Accessibility accessibilityNode;
+
+    Accessibility accessibilityNode;
+    if (g_mockStreamAbstractionAAMP_MPD)
+    {
+        return g_mockStreamAbstractionAAMP_MPD->getAccessibilityNode(accessNode);
```
jfagunde: I would assign accessibilityNode instead of return, so the function has a single exit point.

priv_aamp.h
```diff
@@ -544,8 +544,9 @@ class PrivateInstanceAAMP : public DrmCallbacks, public std::enable_shared_from_
 	/**
  	*  @fn Process json object or language string and save the preferred selection to AampConfig
  	*  @param[in] param - language string or json object
+ 	*  @param[in] isSelectionChange - flag to indicate if accessibility has changed
```
jfagunde: param[out]?

priv_aamp.cpp
```diff
@@ -12085,7 +12085,7 @@ void PrivateInstanceAAMP::SanitizeLanguageList(std::vector<std::string>& languag
 /**
  *  @brief Process json object or language string and save the preferred selection to AampConfig
  */
-void PrivateInstanceAAMP::SavePreferredTextLanguages(const char *param )
+void PrivateInstanceAAMP::SavePreferredTextLanguages(const char *param, bool &isSelectionChange )
```
jfagunde, line 12089: I'd initialise isSelectionChange to false at the start of this method, rather than relying on the value passed in by the caller.

priv_aamp.h
jfagunde, line 827: I'd make these parameters [in,out] to make it clear that the value passed in matters.

cpc005: It has to be "in" if it is also "out" a a func param, no change

priv_aamp.cpp
cpc005: I made a similar change before, but then I had to undo it because the value gets 'carried' through 2 functions. No change

priv_aamp.h
cpc005: done

test/utests/fakes/FakeFragmentCollector_MPD.cpp
cpc005: done

priv_aamp.cpp
cpc005: Nothing I can fix

pstroffolino (approved):

PR #536, opened 2025-09-29
VPLAY-11140: Encrypted content support with mp4demux

Vinish100:
Reason for change: Set encrypted caps if stream is encrypted.
Also add the protection meta data to buffer for decryptor.
Add support for encrypted LLD streams with mp4demux.
Test Procedure: Set 'useMp4Demux=true' and ensure encrypted content plays
Risks: Low
Priority: P1


test/aampcli/AampcliPlaybackCommand.cpp
```diff
@@ -1095,10 +1096,21 @@ void PlaybackCommand::parse( const char *path )
...
+						mp4Demux.Parse(ptr,len);
+						int count = mp4Demux.count();
+						for (int i=0; i<count; i++)
+						{
+							AAMPCLI_PRINTF("Sample No:%d, PTR:%p, SIZE:%zu, PTS:%lf, DTS:%lf, DUR:%lf DRM:%d\n",
+									i + 1,
+									mp4Demux.getPtr(i),
+									mp4Demux.getLen(i),
+									mp4Demux.getPts(i),
+									mp4Demux.getDts(i),
```
pstroffolino, line 1109: can defer making any changes here, but appears setting of DTS is optional/not necessary.  rialto's integration with gstreamer is only setting pts ("aka timestamp").  Unlike PTS, DTS always increments by duration with each new sample, so can be automatically derived under the hood.

test/gstTestHarness/mp4demux.hpp
```diff
@@ -126,6 +126,8 @@ class Mp4Demux
 	uint32_t width_fixed;
 	uint32_t height_fixed;
 	uint16_t language;
+	uint32_t sampleOffset;
+	bool sencPresent;
```
pstroffolino, line 130: Coverity warning that this field is not initialized

test/aampcli/AampcliPlaybackCommand.cpp
Vinish100: agreed Phil. But I suppose lets keep the DTS logic, for non-rialto usecases like third party apps like Charter/Sling/Peacock.

test/gstTestHarness/mp4demux.hpp
Vinish100: Fixed

pstroffolino (approved):

PR #539, opened 2025-09-30
VPLAY-10142: Observed Crash with fingerprint "42622052"

psiva01:
Reason for change: Replaced the current glib malloc calls with non-aborting versions to avoid crash on malloc failure. Updated L1 tests accordingly
Test Procedure: Refer Jira
Risks: No


pstroffolino (changes requested):
I get the idea here, but skeptical it will give any relief.
Today, if memory allocation somehow fails (despite swap space, etc.) we at least get an assert and stack trace.
With this change, we'll hid these failures, but won't necessarily fix anything.
Might make true RCA harder to fine.
Better to understand why the calls failed in first place (if possible).  Could be memory corruption or some other serious issue.

PR #554, opened 2025-10-07
VPLAY-11246: [VIPA] Change the pipeline configuration when Rialto sink is enabled for inband captions with VIPA/Rialto

anshephe:
Reason For Change:

If OOB subtitles are not the selected format then the GStreamer Subtitles pipeline is constructed with just an App source and Rialto Subtitle Sink plugin and caps set to "application/x-subtitle-cc". The pipeline is only used to signal control information. No data is passed through the pipeline, so it only needs to include the Rialto Subtitle Sink.

Risk: Low

Signed-off-by: anshephe <115161257+anshephe@users.noreply.github.com>


middleware/InterfacePlayerRDK.cpp
```diff
@@ -312,6 +314,17 @@ void InterfacePlayerRDK::ConfigurePipeline(int format, int audioFormat, int auxF
...
 	{
 		interfacePlayerPriv->gstPrivateContext->usingRialtoSink = true;
+
+		// If no subtitles defined, then create a closed caption control stream
+		newClosedCaptionsControl = (gstSubFormat == GST_FORMAT_INVALID);
+
+		// However to avoid out of band subtitles being removed during trickplay, 
+		// check if OB subtitles previously configured.
+		// However this check makes it impossible to switch from OB to IB
+		newClosedCaptionsControl = newClosedCaptionsControl &&
```
DomSyna: maybe ?
newClosedCaptionsControl &= (interfacePlayerPriv->gstPrivateContext->stream[eGST_MEDIATYPE_SUBTITLE].format == GST_FORMAT_INVALID);

DomSyna (approved):
Trivial comment / observation. mostly taste. Maybe the comment in the same place could be made clearer / easier to read.

PR #559, opened 2025-10-09
VPLAY-10999:[AAMP] Profile playback stop operation

varshnie:
Reason for change:fragmentcollectorthread optimization Test Procedure: Refer jira ticket
Priority: P1

### DIFF
--- a/priv_aamp.cpp
+++ b/priv_aamp.cpp
@@ -704,14 +704,7 @@ size_t PrivateInstanceAAMP::HandleSSLWriteCallback ( char *ptr, size_t size, siz
 	}
 	else
 	{
-		if(ISCONFIGSET_PRIV(eAAMPConfig_EnableCurlStore) && mOrigManifestUrl.isRemotehost)
-		{
-			ret = (size*nmemb);
-		}
-		else
-		{
-			AAMPLOG_WARN("CurlTrace write_callback - interrupted, ret:%zu", ret);
-		}
+		AAMPLOG_WARN("CurlTrace write_callback - interrupted, ret:%zu", ret);
 	}
 	return ret;
 }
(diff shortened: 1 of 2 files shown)


priv_aamp.cpp
```diff
@@ -1099,7 +1099,7 @@ int PrivateInstanceAAMP::HandleSSLProgressCallback ( void *clientp, double dltot
 	if(rc)
 	{
 		if( !( eCURL_ABORT_REASON_LOW_BANDWIDTH_TIMEDOUT == context->abortReason || eCURL_ABORT_REASON_START_TIMEDOUT == context->abortReason ||\
-			eCURL_ABORT_REASON_STALL_TIMEDOUT == context->abortReason ) && (ISCONFIGSET_PRIV(eAAMPConfig_EnableCurlStore) && mOrigManifestUrl.isRemotehost ) )
+			eCURL_ABORT_REASON_STALL_TIMEDOUT == context->abortReason ) && (ISCONFIGSET_PRIV(eAAMPConfig_EnableCurlStore) && mOrigManifestUrl.isRemotehost ) && context->aamp->mDownloadsEnabled )
```
Vinish100: rc = -1 is only set when (!context->aamp->mDownloadsEnabled && context->aamp->mMediaDownloadsEnabled[context->mediaType]), so this statement is contradictory. I would recommend removing the whole block and understand its impact.

fragmentcollector_mpd.cpp
```diff
@@ -10827,7 +10827,12 @@ void StreamAbstractionAAMP_MPD::Stop(bool clearChannelData)
 	}
 	if (!aamp->DownloadsAreEnabled() && fragmentCollectorThreadID.joinable())
 	{
+		auto t0 = std::chrono::steady_clock::now();
+		AAMPLOG_WARN("Waiting for fragmentCollectorThreadID Join...");
```
Vinish100: This log has any use?

fragmentcollector_mpd.cpp
```diff
@@ -10827,7 +10827,12 @@ void StreamAbstractionAAMP_MPD::Stop(bool clearChannelData)
...
 		fragmentCollectorThreadID.join();
+		auto t1 = std::chrono::steady_clock::now();
+		AAMPLOG_WARN("fragmentCollectorThreadID join took %lld ms",
```
Vinish100: AAMPLOG_MIL

fragmentcollector_mpd.cpp
varshnie: No just a print where fragmentcollector wait starts.

fragmentcollector_mpd.cpp
varshnie: Done

fragmentcollector_mpd.cpp
Vinish100: Lets remove it. The time elapsed would be available in the following log statement

priv_aamp.cpp
```diff
@@ -704,14 +704,7 @@ size_t PrivateInstanceAAMP::HandleSSLWriteCallback ( char *ptr, size_t size, siz
...
 	{
-		if(ISCONFIGSET_PRIV(eAAMPConfig_EnableCurlStore) && mOrigManifestUrl.isRemotehost)
-		{
-			ret = (size*nmemb);
-		}
-		else
-		{
-			AAMPLOG_WARN("CurlTrace write_callback - interrupted, ret:%zu", ret);
-		}
+		AAMPLOG_WARN("CurlTrace write_callback - interrupted, ret:%zu", ret);
```
pstroffolino: why are we printing ret here?  will always be zero if reached, no?

pstroffolino (approved):

PR #561, opened 2025-10-09
VPLAY-10007: Fix for Coverity Issues (DEADCODE, DIVIDE_BY_ZERO, INTEGER_OVERFLOW, MISSING_COMMA, MISSING_BREAK)

srikanthreddybijjam-comcast:
Reason for change: Resolved More Coverity Issues caused by (DEADCODE, DIVIDE_BY_ZERO, INTEGER_OVERFLOW, MISSING_COMMA, MISSING_BREAK)
Test Procedure: Refer Jira ticket VPLAY-10007
Priority: P1

### DIFF
--- a/downloader/AampCurlStore.cpp
+++ b/downloader/AampCurlStore.cpp
@@ -232,17 +232,7 @@ static int eas_curl_debug_callback(CURL *handle, curl_infotype type, char *data,
 		size_t len = size;
 		while( len>0 && data[len-1]<' ' ) len--;
 		std::string printable(data,len);
-		switch (type)
-		{
-		case CURLINFO_TEXT:
-			AAMPLOG_WARN("curl: %s", printable.c_str() );
-			break;
-		case CURLINFO_HEADER_IN:
-			AAMPLOG_WARN("curl header: %s", printable.c_str() );
-			break;
-		default:
-			break; //CID:94999 - Resolve deadcode
-		}
+		AAMPLOG_MIL("curl debug type:%d info:%s", type, printable.c_str() );
 	}
 	return 0;
 }
--- a/middleware/InterfacePlayerRDK.cpp
+++ b/middleware/InterfacePlayerRDK.cpp
@@ -4646,6 +4646,7 @@ static gboolean buffering_timeout (gpointer data)
 			}
 			GstMediaFormat mediaFormatRet;
 			mediaFormatRet = (GstMediaFormat)pInterfacePlayerRDK->m_gstConfigParam->media;
+			uint32_t original_buffering_timeout_cnt = privatePlayer->gstPrivateContext->buffering_timeout_cnt;
 			/* Disable re-tune on buffering timeout for DASH as unlike HLS,
 			 * DRM key acquisition can end after injection, and buffering is not expected
 			 * to be completed by the 1 second timeout
@@ -4661,7 +4662,7 @@ static gboolean buffering_timeout (gpointer data)
 			else if (frames == -1 || frames >= pInterfacePlayerRDK->m_gstConfigParam->framesToQueue || privatePlayer->gstPrivateContext->buffering_timeout_cnt-- == 0)
 			{
 				MW_LOG_MIL("Set pipeline state to %s - buffering_timeout_cnt %u  frames %i",
-				gst_element_state_get_name(privatePlayer->gstPrivateContext->buffering_target_state), (privatePlayer->gstPrivateContext->buffering_timeout_cnt+1), frames);
+				gst_element_state_get_name(privatePlayer->gstPrivateContext->buffering_target_state), original_buffering_timeout_cnt, frames);
 				SetStateWithWarnings (privatePlayer->gstPrivateContext->pipeline, privatePlayer->gstPrivateContext->buffering_target_state);
 				isRateCorrectionDefaultOnPlaying =  privatePlayer->socInterface->SetRateCorrection();
 				
--- a/middleware/InterfacePlayerRDK.cpp
+++ b/middleware/InterfacePlayerRDK.cpp
@@ -4646,7 +4646,6 @@ static gboolean buffering_timeout (gpointer data)
 			}
 			GstMediaFormat mediaFormatRet;
 			mediaFormatRet = (GstMediaFormat)pInterfacePlayerRDK->m_gstConfigParam->media;
-			uint32_t original_buffering_timeout_cnt = privatePlayer->gstPrivateContext->buffering_timeout_cnt;
 			/* Disable re-tune on buffering timeout for DASH as unlike HLS,
 			 * DRM key acquisition can end after injection, and buffering is not expected
 			 * to be completed by the 1 second timeout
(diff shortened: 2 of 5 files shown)


downloader/AampCurlStore.cpp
```diff
@@ -231,16 +231,13 @@ static int eas_curl_debug_callback(CURL *handle, curl_infotype type, char *data,
 		size_t len = size;
 		while( len>0 && data[len-1]<' ' ) len--;
 		std::string printable(data,len);
-		switch (type)
+		if(type == CURLINFO_TEXT)
```
pstroffolino: would it make sense to just combine these log lines to below?

AAMPLOG_WARN("curl: %s", printable.c_str() );

middleware/InterfacePlayerRDK.cpp
```diff
@@ -4543,8 +4543,10 @@ static gboolean buffering_timeout (gpointer data)
 			}
 			else if (frames == -1 || frames >= pInterfacePlayerRDK->m_gstConfigParam->framesToQueue || privatePlayer->gstPrivateContext->buffering_timeout_cnt-- == 0)
 			{
+				// Store the original count before it was decremented for accurate logging
+				uint32_t originalCount = (privatePlayer->gstPrivateContext->buffering_timeout_cnt == UINT_MAX) ? 0 : privatePlayer->gstPrivateContext->buffering_timeout_cnt + 1;
```
pstroffolino: if we want original count, use copy of original buffering_timeout_cnt
don't subtract then add again

priv_aamp.cpp
```diff
@@ -13281,11 +13281,6 @@ long PrivateInstanceAAMP::LoadFogConfig()
 			jsondataForPreference.add("audio", audioPreference);
 			trackAdded = true;
 		}
-		if(tPrefAvail)
-		{
-			jsondataForPreference.add("text", subtitlePreference);
```
pstroffolino, line 13323: if we're going to get rid of this, should remove tPrefAvail=false declaration above, too.

Are we sure this wasn't wrongly left assigned to false?

test/aampcli/Aampcli.cpp
```diff
@@ -735,6 +735,7 @@ void MyAAMPEventListener::Event(const AAMPEventPtr& e)
 		{
 			MonitorAVStatusEventPtr ev = std::dynamic_pointer_cast<MonitorAVStatusEvent>(e);
 			AAMPCLI_PRINTF("[AAMPCLI] AAMP_EVENT_MONITORAV_STATUS\tstatus=%s\tvposition =%" PRId64 "\taposition=%" PRId64 "\ttimeInStateMS= %" PRIu64 "\tdroppedFrames= %" PRIu64 "\n", ev->getMonitorAVStatus().c_str(), ev->getVideoPositionMS(), ev->getAudioPositionMS(), ev->getTimeInStateMS(),ev->getDroppedFrames());
+			break;
```
pstroffolino, line 738: presumably this missing break was causing bogus logging in aampcli

srikanthreddybijjam-comcast: Earlier it was causing bogus logging but now Fixed it with break.

priv_aamp.cpp
srikanthreddybijjam-comcast: Not implemented any Logic for text/subtitle tracks, just declared for future Implementation.
Currently we can treat it as DeadCode and remove it.

middleware/InterfacePlayerRDK.cpp
srikanthreddybijjam-comcast: Copied the original value prior to decrement.

downloader/AampCurlStore.cpp
srikanthreddybijjam-comcast: Combined those two logs as a better fix like 
AAMPLOG_WARN("curl/curl header: %s", printable.c_str() );

downloader/AampCurlStore.cpp
```diff
@@ -232,17 +232,7 @@ static int eas_curl_debug_callback(CURL *handle, curl_infotype type, char *data,
...
-		case CURLINFO_TEXT:
-			AAMPLOG_WARN("curl: %s", printable.c_str() );
-			break;
-		case CURLINFO_HEADER_IN:
-			AAMPLOG_WARN("curl header: %s", printable.c_str() );
-			break;
-		default:
-			break; //CID:94999 - Resolve deadcode
-		}
+		AAMPLOG_WARN("curl/curl header: %s", printable.c_str() );
```
Vinish100: What was the issue here?

srikanthreddybijjam-comcast: unwanted switch case instead we can use simple if-else or, combined those two logs with out any further validation.

Vinish100: Recommend adding the type to the log for better clarity.
"curl/curl header:" with "Debug type:%d info:%s"

downloader/AampCurlStore.cpp
```diff
@@ -232,17 +232,7 @@ static int eas_curl_debug_callback(CURL *handle, curl_infotype type, char *data,
...
-		case CURLINFO_TEXT:
-			AAMPLOG_WARN("curl: %s", printable.c_str() );
-			break;
-		case CURLINFO_HEADER_IN:
-			AAMPLOG_WARN("curl header: %s", printable.c_str() );
-			break;
-		default:
-			break; //CID:94999 - Resolve deadcode
-		}
+		AAMPLOG_WARN("curl debug type:%d info:%s", type, printable.c_str() );
```
jfagunde: Could you use AAMPLOG_MIL instead of WARN? This is not a warning.

middleware/InterfacePlayerRDK.cpp
```diff
@@ -4661,7 +4662,7 @@ static gboolean buffering_timeout (gpointer data)
 			else if (frames == -1 || frames >= pInterfacePlayerRDK->m_gstConfigParam->framesToQueue || privatePlayer->gstPrivateContext->buffering_timeout_cnt-- == 0)
 			{
 				MW_LOG_MIL("Set pipeline state to %s - buffering_timeout_cnt %u  frames %i",
-				gst_element_state_get_name(privatePlayer->gstPrivateContext->buffering_target_state), (privatePlayer->gstPrivateContext->buffering_timeout_cnt+1), frames);
+				gst_element_state_get_name(privatePlayer->gstPrivateContext->buffering_target_state), original_buffering_timeout_cnt, frames);
```
jfagunde, line 4665: This change is what causes the change in L2 test 2044, not the missing break.

srikanthreddybijjam-comcast: Yes, as we are now using the Original buffer value now.

jfagunde (approved):

downloader/AampCurlStore.cpp
srikanthreddybijjam-comcast: Changed.

pstroffolino (approved):

PR #580, opened 2025-10-14
VPLAY-11315 Avoid calling GStreamer subtec sink mute twice when CC muted

jfagunde:
Reason for Change: Double subtec sink mute when CC muted
Test Guidance: No regressions muting in-band or out-of-band subtitles
Risk: Low


priv_aamp.cpp
```diff
@@ -11032,17 +11032,24 @@ void PrivateInstanceAAMP::SetCCStatusInternal(void)
 	AcquireStreamLock();
 	// Mute subtitles if either video is muted or subtitles are muted
 	bool mute_subtitles_applied = video_muted || subtitles_muted;
-	PlayerCCManager::GetInstance()->SetStatus(!mute_subtitles_applied);
-
-	if (mpStreamAbstractionAAMP)
+	AAMPLOG_TRACE("mIsInbandCC %d, GstSubtecEnabled %d, mute_subtitles_applied %d",
```
DomSyna: would it be worth adding video_muted and subtitles_muted  ?

test/utests/tests/PrivateInstanceAAMP/SubtitleMuteTests.cpp
```diff
@@ -52,7 +53,20 @@ class SubtitleMuteTests : public testing::TestWithParam< std::pair<bool, bool> >
 			gpGlobalConfig =  new AampConfig();
 		}
 
+		// Set up mock config to control eAAMPConfig_GstSubtecEnabled
+		g_mockAampConfig = new NiceMock<MockAampConfig>();
```
DomSyna, line 57: Smart ptr ? given its declaration is a global should we be doing something like:
std::shared_ptr<NiceMock<MockAampConfig>> mMockConfigPtr;

		mMockConfigPtr = std::make_unique<NiceMock<MockAampConfig>>();
		g_mockAampConfig = mMockConfigPtr.get();


		mMockConfigPtr.reset();
		g_mockAampConfig = nullptr;

Well unique ptr would be better but then you need the ctor, maybe add one to a common file ?
Changing the tests like this a bit at a time will mean eventually we can change the type of g_mockAampConfig and move completely to a smart pointer ?

DomSyna (approved):
I've raised a couple of comments, but they are minor so given +1 incase you don't want to take them on board. We need to look at moving the cpp version forward to have unique ptr

test/utests/tests/PrivateInstanceAAMP/SubtitleMuteTests.cpp
jfagunde: I think that makes the code harder to understand. To use a smart pointer for a mock, I'd do it in one go in a PR specifically for this.

priv_aamp.cpp
jfagunde: Done.

pstroffolino (approved):

PR #588, opened 2025-10-16
VPLAY-11442 : [Update UVE Doc] Ability to configure + / - av sync thr…

Gnanesha:
…eshold values for MonitorAV.

Reason for change: added documentation for monitorAV sync
Test Procedure: No Test required
Priority: P1


pstroffolino (changes requested):
Documentation update needs improvement - as described, use and semantics are not clear at all.

One of these is threshold for audio leading (before) video to be considered unacceptable.

One of these is threshold for audio lagging (late) video to be considered unacceptable.

Just saying "positive" or "negative" doesn't tell me anything as a user what to fill out or what it means.

Also confusing because a user might try and fill in an actual negative number - both values are (per my understanding) non-negative configurations.

PR #590, opened 2025-10-16
VPLAY-11245: [VIPA] Add CC Manager class for Rialto

kd-syna:
Reason For Change:

Add CC manager class for when Rialto is being used to render closed captions via a GStreamer Subtitle pipeline. When Rialto is enabled, an instance of this class is created to control the CC (mute/unmute/track selection) instead of a Subtec CC manager.

Risk: Low


middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ *  @file PlayerRialtoCCManager.cpp
+ *
+ *  @brief Impl of Rialto ClosedCaption integration layer
+ *
+ */
+#include "PlayerRialtoCCManager.h"
+#include "PlayerLogManager.h" // Included for MW_LOG
+#include "InterfacePlayerRDK.h" // Included for MW_LOG
```
DomSyna: I didn't spot why we needed this include, and is the comment correct ?

middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
```diff
@@ -0,0 +1,203 @@
...
+	bool changedHandle = (handle != mSubtitleControlHandle);
+
+	mSubtitleControlHandle = handle;
+
+	if (GetTrack().empty())
+	{
+		// Apps expect to render default CC as CC1, so set that here in case
+		// they do not explicitly call SetTrack().
+		MW_LOG_INFO("PlayerRialtoCCManager::Setting default to \"CC1\"");
+		(void) SetTrack("CC1");
```
DomSyna, line 45: guessing it doesn't hurt, but aamp does this already in tune helper by calling RestoreCC

middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+	return;
+}
+
+/**
+ *  @brief Set CC track
+ */
+int PlayerRialtoCCManager::SetTrack(const std::string &track, const CCFormat format)
+{
+	mTrack = track;	// For PlayerCCManager::GetTrack()
```
DomSyna, line 111: Would it hurt to wrap this in 
if (mTrack != track) {
?

middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
```diff
@@ -0,0 +1,203 @@
...
+	return;
+}
+
+/* NOTE WELL: SetDigitalChannel() and SetAnalogChannel() should never be
+** called as they are only called from the base class implementation of
+** SetTrack(), which we override.
+**
+** However, they are declared pure virtual in the base class, so we need
+** these stubs to statisfy that.
+** Further, their return code is strictly an enum which is subtec-specific
```
DomSyna, line 169: Raise a tech dept ticket ?

middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
```diff
@@ -0,0 +1,203 @@
...
+*/
+
+/**
+ * @fn SetDigitalChannel
+ *
+ * @return CC_VL_OS_API_RESULT
+ */
+int PlayerRialtoCCManager::SetDigitalChannel(unsigned int id)
+{
+	MW_LOG_WARN("PlayerRialtoCCManager::Should not be called! (%u)", id);
```
DomSyna: ERR not WARN ?

middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/**
+ * @fn SetAnalogChannel
+ *
+ * @return CC_VL_OS_API_RESULT
+ */
+int PlayerRialtoCCManager::SetAnalogChannel(unsigned int id)
+{
+	MW_LOG_WARN("PlayerRialtoCCManager::Should not be called! (%u)", id);
```
DomSyna: ERROR ?

middleware/closedcaptions/PlayerCCManager.cpp
```diff
@@ -819,15 +825,56 @@ PlayerCCManagerBase *PlayerCCManager::GetInstance()
...
+	mTrickplayStarted = false;
+	mParentalCtrlLocked = false;
+}
+
+/**
+ *  @brief Set the variant required
+ */
+void PlayerCCManager::SetRialto(bool bIsRialto)
+{
+	if (mIsRialto == bIsRialto || mInstance == NULL)
```
DomSyna: Not sure I understand this if condition ?

middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
kd-syna: Done

middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
kd-syna: Done

middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
kd-syna: Sure. I was wary of doing anything to the base class that could have unintended consequences for the Subtec class, so I left it as it was for this ticket.

middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
kd-syna: I'm not convinced it's worth it. These track identifiers are short (a handful of bytes at most) aren't they? Unless unnecessary writes to RAM are an issue, comparing them first seems unnecessary given that we _always_ want them to be the same.

middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
kd-syna: My intention was to follow what the Subtec class does as much as possible so that the Rialto class presents behavior consistent with that to their common caller.  Subtec explicitly sets this default, so Rialto does too.

middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
kd-syna: The comment is _definitely_ wrong! Removed the #include.

middleware/closedcaptions/PlayerCCManager.cpp
kd-syna: It means "If the request is valid". Setting the manager class variant type to what it is already set to is fine (even if it has already been created) and so is setting it to _anything_ if it _hasn't_ yet been created (the pointer is NULL).

The inverse (setting it to something it is not when it already exists) is the error branch.

If we don't care about the logging when we're setting it to what it already is, then it can be made clearer so I'll do that.

middleware/closedcaptions/PlayerCCManager.cpp
```diff
@@ -819,15 +825,56 @@ PlayerCCManagerBase *PlayerCCManager::GetInstance()
...
 	}
 	return mInstance;
 }
 
+/**
+ *  @brief Reset the state.
+ */
+void PlayerCCManagerBase::ResetState()
+{
+	MW_LOG_INFO("PlayerRialtoCCManagerBase::Resetting");
```
jfagunde: PlayerCCManagerBase instead of PlayerRialtoCCManagerBase

middleware/closedcaptions/PlayerCCManager.h
```diff
@@ -239,7 +239,16 @@ class PlayerCCManagerBase
 	 */
 	void Stop();
 
-	
+	/**
+	 * @fn ResetState
+	 *
+	 * @return void
```
jfagunde, line 245: I'd remove the `@return void` comments.

priv_aamp.cpp
```diff
@@ -1303,6 +1303,9 @@ PrivateInstanceAAMP::PrivateInstanceAAMP(AampConfig *config) : mReportProgressPo
 	// Create the CMCD collector
 	mCMCDCollector = new AampCMCDCollector();
 
+	// Ensure the correct CC variant class will be used
+	PlayerCCManager::SetRialto(GETCONFIGVALUE_PRIV(eAAMPConfig_useRialtoSink));
```
jfagunde, line 1307: You could use ISCONFIGSET_PRIV() instead of GETCONFIGVALUE_PRIV(). That is what is normally used and the reason why there were no mocks for this.

middleware/closedcaptions/rialto/PlayerRialtoCCManager.cpp
DomSyna: I was thinking more to stop the call to g_object_set with the same CC as is currently being used as this then breaks rialto

priv_aamp.cpp
DomSyna: I don't think you should write the main code based on what's supported in the mocks. Don't think it matters which you use, otherwise one should be marked as deprecated to encourage the use of the other.

middleware/closedcaptions/PlayerCCManager.h
DomSyna: maintaining same style as rest of file

jfagunde (approved):

DomSyna (approved):

pstroffolino (approved):

PR #591, opened 2025-10-17
VPLAY-11435 Switch to play when the beginning of the TSB is reached

jfagunde:
Reason for Change: Rewind to BoS can cause jump in position of minutes
Test Guidance: Rewind to BoS with full TSB and check transition to play
Risk: Medium


priv_aamp.cpp
```diff
@@ -2119,10 +2119,12 @@ void PrivateInstanceAAMP::ReportProgress(bool sync, bool beginningOfStream)
 			//AAMPLOG_WARN("aamp clamp end");
 			position = end;
 		}
-		else if (position < start)
-		{ // clamp start
-			AAMPLOG_WARN( "clamp position %fms < start %fms", position, start );
+		else if (position < start || (beginningOfStream && rate < AAMP_RATE_PAUSE))
```
DomSyna: when will position be >= start and beginningOfStream == true ?
Isn't position always 0 at this point if beginningOfStream

priv_aamp.cpp
```diff
@@ -2119,10 +2119,12 @@ void PrivateInstanceAAMP::ReportProgress(bool sync, bool beginningOfStream)
...
+		{ // clamp start or handle BOS during rewind
+			AAMPLOG_MIL("Reached BoS position %fms < start %fms beginningOfStream %d rate %f",
+				position, start, beginningOfStream, rate);
 			position = start;
+			HandleBeginningOfStreamReached();
```
DomSyna: Previously this was done before calling report position with true for BoS, how does doing it here fix the issue ?

test/utests/tests/PrivAampTests/PrivAampTests.cpp
```diff
@@ -966,6 +966,50 @@ TEST_F(PrivAampTests,ReportProgressTest6)
 	p_aamp->ReportProgress(sync,beginningOfStream);
 }
 
+/**
+ * @brief Test ReportProgress when rewinding reaches the beginning of the TSB.
+ * Verifies that HandleBeginningOfStreamReached() is called and rate is reset to normal play.
+ */
+TEST_F(PrivAampTests, ReportProgressRewindToBeginningOfTSB)
```
DomSyna: can we have a test for position >= start and BoS is reached so the second part of the || is executed in 
		else if (position < start || (beginningOfStream && rate < AAMP_RATE_PAUSE))
:)

priv_aamp.cpp
jfagunde: Yes, but I am not sure if start can be 0. I just realized that in the code, ReportProgress() is only called with beginningOfStream=true if rate < 0, so I can remove that condition. I.e.
`else if (position < start || beginningOfStream)`
Another possibility could be
`else if (position <= start)`

priv_aamp.cpp
jfagunde: Previously this was done in NotifyEOSReached() and ReportProgress() was called as part of this sequence. Now this is done inside ReportProgress()
This change solves the problem by detecting as soon as possible (ReportProgress() is called very frequently) that the current position being played is smaller than the current beginning of the TSB. Previously, AAMP would only realize after getting EOS notification from GStreamer, after it played all the i-frames injected (i.e. until it reached the BoS when rewind started).

dnfaulkner: I think that the function name is a bit ambiguous in describing what action it takes. I think that         **SeekToTSBStart()** would be clearer.

I think that function names that hint what they're doing rather than what condition they're handling (except for event handlers) is easier to read in unfamiliar code.

priv_aamp.cpp
```diff
@@ -3254,6 +3256,30 @@ int PrivateInstanceAAMP::GetCurrentAudioTrackId()
 	return trackId;
 }
 
+/**
+ * @brief Handle beginning of stream reached during rewind
+ */
+void PrivateInstanceAAMP::HandleBeginningOfStreamReached()
```
dnfaulkner: See above for suggested function rename

priv_aamp.cpp
```diff
@@ -3254,6 +3256,30 @@ int PrivateInstanceAAMP::GetCurrentAudioTrackId()
...
+{
+	if (rate < AAMP_RATE_PAUSE)
```
dnfaulkner: You don't need to test this again - you only call it above if line 2122 check has passed

priv_aamp.cpp
```diff
@@ -3254,6 +3256,30 @@ int PrivateInstanceAAMP::GetCurrentAudioTrackId()
...
+	{
+		seek_pos_seconds = culledSeconds;
+		AAMPLOG_MIL("Updated seek_pos_seconds %f on BOS", seek_pos_seconds);
```
dnfaulkner: "BOS" is a bit cryptic - "TSB Start" or "Start of TSB" instead?

priv_aamp.cpp
```diff
@@ -2119,10 +2119,12 @@ void PrivateInstanceAAMP::ReportProgress(bool sync, bool beginningOfStream)
 			//AAMPLOG_WARN("aamp clamp end");
 			position = end;
 		}
-		else if (position < start)
-		{ // clamp start
-			AAMPLOG_WARN( "clamp position %fms < start %fms", position, start );
+		else if (position < start || (beginningOfStream && rate < AAMP_RATE_PAUSE))
+		{ // clamp start or handle BOS during rewind
+			AAMPLOG_MIL("Reached BoS position %fms < start %fms beginningOfStream %d rate %f",
```
dnfaulkner: "BoS" is a bit cryptic - "TSB Start" or "Start of TSB" instead?
Maybe say "Rewound past start of TSB" instead?

dnfaulkner (commented):
Some "grammatical" suggestions ;)

test/utests/tests/PrivAampTests/PrivAampTests.cpp
jfagunde: I have changed the condition to `else if (position < start || beginningOfStream)`. The second part of that condition is already tested in NotifyBOSReachedREWSeekPositionCalculation, but I'll add a new test similar to this as different things are verified.

priv_aamp.cpp
jfagunde: As discussed, `else if (position <= start)` would be risky because position could be = start when tuning. Using `else if (position < start || beginningOfStream)` instead.

priv_aamp.cpp
jfagunde: Changed to `Reached start of TSB`

priv_aamp.cpp
jfagunde: Renamed to PlayFromTsbStart()

priv_aamp.cpp

priv_aamp.cpp
jfagunde: This is needed.
ReportProgress() is called several times per second. Without this check, it is called when rewinding to the start of the TSB and then it keeps being called and flushing the pipeline and starting the playback in a loop because there is not enough time for the position to be updated.

priv_aamp.cpp
jfagunde: Changed (note that this was an existing log line).

priv_aamp.cpp
dnfaulkner: I'd move the check to where it's called from then. The code flows more logically then and the function does exactly what it's called.

priv_aamp.cpp
dnfaulkner: See comment on line 3264 and move the check to here so that you only call if (rate < AAMP_RATE_PAUSE)

priv_aamp.h
```diff
@@ -4179,6 +4179,11 @@ class PrivateInstanceAAMP : public DrmCallbacks, public std::enable_shared_from_
 	bool mLocalAAMPTsbFromConfig;						/**< AAMP TSB enabled in the configuration, regardless of the current channel */
 
 private:
+	/**
+	 * @brief Handle beginning of stream reached during rewind
+	 */
+	void HandleBeginningOfStreamReached();
```
dnfaulkner: Rename, as agreed

test/utests/tests/PrivAampTests/PrivAampTests.cpp
```diff
@@ -966,6 +966,50 @@ TEST_F(PrivAampTests,ReportProgressTest6)
 	p_aamp->ReportProgress(sync,beginningOfStream);
 }
 
+/**
+ * @brief Test ReportProgress when rewinding reaches the beginning of the TSB.
+ * Verifies that HandleBeginningOfStreamReached() is called and rate is reset to normal play.
```
dnfaulkner: Rename, as agreed

test/utests/tests/PrivAampTests/PrivAampTests.cpp
```diff
@@ -966,6 +966,50 @@ TEST_F(PrivAampTests,ReportProgressTest6)
...
+	// Setup mocks for the flow
+	EXPECT_CALL(*g_mockAampConfig, IsConfigSet(_)).WillRepeatedly(Return(false));
+	EXPECT_CALL(*g_mockAampConfig, IsConfigSet(eAAMPConfig_EnablePTSReStamp)).WillRepeatedly(Return(true));
+	EXPECT_CALL(*g_mockAampStreamSinkManager, GetStreamSink(_)).WillRepeatedly(Return(g_mockAampGstPlayer));
+
+	// Expect NotifySpeedChanged to be called with AAMP_NORMAL_PLAY_RATE
+	EXPECT_CALL(*g_mockAampEventManager, SendEvent(SpeedChanged(AAMP_NORMAL_PLAY_RATE), _)).Times(1);
+
+	// Call ReportProgress - position will be less than start (culledSeconds * 1000)
+	// This triggers HandleBeginningOfStreamReached()
```
dnfaulkner: Rename, as agreed

priv_aamp.h
jfagunde: Done

test/utests/tests/PrivAampTests/PrivAampTests.cpp
jfagunde: Done

test/utests/tests/PrivAampTests/PrivAampTests.cpp
jfagunde: Done

priv_aamp.cpp

priv_aamp.cpp

dnfaulkner (approved):

priv_aamp.cpp
```diff
@@ -2119,10 +2119,18 @@ void PrivateInstanceAAMP::ReportProgress(bool sync, bool beginningOfStream)
...
 			position = end;
 		}
-		else if (position < start)
-		{ // clamp start
-			AAMPLOG_WARN( "clamp position %fms < start %fms", position, start );
+		// If beginningOfStream is true or position < start, it means rewind has reached BoS
+		// Note: position could be = start immediately after tuning
+		else if (position < start || beginningOfStream)
+		{ // clamp start or handle BOS during rewind
+			AAMPLOG_MIL("Reached start of TSB, position %fms < start %fms, beginningOfStream %d, rate %f",
```
cpc005: verbose log line

priv_aamp.cpp
```diff
@@ -2119,10 +2119,18 @@ void PrivateInstanceAAMP::ReportProgress(bool sync, bool beginningOfStream)
...
+				position, start, beginningOfStream, rate);
 			position = start;
+			// Check the rate so that PlayFromTsbStart() is not called repeatedly
+			if (rate < AAMP_RATE_PAUSE)
+			{
+				PlayFromTsbStart();
```
cpc005, line 2130: Now ReportProgress() is doing more than the function name suggests??
Can it return a flag and the tune etc get done outside

priv_aamp.cpp
jfagunde: Yes, I agree. I will change it to an INFO.
This used to be a WARN, that is the reason I made it a MIL; but when playing from the start of the TSB this line is printed frequently.

priv_aamp.cpp
jfagunde: It is not easy to do that, since ReportProgress() is called from 4 different places.
How about renaming the method to MonitorProgress() as suggested?

priv_aamp.cpp
jfagunde: Replaced it with TRACE instead because it is really noisy when playing from the start of the TSB.

pstroffolino (approved):

PR #594, opened 2025-10-17
VPLAY-11384 Get access (session) token from aut service

jameslofthouse-RED:
Reason for change: Updating deprecated method of getting access token
Test Procedure: See ticket
Risks: Low

### DIFF
--- a/drm/AampDRMLicManager.cpp
+++ b/drm/AampDRMLicManager.cpp
@@ -283,7 +283,7 @@ KeyState AampDRMLicenseManager::acquireLicense( int& responseCode, std::shared_p
 				const char *sessionToken = NULL;
 				if(!usingAppDefinedAuthToken)
 				{ /* authToken not set externally by app */
-					sessionToken = getAccessToken(tokenLen, tokenError , aampInstance->mConfig->IsConfigSet(eAAMPConfig_SslVerifyPeer));
+					sessionToken = getAccessToken(tokenLen, tokenError);
 					AAMPLOG_WARN("Access Token from AuthServer");
 				}
 				else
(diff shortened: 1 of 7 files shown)


drm/AampDRMLicManager.cpp
```diff
@@ -621,86 +621,49 @@ string extractSubstring(string parentStr, string startStr, string endStr)
 	return ret;
 }
 
-
 /**
  *  @brief Get the accessToken from authService.
  */
-const char * AampDRMLicenseManager::getAccessToken(int &tokenLen, int &error_code , bool bSslPeerVerify)
+const char * AampDRMLicenseManager::getAccessToken(int &tokenLen, int &error_code)
```
Vinish100, line 628: Cant we just keep the token as a std::string? Any advantages of maintaining as a pointer and length?

middleware/externals/contentsecuritymanager/ContentSecurityManager.cpp
```diff
@@ -88,6 +88,14 @@ void ContentSecurityManager::DestroyInstance()
 	}
 }
 
+/**
+ * @brief To acquire an access token from auth service
+ */
+bool ContentSecurityManager::getSessionToken(std::string &token)
+{
+	return false;
```
Vinish100, line 96: Should we add a warning log to be notified if we call this function accidentally?

jameslofthouse-RED: Can do but it should not really happen

drm/AampDRMLicManager.cpp
jameslofthouse-RED: I was just trying to avoid changing too much code but I can look into it - everywhere it is used currently uses it as char *. A string would be nicer

PR #597, opened 2025-10-17
VPLAY-11251 AAMP-JS ( jsbinding ) prints are not genereated

Abhi-jith-S:
Reason for change : Enabling USE_ETHAN_LOG macro before building aampjsbinding

Test Procedure: See ticket

Risks: Low


CMakeLists.txt
```diff
@@ -213,7 +213,18 @@ install(TARGETS tsb
 		DESTINATION lib
 		PUBLIC_HEADER DESTINATION include
 )
+#Enabling  ethan logging macro
+if(CMAKE_USE_ETHAN_LOG)
```
Vinish100, line 225: Please also move the CMAKE_SYSTEMD_JOURNAL block to here so that the LIBAAMPJSBINDINGS_DEFINES macro is updated properly

PR #606, opened 2025-10-21
VPLAY-11126 [AAMP TSB] TsbReader waits for fragments in a tight loop

hthakkar-synamedia:
Reason for Change: When TsbReader tries to read the fragments that are not downloaded yet by the FetcherLoop, it enters into a quick tight loop. Fix here implements producer consumer design and wait for a conditional variable until new fragment is available. 
Added L1 microtest
Test Procedure: - Refer Ticket
Risks:low


test/utests/fakes/FakeTSBSessionManager.cpp
```diff
@@ -246,3 +246,13 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
 {
 	return nullptr;
 }
+
+void AampTSBSessionManager::WaitForNewTsbFragment()
+{
+
```
DomSyna: The style elsewhere in this file, is not to have a blank line for the empty fns

test/utests/tests/FragmentCollectorMpdTests/FragmentCollectorMpdTestCases.cpp
```diff
@@ -127,7 +127,10 @@ TEST_F(StreamAbstractionAAMP_MPD_Test, AdvanceTsbFetchTest)
 	EXPECT_CALL(*g_mockTSBSessionManager, PushNextTsbFragment(mediaStreamContext, _)).WillOnce(Return(true));
 
 	// Call the protected method through testable wrapper
-	mMpdStream->TestAdvanceTsbFetch(trackIdx, trickPlay, delta, waitForFreeFrag, bCacheFullState);
+	bool result = mMpdStream->TestAdvanceTsbFetch(trackIdx, trickPlay, delta, waitForFreeFrag, bCacheFullState);
+
+	// Verify that a fragment was found
```
DomSyna: When adding a comment, please consider if the comment actually adds and value, or is just stating the obvious and adding clutter.

For example in this test its pretty obvious from the test description what's going on so none of the comments serve any real purpose.

AampTSBSessionManager.cpp
```diff
@@ -374,6 +375,18 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
 	return removedFragment;
 }
 
+void AampTSBSessionManager::raiseNewVideoTsbContentNotification()
+{
+	mNewVideoTsbContentCV.notify_one();
```
DomSyna, line 384: Not sure if having the notify wrapped in a method really serves any purpose ? probably a taste thing

AampTSBSessionManager.cpp
```diff
@@ -450,7 +463,15 @@ void AampTSBSessionManager::ProcessWriteQueue()
 							}
 						}
 					}
+
 					UnlockReadMutex();
+					// Notify only for video fragments
```
DomSyna: Does this comment add any purpose ? isn't it obvious from the 
if (mediatype == eMEDIATYPE_VIDEO)

If you think a comment is needed, then probably needs more words to say why only notify for video, not just state what is obvious from looking at the trivial condition

AampTSBSessionManager.h
```diff
@@ -88,6 +88,15 @@ class AampTSBSessionManager
 	 * @brief Monitors the write queue and writes any pending data to AAMP TSB
 	 */
 	void ProcessWriteQueue();
+	/**
+	 * @brief Wait for new TSB fragment to be available
+	 */
+	void WaitForNewTsbFragment();
+
```
DomSyna: THe other methods don't seem to have a blank line between the declaration, and the next doxygen header

AampTSBSessionManager.h
```diff
@@ -88,6 +88,15 @@ class AampTSBSessionManager
...
+	/**
+	 * @brief Raises the new video TSB content notification
+	 */
+	void raiseNewVideoTsbContentNotification();
```
DomSyna: RaiseN... ? for consistency

AampTSBSessionManager.h
```diff
@@ -407,13 +416,15 @@ class AampTSBSessionManager
 	std::mutex mWriteQueueMutex;			// Mutex to synchronize access to the write queue.
 	std::mutex mReadMutex;					// Mutex to synchronize access to the data manager from reader and writer.
 	std::condition_variable mWriteThreadCV; // Condition variable to signal when data is available in the write queue
+	std::condition_variable mNewVideoTsbContentCV;	// Condition variable for signaling new TSB content
```
DomSyna: How about a more useful comment ?
maybe something like:
Conditional variable used to signal when content has been written to the TSB

AampTSBSessionManager.h
```diff
@@ -407,13 +416,15 @@ class AampTSBSessionManager
...
 	std::queue<TSBWriteData> mWriteQueue;	// Queue to store write data.
 	double mLastVideoPos;
 	double mStoreEndPosition; 		/**< Last reported TSB Store end position*/
 	double mLiveEndPosition;		/**< Last reported Live end position*/
 	AampTime  mCurrentWritePosition; /**< The last fragment position written to the TSB */
 	std::shared_ptr<AampTsbMetaData> mLastAdReservationMetaDataProcessed; /**< Last ad reservation metadata processed */
 	std::shared_ptr<AampTsbMetaData> mLastAdPlacementMetaDataProcessed; /**< Last ad placement metadata processed */
+	bool mHasNewVideoTsbContent;			// Flag to indicate new TSB content is available
```
DomSyna: Maybe something like:
Flag used in conjunction with condvar mNewVideoTsbContentCV to signal a segment has been added to the TSB

fragmentcollector_mpd.cpp
```diff
@@ -10379,9 +10383,16 @@ void StreamAbstractionAAMP_MPD::TsbReader()
 					}
 				}
 				else
-				{	// This sleep will hit when there is no content to download and cache is not full
-					// and refresh interval timeout not reached . To Avoid tight loop adding a min delay
-					aamp->interruptibleMsSleep(50);
+				{
+					// Check if we found a segment, otherwise wait for new content
```
DomSyna: Not sure this comment is correct or useful ?

fragmentcollector_mpd.cpp
```diff
@@ -10379,9 +10383,16 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
+					if (segmentFound)
+					{
+						aamp->interruptibleMsSleep(50);				//To Avoid tight loop adding a small delay
+					}
+					else if (aamp->rate > AAMP_NORMAL_PLAY_RATE)
+					{
+						tsbSessionManager->WaitForNewTsbFragment();
```
DomSyna: Maybe remove the comment at 10387, and add a useful one here stating why we are waiting

test/utests/fakes/FakeTSBSessionManager.cpp
hthakkar-synamedia: Done that

test/utests/tests/FragmentCollectorMpdTests/FragmentCollectorMpdTestCases.cpp
hthakkar-synamedia: Removed them

AampTSBSessionManager.cpp
hthakkar-synamedia: I wrote this public method so that I could raise a notification in the priv_aamp in the DisableDownloads method

AampTSBSessionManager.cpp
hthakkar-synamedia: Sure, I have removed the comment.

AampTSBSessionManager.h
hthakkar-synamedia: Removed that

AampTSBSessionManager.h
hthakkar-synamedia: Used your comment. Thanks

fragmentcollector_mpd.cpp
hthakkar-synamedia: Removed this comment

fragmentcollector_mpd.cpp
hthakkar-synamedia: Added a comment explaining why are we waiting

AampTSBSessionManager.h
hthakkar-synamedia: Thanks, changed that

AampTSBSessionManager.h
hthakkar-synamedia: Yep, changed that

hthakkar-synamedia:
Addressed the comments

hthakkar-synamedia (commented):
Addressed the comments

AampTSBSessionManager.cpp
```diff
@@ -450,7 +463,15 @@ void AampTSBSessionManager::ProcessWriteQueue()
 							}
 						}
 					}
+
 					UnlockReadMutex();
+
+					if (mediatype == eMEDIATYPE_VIDEO)
+					{
+						mHasNewVideoTsbContent = true;
+						RaiseNewVideoTsbContentNotification();
```
DomSyna: Given the wait method clears the flag, would it be neater if the Raise fn took a bool and if true, set the flag ?

fragmentcollector_mpd.cpp
```diff
@@ -10379,9 +10383,16 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
-					// and refresh interval timeout not reached . To Avoid tight loop adding a min delay
-					aamp->interruptibleMsSleep(50);
+				{
+					if (segmentFound)
+					{
+						aamp->interruptibleMsSleep(50);				//To Avoid tight loop adding a small delay
+					}
+					else if (aamp->rate > AAMP_NORMAL_PLAY_RATE)
+					{
+						// All the segments in TSB have been send to gstreamer, wait for new fragments to be available in TSB
```
DomSyna: typo... sent

hthakkar-synamedia: done

AampTSBSessionManager.cpp
hthakkar-synamedia: Done

hthakkar-synamedia (commented):
fixed typo and modified the RaiseNewVideoTsbContentNotification function

fragmentcollector_mpd.cpp
```diff
@@ -10379,9 +10383,16 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
 				else
-				{	// This sleep will hit when there is no content to download and cache is not full
-					// and refresh interval timeout not reached . To Avoid tight loop adding a min delay
-					aamp->interruptibleMsSleep(50);
+				{
+					if (segmentFound)
+					{
+						aamp->interruptibleMsSleep(50);				//To Avoid tight loop adding a small delay
+					}
+					else if (aamp->rate > AAMP_NORMAL_PLAY_RATE)
```
jfagunde, line 10014: I would add an else with a WARN. I think that should never happen.

fragmentcollector_mpd.cpp
```diff
@@ -10364,7 +10368,7 @@ void StreamAbstractionAAMP_MPD::TsbReader()
 						break;
 					}
 					AAMPLOG_TRACE("EOS from both tracks - Wait for next fragment");
```
jfagunde, line 9989: Maybe make this an INFO?

AampTSBSessionManager.cpp
```diff
@@ -374,6 +375,22 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
 	return removedFragment;
 }
 
+void AampTSBSessionManager::RaiseNewVideoTsbContentNotification(bool setFlag)
+{
+	if (setFlag)
+	{
+		mHasNewVideoTsbContent = setFlag;
```
jfagunde: mReadMutex should be locked before accessing this flag.

AampTSBSessionManager.cpp
```diff
@@ -374,6 +375,22 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
...
+	}
+	mNewVideoTsbContentCV.notify_one();
+}
+
+void AampTSBSessionManager::WaitForNewTsbFragment()
+{
+	std::unique_lock<std::mutex> lock(mReadMutex);
+	mNewVideoTsbContentCV.wait(lock, [this]() { return mHasNewVideoTsbContent || !mAamp->mDownloadsEnabled; });
+	mHasNewVideoTsbContent = false;
```
jfagunde: How about clearing the flag before the wait.

AampTSBSessionManager.h
```diff
@@ -88,6 +88,16 @@ class AampTSBSessionManager
 	 * @brief Monitors the write queue and writes any pending data to AAMP TSB
 	 */
 	void ProcessWriteQueue();
+	/**
+	 * @brief Wait for new TSB fragment to be available
```
jfagunde: Should this be "...new TSB video fragment..."?

AampTSBSessionManager.h
```diff
@@ -88,6 +88,16 @@ class AampTSBSessionManager
...
+	 */
+	void WaitForNewTsbFragment();
```
jfagunde: Should the name of the function be WaitForNewVideoTsbFragment(), to be consistent?

priv_aamp.cpp
```diff
@@ -7244,6 +7244,11 @@ void PrivateInstanceAAMP::DisableDownloads(void)
 		mDownloadsEnabled = false;
 		mDownloadsDisabled.notify_all();
 	}
+
+	if (mTSBSessionManager)
+	{
+		mTSBSessionManager->RaiseNewVideoTsbContentNotification(false);
```
jfagunde: I would add a comment explaining that this is called to unblock any thread waiting on this condition.

fragmentcollector_mpd.cpp
hthakkar-synamedia: Done

fragmentcollector_mpd.cpp
hthakkar-synamedia: Done

AampTSBSessionManager.cpp
hthakkar-synamedia: Done

AampTSBSessionManager.h
hthakkar-synamedia: Done

priv_aamp.cpp
hthakkar-synamedia: Done

AampTSBSessionManager.cpp
hthakkar-synamedia: Done

AampTSBSessionManager.h
hthakkar-synamedia: Done

fragmentcollector_mpd.cpp
```diff
@@ -10323,14 +10325,16 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
 			while (!exitLoop)
 			{
+				bool segmentFound {false};
 				for (int trackIdx = (mNumberOfTracks - 1); trackIdx >= 0; trackIdx--)
 				{
 					cacheFullStatus[trackIdx] = true;
 					if (!tsbSessionManager->GetTsbReader((AampMediaType) trackIdx)->IsEos())
 					{
-						AdvanceTsbFetch(trackIdx, trickPlay, delta, waitForFreeFrag, cacheFullStatus[trackIdx]);
+						segmentFound = AdvanceTsbFetch(trackIdx, trickPlay, delta, waitForFreeFrag, cacheFullStatus[trackIdx]);
```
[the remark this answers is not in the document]
hthakkar-synamedia: last value was required

fragmentcollector_mpd.cpp
```diff
@@ -10379,9 +10383,20 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
+						aamp->interruptibleMsSleep(50);				//To Avoid tight loop adding a small delay
+					}
+					else if (aamp->rate > AAMP_NORMAL_PLAY_RATE)
+					{
+						// All the segments in TSB have been sent to gstreamer, wait for new fragments to be available in TSB
+						tsbSessionManager->WaitForNewVideoTsbFragment();
+					}
+					else
+					{
+						AAMPLOG_WARN("Should not wait for AAMP_NORMAL_PLAY_RATE <= 1");
```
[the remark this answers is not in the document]
hthakkar-synamedia: done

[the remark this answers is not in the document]
hthakkar-synamedia: WaitforNewVideoTsbFragment is only applicable to WaitForNewVideoTsbFragment>AAMP_NORMAL_PLAY_RATE

fragmentcollector_mpd.cpp
```diff
@@ -10318,14 +10320,16 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
 			while (!exitLoop)
 			{
+				bool segmentFound {false};
 				for (int trackIdx = (mNumberOfTracks - 1); trackIdx >= 0; trackIdx--)
 				{
 					cacheFullStatus[trackIdx] = true;
 					if (!tsbSessionManager->GetTsbReader((AampMediaType) trackIdx)->IsEos())
 					{
-						AdvanceTsbFetch(trackIdx, trickPlay, delta, waitForFreeFrag, cacheFullStatus[trackIdx]);
+						segmentFound = AdvanceTsbFetch(trackIdx, trickPlay, delta, waitForFreeFrag, cacheFullStatus[trackIdx]);
```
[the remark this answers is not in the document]
jfagunde: I think Copilot is right. AdvanceTsbFetch() is called in a loop for each track. With the current code, the variable segmentFound will hold the value returned by AdvanceTsbFetch() for the last track in the loop (i.e. subtitles).
Using the OR assignment operator, as suggested, would mean that if a segment was read for any track, AAMP would not wait for new segments to be downloaded. It sounds like a sensible idea. Alternatively, you could check only the video track.

fragmentcollector_mpd.cpp
```diff
@@ -10374,9 +10378,20 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
+						aamp->interruptibleMsSleep(50);				//To Avoid tight loop adding a small delay
+					}
+					else if (aamp->rate > AAMP_NORMAL_PLAY_RATE)
+					{
+						// All the segments in TSB have been sent to gstreamer, wait for new fragments to be available in TSB
+						tsbSessionManager->WaitForNewVideoTsbFragment();
+					}
+					else
+					{
+						AAMPLOG_WARN("Unexpected condition: Should not wait for rate <= AAMP_NORMAL_PLAY_RATE");
```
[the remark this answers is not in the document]
jfagunde: The text "No segment found for rate <= AAMP_NORMAL_PLAY_RATE" is probably clearer describing what is happening. But AAMP should not wait if the rate is PLAY or rewind.

AampTSBSessionManager.cpp
```diff
@@ -374,6 +375,23 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
...
 }
 
+void AampTSBSessionManager::RaiseNewVideoTsbContentNotification(bool setFlag)
+{
+	if (setFlag)
+	{
+		std::unique_lock<std::mutex> lock(mReadMutex);
+		mHasNewVideoTsbContent = setFlag;
+	}
+	mNewVideoTsbContentCV.notify_one();
```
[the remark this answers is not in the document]
jfagunde: It is intentional to still notify when setFlag is false. This is used when downloads are disabled to unblock any thread waiting for a new segment. The predicate used for the wait is mHasNewVideoTsbContent or !mDownloadsEnabled
Maybe adding a comment would help understand the code? Something like...
"If setFlag is false, still call notify to unblock any thread waiting for a segment to be downloaded. This is used when downloads are disabled. The predicate used for the wait is an OR of the new video TSB content and downloads disabled flags."

fragmentcollector_mpd.cpp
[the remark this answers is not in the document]
hthakkar-synamedia: I have used OR assignment operator

AampTSBSessionManager.cpp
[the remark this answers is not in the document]
hthakkar-synamedia: Added the suggested comment

fragmentcollector_mpd.cpp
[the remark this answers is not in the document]
hthakkar-synamedia: Modified the comment

jfagunde (approved):

AampTSBSessionManager.cpp
```diff
@@ -374,6 +375,26 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
...
 }
 
+void AampTSBSessionManager::RaiseNewVideoTsbContentNotification(bool setFlag)
+{
+	// If setFlag is false, still call notify to unblock any thread waiting for a segment to be downloaded.
+	// This is used when downloads are disabled. The predicate used for the wait is an OR of the new video
+	// TSB content and downloads disabled flags.
+	if (setFlag)
+	{
+		std::unique_lock<std::mutex> lock(mReadMutex);
```
DomSyna: I don't think adding this lock adds any value. The wait that reads the value will check it before suspending the thread, and each time its woken up by the notify. It should then clear it. 
If this happens first we don't suspend, if it happens second we will wake up almost immediately.

AampTSBSessionManager.cpp
```diff
@@ -374,6 +375,26 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
...
+		mHasNewVideoTsbContent = setFlag;
+	}
+	mNewVideoTsbContentCV.notify_one();
+}
+
+void AampTSBSessionManager::WaitForNewVideoTsbFragment()
+{
+	std::unique_lock<std::mutex> lock(mReadMutex);
+	mHasNewVideoTsbContent = false;
```
DomSyna: We should clear the var after waking up. Doing it here adds the race condition where we could be clearing it just before waiting. If it was just set as part of reading the last new segment signalled in the current manifest we may have to wait ~6 seconds before we read any more segments as that's the manifest poll interval.

fragmentcollector_mpd.cpp
```diff
@@ -10374,9 +10378,20 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
+						aamp->interruptibleMsSleep(50);				//To Avoid tight loop adding a small delay
+					}
+					else if (aamp->rate > AAMP_NORMAL_PLAY_RATE)
+					{
+						// All the segments in TSB have been sent to gstreamer, wait for new fragments to be available in TSB
+						tsbSessionManager->WaitForNewVideoTsbFragment();
+					}
+					else
+					{
+						AAMPLOG_WARN("No segment found for rate <= AAMP_NORMAL_PLAY_RATE");
```
DomSyna: Don't think this is possible, but doesn't hurt

AampTSBSessionManager.cpp
hthakkar-synamedia: I have accepted your advice and removed the lock

AampTSBSessionManager.cpp
hthakkar-synamedia: Yes, your explanation is right. Some of the L2 tests failed for the reasons that you have mentioned.

AampTSBSessionManager.cpp
jfagunde: I don't see the mutex locked before setting mHasNewVideoTsbContent. The code should be something like this:
```
if (setFlag)
{
	std::unique_lock<std::mutex> lock(mReadMutex);
	mHasNewVideoTsbContent = setFlag;
}
```

AampTSBSessionManager.cpp
```diff
@@ -375,6 +376,25 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
 	return removedFragment;
 }
 
+void AampTSBSessionManager::RaiseNewVideoTsbContentNotification(bool setFlag)
```
jfagunde: It is not very clear the usage of the flag parameter. How about moving the logic to set it inside this function? I.e.
```
void AampTSBSessionManager::RaiseNewVideoTsbContentNotification()
{
	if (mAamp->mDownloadsEnabled)
	{
		std::unique_lock<std::mutex> lock(mReadMutex);
		mHasNewVideoTsbContent = true;
	}
	mNewVideoTsbContentCV.notify_one();
}
```

AampTSBSessionManager.cpp
hthakkar-synamedia: With explanation from @jfagunde and query to co-pilot, it is advisable to keep the lock. There is a race condition without using lock.  From cppreference the wait is equivalent to 
while (!pred())
    wait(lock);. 
In edge case, the predicate is not set and before the thread enters wait, the other thread can wake up and set the predicate and raise notification. Now the wait is entered and notification is lost.

AampTSBSessionManager.cpp
hthakkar-synamedia: Done

jfagunde (approved):

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -581,3 +581,62 @@ TEST_F(AampTsbSessionManagerTests, PushNextTsbFragment_SkipFragment_BoS)
...
+
+	//waitForNewVideoTsbFragment thread is running, now exercise the test criteria
+	mAamp->mDownloadsEnabled = false;
+
+	// Raise the notification from the main thread
+	mAampTSBSessionManager->RaiseNewVideoTsbContentNotification();
+
+	// Wait for the consumer thread to complete
+	consumerThread.join();
+}
```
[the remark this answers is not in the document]
hthakkar-synamedia: This is not an edge case. The downloads need to be enabled for the WaitForNewVideoTsbFragment thread to start executing. This test verifies the behaviour when downloads are disabled and then notification is raised.

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -581,3 +581,62 @@ TEST_F(AampTsbSessionManagerTests, PushNextTsbFragment_SkipFragment_BoS)
...
+{
+	mAamp->mDownloadsEnabled = false;
+
+	// Spawn a thread that waits for and consumes the notification
+	std::thread consumerThread([this]() {
+		mAampTSBSessionManager->WaitForNewVideoTsbFragment();
+	});
+	// Wait for the consumer thread to complete
+	consumerThread.join();
+}
```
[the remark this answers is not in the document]
hthakkar-synamedia: The test is sufficient to evaluate the behviour

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -581,3 +581,62 @@ TEST_F(AampTsbSessionManagerTests, PushNextTsbFragment_SkipFragment_BoS)
...
+{
+	mAamp->mDownloadsEnabled = true;
+
+	// Spawn a thread that waits for and consumes the notification
+	std::thread consumerThread([this]() {
+		mAampTSBSessionManager->WaitForNewVideoTsbFragment();
+	});
+
+	// Give the consumer thread time to start waiting on the condition variable
+	std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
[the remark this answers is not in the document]
hthakkar-synamedia: Introduced a larger timeout

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -581,3 +581,62 @@ TEST_F(AampTsbSessionManagerTests, PushNextTsbFragment_SkipFragment_BoS)
...
+	// Necessary to ensure that waitForNewVideoTsbFragment does not return immediately
+	mAamp->mDownloadsEnabled = true;
+
+	// Spawn a thread that waits for and consumes the notification
+	std::thread consumerThread([this]() {
+		mAampTSBSessionManager->WaitForNewVideoTsbFragment();
+	});
+
+	// Give the consumer thread time to start waiting on the condition variable
+	std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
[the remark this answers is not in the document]
hthakkar-synamedia: Introduced a larger timeout

priv_aamp.cpp
```diff
@@ -7772,6 +7772,11 @@ void PrivateInstanceAAMP::DisableDownloads(void)
 		mDownloadsEnabled = false;
 		mDownloadsDisabled.notify_all();
 	}
+	// Unblock any thread waiting on this condition.
+	if (mTSBSessionManager)
+	{
+		mTSBSessionManager->RaiseNewVideoTsbContentNotification();
+	}
```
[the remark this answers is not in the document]
hthakkar-synamedia: Modified the comment

AampTSBSessionManager.h
```diff
@@ -88,6 +88,14 @@ class AampTSBSessionManager
 	 * @brief Monitors the write queue and writes any pending data to AAMP TSB
 	 */
 	void ProcessWriteQueue();
+	/**
+	 * @brief Wait for new TSB video fragment to be available
+	 */
+	void WaitForNewVideoTsbFragment();
+	/**
+	 * @brief Raises the new video TSB content notification
```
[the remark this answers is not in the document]
hthakkar-synamedia: Igoring the comment - it is a too versbose explanation. The amount of description added is similar to description added in all of software.

fragmentcollector_mpd.cpp
```diff
@@ -9944,9 +9948,20 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
```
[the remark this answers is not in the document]
hthakkar-synamedia: The WaitForNewVideoTsbFragment is intended for rates > 1.0 it will not happen on other rates. The else branch is there to warn that the segment found for rates < Normal playrate; giving information to developer that there is fault in logic and that should be corrected.

AampTSBSessionManager.h
```diff
@@ -407,13 +415,15 @@ class AampTSBSessionManager
...
 	std::condition_variable mWriteThreadCV; // Condition variable to signal when data is available in the write queue
+	std::condition_variable mNewVideoTsbContentCV;	// Conditional variable used to signal when content has been written to the TSB
 	std::queue<TSBWriteData> mWriteQueue;	// Queue to store write data.
 	double mLastVideoPos;
 	double mStoreEndPosition; 		/**< Last reported TSB Store end position*/
 	double mLiveEndPosition;		/**< Last reported Live end position*/
 	AampTime  mCurrentWritePosition; /**< The last fragment position written to the TSB */
 	std::shared_ptr<AampTsbMetaData> mLastAdReservationMetaDataProcessed; /**< Last ad reservation metadata processed */
 	std::shared_ptr<AampTsbMetaData> mLastAdPlacementMetaDataProcessed; /**< Last ad placement metadata processed */
+	bool mHasNewVideoTsbContent;			// Flag used in conjunction with conditional variable mNewVideoTsbContentCV to signal a segment has been added to the TSB
```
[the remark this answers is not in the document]
hthakkar-synamedia: mHasNewVideoTsbContent  is protected by mutex (mReadMutex). It is seen in it implementation.

fragmentcollector_mpd.cpp
```diff
@@ -9888,14 +9890,16 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
 			while (!exitLoop)
 			{
+				bool segmentFound {false};
 				for (int trackIdx = (mNumberOfTracks - 1); trackIdx >= 0; trackIdx--)
 				{
 					cacheFullStatus[trackIdx] = true;
 					if (!tsbSessionManager->GetTsbReader((AampMediaType) trackIdx)->IsEos())
 					{
-						AdvanceTsbFetch(trackIdx, trickPlay, delta, waitForFreeFrag, cacheFullStatus[trackIdx]);
+						segmentFound |= AdvanceTsbFetch(trackIdx, trickPlay, delta, waitForFreeFrag, cacheFullStatus[trackIdx]);
```
cpc005: Bitwise OR ??
but if you do 
segmentFound = segmentFound  ||  AdvanceTsbFetch() 
you may not call AdvanceTsbFetch()
what is the intent here

AampTSBSessionManager.cpp
```diff
@@ -375,6 +376,23 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
 	return removedFragment;
 }
 
+void AampTSBSessionManager::RaiseNewVideoTsbContentNotification()
+{	// The predicate used for the wait is an OR of the new video TSB content and downloads disabled flags.
+	if (mAamp->mDownloadsEnabled)
```
cpc005: Cannot understand why we have the test 
if (mAamp->if (mAamp->mDownloadsEnabled))

as we are making a more complex relationship between mAamp->mDownloadsEnabled and mHasNewVideoTsbContent 

Any reason why we cannot just have
std::unique_lock<std::mutex> lock(mReadMutex);
mHasNewVideoTsbContent = true;
mNewVideoTsbContentCV.notify_one();

fragmentcollector_mpd.cpp
hthakkar-synamedia: I have replaced bitwise OR with logical OR in TsbReader. In C++ both these 
1) Replaced logic bool trackSegmentFound = AdvanceTsbFetch();
segmentFound = segmentFound || trackSegmentFound;
and
2) Original segmentFound |= AdvanceTsbFetch();
have same effect. However option 1 is better approach since it is easier to understand that AdvanceTsbFetch() will be called for every track.

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -581,3 +581,22 @@ TEST_F(AampTsbSessionManagerTests, PushNextTsbFragment_SkipFragment_BoS)
 
 	EXPECT_TRUE(mAampTSBSessionManager->PushNextTsbFragment(mMediaStreamContext.get(), numFreeFragments));
 }
+
+// Test RaiseNewVideoTsbContentNotification functionality
+TEST_F(AampTsbSessionManagerTests, RaiseNewVideoTsbContentNotification)
```
jfagunde: You could add another test that calls RaiseNewVideoTsbContentNotification() first and then spawns the thread that calls WaitForNewVideoTsbFragment(). I.e.
```
	mAampTSBSessionManager->RaiseNewVideoTsbContentNotification();

	std::thread consumerThread([this]() {
		mAampTSBSessionManager->WaitForNewVideoTsbFragment();
	});

	consumerThread.join();
```

AampTSBSessionManager.cpp
hthakkar-synamedia: I have followed the suggestion. I have renamed mHasNewVideoTsbContent  with mStopWaitingForVideoTsb to clarify the implementation

jfagunde (approved):

AampTSBSessionManager.cpp
```diff
@@ -375,6 +376,20 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
...
+void AampTSBSessionManager::RaiseNewVideoTsbContentNotification()
+{
+	std::unique_lock<std::mutex> lock(mReadMutex);
+	mStopWaitingForVideoTsb = true;
+	mNewVideoTsbContentCV.notify_one();
+}
+
+void AampTSBSessionManager::WaitForNewVideoTsbFragment()
+{
+	std::unique_lock<std::mutex> lock(mReadMutex);
```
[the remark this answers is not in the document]
hthakkar-synamedia: mStopWaitingForVideoTsb set to false in the Init method

AampTSBSessionManager.cpp
```diff
@@ -452,7 +467,14 @@ void AampTSBSessionManager::ProcessWriteQueue()
 							}
 						}
 					}
+
 					UnlockReadMutex();
+
+					if (mediatype == eMEDIATYPE_VIDEO)
+					{
+						RaiseNewVideoTsbContentNotification();
+					}
```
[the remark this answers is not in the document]
hthakkar-synamedia: We wait for video fragments since there is no point in progressing if video fragments have not been added to Tsb

fragmentcollector_mpd.cpp
```diff
@@ -9944,9 +9949,21 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
+					}
+					// AAMP could reach the end of the TSB only when doing FF (rate > AAMP_NORMAL_PLAY_RATE)
+					else if (aamp->rate > AAMP_NORMAL_PLAY_RATE)
+					{
+						// All the segments in TSB have been sent to gstreamer, wait for new fragments to be available in TSB
+						tsbSessionManager->WaitForNewVideoTsbFragment();
+					}
+					else
+					{
+						AAMPLOG_WARN("No segment found for rate <= AAMP_NORMAL_PLAY_RATE");
```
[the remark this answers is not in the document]
hthakkar-synamedia: This warning message should never be hit. If it does hit, it is upto the developer to investigate the scenario. The existing implementation is appropriate.

AampTSBSessionManager.cpp
```diff
@@ -375,6 +376,20 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
...
+{
+	std::unique_lock<std::mutex> lock(mReadMutex);
+	mStopWaitingForVideoTsb = true;
+	mNewVideoTsbContentCV.notify_one();
+}
+
+void AampTSBSessionManager::WaitForNewVideoTsbFragment()
+{
+	std::unique_lock<std::mutex> lock(mReadMutex);
+	mNewVideoTsbContentCV.wait(lock, [this]() { return mStopWaitingForVideoTsb; });
```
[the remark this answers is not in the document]
hthakkar-synamedia: When the downloads are disabled, we call RaiseNewVideoTsbContentNotification effectively setting the predicate mStopWaitingForVideoTsb.

jfagunde (approved):

AampTSBSessionManager.cpp
```diff
@@ -106,6 +107,7 @@ void AampTSBSessionManager::Init()
 			// Initialize TSB readers
 			InitializeTsbReaders();
 			mStopThread_.store(false);
+			mStopWaitingForVideoTsb = false;
```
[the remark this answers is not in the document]
jfagunde: I would follow suggestion number 3 and add a comment explaining that mReadMutex does not need to be locked because the threads that access this variable are not running yet.

AampTSBSessionManager.cpp
```diff
@@ -375,6 +377,20 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
...
+	mStopWaitingForVideoTsb = true;
+	mNewVideoTsbContentCV.notify_one();
+}
+
+void AampTSBSessionManager::WaitForVideoTsbContentOrAbort()
+{
+	std::unique_lock<std::mutex> lock(mReadMutex);
+	mNewVideoTsbContentCV.wait(lock, [this]() { return mStopWaitingForVideoTsb; });
+	mStopWaitingForVideoTsb = false;
+}
```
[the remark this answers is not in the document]
jfagunde: This comment is incorrect. The conditional variable wait releases the mutex while waiting (see https://en.cppreference.com/w/cpp/thread/condition_variable/wait.html). It is true that this mutex is locked in multiple places in the code, but always for a short period of time. So this is not an issue.

fragmentcollector_mpd.cpp
```diff
@@ -9928,7 +9933,7 @@ void StreamAbstractionAAMP_MPD::TsbReader()
 						exitLoop = true;
 						break;
 					}
-					AAMPLOG_TRACE("EOS from both tracks - Wait for next fragment");
+					AAMPLOG_INFO("EOS from both tracks - Wait for next fragment");
```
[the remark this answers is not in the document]
jfagunde: I don't think this is an issue, EOS is an event that does not happen very frequently.

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -581,3 +581,33 @@ TEST_F(AampTsbSessionManagerTests, PushNextTsbFragment_SkipFragment_BoS)
...
+{
+	mAampTSBSessionManager->NotifyVideoTsbWaiters();
+
+	// Spawn a thread that waits for the notification
+	std::thread consumerThread([this]() {
+		mAampTSBSessionManager->WaitForVideoTsbContentOrAbort();
+	});
+
+	consumerThread.join();
+}
```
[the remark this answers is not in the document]
jfagunde: The comment on lines 387-392 is wrong and the test is correct.

AampTSBSessionManager.cpp
```diff
@@ -375,6 +377,20 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
...
+	std::unique_lock<std::mutex> lock(mReadMutex);
+	mStopWaitingForVideoTsb = true;
+	mNewVideoTsbContentCV.notify_one();
+}
+
+void AampTSBSessionManager::WaitForVideoTsbContentOrAbort()
+{
+	std::unique_lock<std::mutex> lock(mReadMutex);
+	mNewVideoTsbContentCV.wait(lock, [this]() { return mStopWaitingForVideoTsb; });
+	mStopWaitingForVideoTsb = false;
```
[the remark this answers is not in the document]
jfagunde: This comment is wrong. In the sequence described, thread B sets the flag before notifying the waiting thread (inside NotifyVideoTsbWaiters()), so when thread A calls WaitForVideoTsbContentOrAbort(), the flag is set and it will not wait. It is not a good idea to clear the flag before waiting because if the video fragment has been fetched when WaitForVideoTsbContentOrAbort() is called, there is no need to wait.

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -581,3 +581,33 @@ TEST_F(AampTsbSessionManagerTests, PushNextTsbFragment_SkipFragment_BoS)
...
+		mAampTSBSessionManager->WaitForVideoTsbContentOrAbort();
+	});
+
+	// Brief sleep to allow thread to enter waiting state
+	// Note: This is a timing assumption, but std::thread provides no "is waiting" status check
+	std::this_thread::sleep_for(std::chrono::milliseconds(10));
+
+	mAampTSBSessionManager->NotifyVideoTsbWaiters();
+
+	consumerThread.join();
```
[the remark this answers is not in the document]
jfagunde: This comment is incorrect. In the unlikely event that NotifyVideoTsbWaiters() is called before the spawned thread is waiting, the test would pass. Adding a synchronization mechanism to avoid the sleep would make the test more complex.

AampTSBSessionManager.cpp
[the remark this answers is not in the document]
hthakkar-synamedia: Added comment why mutex is not needed

AampTSBSessionManager.cpp
```diff
@@ -452,7 +468,14 @@ void AampTSBSessionManager::ProcessWriteQueue()
 							}
 						}
 					}
+
 					UnlockReadMutex();
+
+					if (mediatype == eMEDIATYPE_VIDEO)
+					{
+						NotifyVideoTsbWaiters();
+					}
```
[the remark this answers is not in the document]
hthakkar-synamedia: Rasing notification for video fragments is sufficient

AampTSBSessionManager.h
```diff
@@ -88,6 +88,14 @@ class AampTSBSessionManager
 	 * @brief Monitors the write queue and writes any pending data to AAMP TSB
 	 */
 	void ProcessWriteQueue();
+	/**
+	 * @brief Wait for new TSB video fragment to be available or abort signal
+	 */
+	void WaitForVideoTsbContentOrAbort();
+	/**
+	 * @brief Notifies waiting threads when new video TSB content is available or downloads are disabled
+	 */
```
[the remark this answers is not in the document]
hthakkar-synamedia: Modified comment

test/utests/fakes/FakeTSBSessionManager.cpp
```diff
@@ -246,3 +246,11 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
 {
 	return nullptr;
 }
+
+void AampTSBSessionManager::WaitForVideoTsbContentOrAbort()
+{
+}
+
+void AampTSBSessionManager::NotifyVideoTsbWaiters()
+{
```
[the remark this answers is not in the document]
hthakkar-synamedia: The mock methods are added as needed.

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -581,3 +581,33 @@ TEST_F(AampTsbSessionManagerTests, PushNextTsbFragment_SkipFragment_BoS)
...
+TEST_F(AampTsbSessionManagerTests, NotifyVideoTsbWaiters)
+{
+	// Spawn a thread that waits for and consume the notification
+	std::thread consumerThread([this]() {
+		mAampTSBSessionManager->WaitForVideoTsbContentOrAbort();
+	});
+
+	// Brief sleep to allow thread to enter waiting state
+	// Note: This is a timing assumption, but std::thread provides no "is waiting" status check
+	std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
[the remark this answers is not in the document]
hthakkar-synamedia: header thread and chrono included in the test

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -581,3 +581,33 @@ TEST_F(AampTsbSessionManagerTests, PushNextTsbFragment_SkipFragment_BoS)
...
+
```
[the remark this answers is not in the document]
hthakkar-synamedia: 10 millisecon is satisfactory wait time and it keeps test simple.

test/utests/tests/AampTsbSessionManagerTests_Mocked/FunctionalTests.cpp
```diff
@@ -581,3 +581,33 @@ TEST_F(AampTsbSessionManagerTests, PushNextTsbFragment_SkipFragment_BoS)
...
+TEST_F(AampTsbSessionManagerTests, NotifyVideoTsbWaiters_BeforeWait)
+{
+	mAampTSBSessionManager->NotifyVideoTsbWaiters();
+
+	// Spawn a thread that waits for the notification
+	std::thread consumerThread([this]() {
+		mAampTSBSessionManager->WaitForVideoTsbContentOrAbort();
+	});
+
+	consumerThread.join();
```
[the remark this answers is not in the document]
hthakkar-synamedia: Google framework timeout will take care of it

AampTSBSessionManager.cpp
```diff
@@ -374,6 +377,20 @@ TsbFragmentDataPtr AampTSBSessionManager::RemoveFragmentDeleteInit(AampMediaType
 	return removedFragment;
 }
 
+void AampTSBSessionManager::NotifyVideoTsbWaiters()
+{
+	std::unique_lock<std::mutex> lock(mReadMutex);
+	mStopWaitingForVideoTsb = true;
+	mNewVideoTsbContentCV.notify_one();
```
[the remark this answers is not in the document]
hthakkar-synamedia: Only one waiter. Changed the NotifyVideoTsbWaiters to NotifyVideoTsbWaiter

AampTSBSessionManager.h
```diff
@@ -407,13 +415,15 @@ class AampTSBSessionManager
 	std::mutex mWriteQueueMutex;			// Mutex to synchronize access to the write queue.
 	std::mutex mReadMutex;					// Mutex to synchronize access to the data manager from reader and writer.
 	std::condition_variable mWriteThreadCV; // Condition variable to signal when data is available in the write queue
+	std::condition_variable mNewVideoTsbContentCV;	// Conditional variable used to signal when content has been written to the TSB
```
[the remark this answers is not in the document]
hthakkar-synamedia: Changed to condition variable

fragmentcollector_mpd.cpp
```diff
@@ -10000,9 +10005,21 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
+					// AAMP could reach the end of the TSB only when doing FF (rate > AAMP_NORMAL_PLAY_RATE)
+					else if (aamp->rate > AAMP_NORMAL_PLAY_RATE)
+					{
+						// All the segments in TSB have been sent to gstreamer, wait for new fragments to be available in TSB
+						tsbSessionManager->WaitForVideoTsbContentOrAbort();
+					}
+					else
+					{
+						AAMPLOG_WARN("No segment found for rate <= AAMP_NORMAL_PLAY_RATE");
+					}
```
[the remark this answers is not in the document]
pstroffolino: valid issue?  the use of WaitForVideoTsbContentOrAbort is limited to aamp->rate > AAMP_NORMAL_PLAY_RATE

fragmentcollector_mpd.cpp
```diff
@@ -10000,9 +10005,24 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
+					else if (aamp->rate > AAMP_NORMAL_PLAY_RATE)
+					{
+						// All the segments in TSB have been sent to gstreamer, wait for new fragments to be available in TSB
+						tsbSessionManager->WaitForVideoTsbContentOrAbort();
+					}
+					else
+					{
+						// Unexpected: no segment found at normal/slow rate outside of EOS. Sleep to
+						// avoid a tight loop and throttle log output until the condition resolves.
+						AAMPLOG_TRACE("No segment found for rate <= AAMP_NORMAL_PLAY_RATE, retrying");
```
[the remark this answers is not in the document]
pstroffolino: Yes, it's true that log flooding every 50ms  could still occur with trace level logging if scenario occurs.  But this is not something that needs to be addressed (at least not now), as trace level logging is not used in production.

We have option to rate limit as per above suggestion.  Himanshu, do you think it's worth bothering?

fragmentcollector_mpd.cpp
```diff
@@ -10000,9 +10005,26 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
+						// blocking in WaitForVideoTsbContentOrAbort by using a bounded, interruptible
+						// sleep before retrying the fetch loop.
+						aamp->interruptibleMsSleep(50);
+					}
+					else
+					{
+						// Unexpected: no segment found at normal/slow rate outside of EOS. Sleep to
+						// avoid a tight loop and throttle log output until the condition resolves.
+						AAMPLOG_TRACE("No segment found for rate <= AAMP_NORMAL_PLAY_RATE, retrying");
+						aamp->interruptibleMsSleep(50);
```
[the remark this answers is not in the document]
pstroffolino: is this valid issue?  If so, how important to address?

fragmentcollector_mpd.cpp
```diff
@@ -10476,6 +10497,13 @@ void StreamAbstractionAAMP_MPD::Stop(bool clearChannelData)
 		abortTsbReader = true;
 		// Signal TsbReader thread to exit wait for manifest update if waiting
 		AbortWaitForManifestUpdate();
+		// Unblock TsbReader if it is waiting for new video TSB content (e.g. during FF
+		// on local AAMP TSB where DisableDownloads is not called and therefore
+		// NotifyVideoTsbWaiters is not triggered via the downloads-disabled path).
+		if (AampTSBSessionManager *tsbMgr = aamp->GetTSBSessionManager())
+		{
+			tsbMgr->NotifyVideoTsbWaiters();
+		}
```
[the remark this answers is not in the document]
pstroffolino: valid issue?

fragmentcollector_mpd.cpp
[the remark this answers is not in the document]
hthakkar-synamedia: I do not think this is a valid issue

PR #613, opened 2025-10-22
VPLAY-11408:Same JS event listeners should not be registered multiple…

varshnie:
… times

Reason for change:Same JS event listeners should not be registered multiple times Test Procedure: Refer jira ticket
Priority: P1


Vinish100 (commented):
Looks good to me 👍

pstroffolino (approved):

PR #615, opened 2025-10-23

lashmintha:
Risks: Low
Priority: P1


fragmentcollector_mpd.cpp
```diff
@@ -7648,13 +7649,61 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 							// chosenAdaptationIdxs.insert(0);
 						}
 					}
+					// Check if representation should be reset for period change
+					// Use adaptation set 0 (video) to check for representation changes
+					bool shouldResetRepresentationIndex = false;
+					int currentrepresentationIndex = pMediaStreamContext->representationIndex;
```
Vinish100: currentRepresentationIndex

fragmentcollector_mpd.cpp
```diff
@@ -7648,13 +7649,61 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 							// chosenAdaptationIdxs.insert(0);
 						}
 					}
+					// Check if representation should be reset for period change
+					// Use adaptation set 0 (video) to check for representation changes
+					bool shouldResetRepresentationIndex = false;
```
Vinish100: shouldResetRepresentationIndex = (representationCount != GetProfileCount())
and the below check is required only in shouldResetRepresentationIndex is false.
Checking against profileCount is a straight forward and if repIdx needs to be reset there is no need to do this complex check

fragmentcollector_mpd.cpp
```diff
@@ -7648,13 +7649,61 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
+					int currentrepresentationIndex = pMediaStreamContext->representationIndex;
+					if (periodChanged && i == eMEDIATYPE_VIDEO && !ISCONFIGSET(eAAMPConfig_AudioOnlyPlayback))
+					{
+						// For video period changes, check if current profile bandwidth exists at the same representation index
+						bool representationChanged = true; // Default to changed
+						if ( adaptationSets.size() > 0)
+						{
+							// Use adaptation set 0 for video only
+							IAdaptationSet* adaptationSet = adaptationSets.at(0);
```
Vinish100: In Intra-Asset streams, there will be multiple video adaptations, so assuming adapIdx 0 might break this logic. You could check if the adapIdx from previous period is valid in adaptationSets.size().
Eg for IA 
SD - 3 profiles
HD - 2 profiles
UHD - 1 profiles
Here, we might be in UHD profile in previous period, checking that and bandwidth against SD profile from new period will always fail

fragmentcollector_mpd.cpp
```diff
@@ -7648,13 +7649,61 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
+								else
+								{
+									representationChanged = true;
+									AAMPLOG_INFO("Representation changed - Current profile bandwidth %" PRIu32 " doesn't match representation bandwidth %u at index %d in adaptationSet 0",
+										currentProfileBandwidth, representationBandwidth, currentrepresentationIndex);
+								}
+							}
+							else
+							{
+								representationChanged = true;
```
Vinish100: representationChanged is by default true, required?

fragmentcollector_mpd.cpp
```diff
@@ -7936,7 +7985,23 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 							}
 							UpdateIframeTracks();
 						}
-						currentProfileIndex = GetDesiredProfile(false);
+
+						if(ISCONFIGSET(eAAMPConfig_RetuneForGSTError) || ISCONFIGSET(eAAMPConfig_ForceMultiPeriodDiscontinuity) || ISCONFIGSET(eAAMPConfig_InternalReTune))
```
Vinish100: Why these checks are added here?

fragmentcollector_mpd.cpp
```diff
@@ -7936,7 +7985,23 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
+						{
+							currentProfileIndex = GetDesiredProfile(false);
```
Vinish100, line 8045: Only during Init() we should call GetDesiredProfile. Can we use periodChanged flag to determine which API to call?

lashmintha: No vinish as during the period change case also, if the Retune or discontinuity is enabled, its failing so that I have included those checking

fragmentcollector_mpd.cpp
lashmintha: During the retune and discontinuity enabled scenario if the UpdateProfileBasedOnFragmentCache() have been facing 'http 404' error

test/utests/tests/StreamAbstractionAAMP_MPD/AudioTrackSelectionTests.cpp
```diff
@@ -498,7 +498,7 @@ R"(<?xml version="1.0" encoding="utf-8"?>
 	</AdaptationSet>
   </Period>
 </MPD>)";
-	fragmentUrl = std::string(TEST_BASE_URL) + std::string("ac4/audio_init.mp4");
+	fragmentUrl = std::string(TEST_BASE_URL) + std::string("aac/audio_init.mp4");
```
pstroffolino, line 501: why this change?

fragmentcollector_mpd.cpp
```diff
@@ -7648,16 +7650,115 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 							// chosenAdaptationIdxs.insert(0);
 						}
 					}
-					if ((representationCount != GetProfileCount()) && mStreamInfo)
+					// Check if representation should be reset for period change
```
pstroffolino, line 7653: at first glance, this amount of code change seems excessive given what we're trying to fix

fragmentcollector_mpd.cpp
```diff
@@ -7648,16 +7650,115 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
+					bool shouldResetRepresentationIndex = (representationCount != GetProfileCount()) && (mStreamInfo != nullptr);
+					bool desiredProfileTobeMaintained = false;
+					if( shouldResetRepresentationIndex == false )
 					{
-						SAFE_DELETE_ARRAY(mStreamInfo);
-
+						int currentRepresentationIndex = pMediaStreamContext->representationIndex;
+						int adaptationSetIndex = pMediaStreamContext->adaptationSetIdx;
+						if (periodChanged && i == eMEDIATYPE_VIDEO && !ISCONFIGSET(eAAMPConfig_AudioOnlyPlayback))
```
pstroffolino, line 7660: is the ISCONFIGSET(eAAMPConfig_AudioOnlyPlayback) check needed?  Here we're already applying change only to video periods

fragmentcollector_mpd.cpp
```diff
@@ -7648,16 +7650,115 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
+						{
+							// For video period changes, check if current profile bandwidth exists at the same representation index
```
pstroffolino, line 7662: wouldn't it be easier to just use the current profile bandwidth as input to abr state when picking starting profile in next period?  Treat it like existing initial profile configuration?

fragmentcollector_mpd.cpp
```diff
@@ -7648,16 +7650,115 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
+							bool representationChanged = true; // Default to changed
+							if ( adaptationSetIndex >=0 && adaptationSetIndex < adaptationSets.size() )
+							{
+								// Use adaptation set 0 for video only
```
pstroffolino, line 7666: what is meaning of this comment?  are we assuming adaptation set 0 is only/always for video?

fragmentcollector_mpd.cpp
```diff
@@ -7648,16 +7650,115 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
+							else
+							{
+								AAMPLOG_INFO("Representation changed - invalid no adaptation sets");
+							}
+							shouldResetRepresentationIndex = representationChanged;
+						}
+					}
+					if (shouldResetRepresentationIndex)
+					{
+						/*If the representation count across the periods changes, currentprofileIndex is invalid
```
pstroffolino, line 7705: this seems overcomplicated and unnecessary - can't we address in acceptable manner without special cases?

pstroffolino (commented):
see comments - would prefer simpler fix if possible; we can have breakout call to discuss options/requirements if needed

fragmentcollector_mpd.cpp
lashmintha: will be optimizing that phill, will share the updated review asap

fragmentcollector_mpd.cpp
lashmintha: Without this condition checking the condition executes and it try to download the video content even in the Audio only streams...!

PR #630, opened 2025-10-28
VPLAY-11554 WPEWebProcess crash due to lockup in GetManifestEndDelta

jameslofthouse-RED:
Reason for change: Avoid deadlock in GetManifestEndDelta() on stop
Test Procedure: see ticket
Risks: Low


priv_aamp.cpp
```diff
@@ -7658,6 +7658,7 @@ void PrivateInstanceAAMP::Stop( bool isDestructing )
 		double bufferedDuration = 0.0;
 		if (mpStreamAbstractionAAMP)
 		{
+			mpStreamAbstractionAAMP->UnblockVideoWaitForCachedFragmentChunk(); // avoid mutex lock if waiting for cached fragments
```
pstroffolino: s the deadlock video-specific or could it occur with audio or text track?

pstroffolino (approved):

PR #634, opened 2025-10-29
VPLAY-11140: Encrypted content support with mp4demux

varshnie:
Reason for change: Removed SUPPORTS_MP4DEMUX macro
Test Procedure: Set 'useMp4Demux=true' and ensure encrypted content plays
Risks: Low
Priority: P1

### DIFF
--- a/middleware/InterfacePlayerRDK.cpp
+++ b/middleware/InterfacePlayerRDK.cpp
@@ -1829,6 +1829,11 @@ void InterfacePlayerRDK::InitializeSourceForPlayer(void *PlayerInstance, void * 
 	{
 		caps = GetCaps(static_cast<GstStreamOutputFormat>(stream->format));
 	}
+	else
+	{
+		MW_LOG_MIL("Skipping caps for now, will be set from mp4Demux later");
+	}
+
 	if (caps != NULL)
 	{
 		gst_app_src_set_caps(GST_APP_SRC(source), caps);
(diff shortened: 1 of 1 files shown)


middleware/InterfacePlayerRDK.cpp
```diff
@@ -3123,10 +3123,10 @@ bool InterfacePlayerRDK::SendHelper(int type, const void *ptr, size_t len, doubl
 			{
 				interfacePlayerPriv->ForwardBuffersToAuxPipeline(buffer, mPauseInjector, this);
 			}
-#ifdef SUPPORTS_MP4DEMUX
 			if( mediaType<2 && m_gstConfigParam->useMp4Demux &&
 			   !copy /* avoid using this path for hls/ts */ )
 			{
+				AAMPLOG_WARN("Using Mp4Demux ");
```
Vinish100: This log will log everytime a fragment is pushed using mp4demux. I would not recommend it here and as a WARN log

middleware/InterfacePlayerRDK.cpp
```diff
@@ -1829,6 +1829,10 @@ void InterfacePlayerRDK::InitializeSourceForPlayer(void *PlayerInstance, void *
 	{
 		caps = GetCaps(static_cast<GstStreamOutputFormat>(stream->format));
 	}
+	else
```
Vinish100: else if (m_gstConfigParam->useMp4Demux)
{
   MW_LOG_MIL("Skipping caps for now, will be set from mp4Demux later");
}

Vinish100 (approved):

PR #637, opened 2025-10-29
STBT-55040 : Fossid scan check 

arepallikrishnasree:
added a comment message in tools.sh and checking for fossid scan


mhughesacn:
Hi @arepallikrishnasree - Is it intended to merge this PR or is it just for testing scans?  If the PR is to be merged, then copyright.c will need some special handling: at least in a separate directory with an explanation in NOTICE.  Is copyright.c ever compiled and linked?

PR #640, opened 2025-10-31
VPLAY-11328 - Linear - Progress Bar jumping to the beginning while trickplay

anjali-syna:
Issue observed is that the Progress Bar Point is jumping to the End/beginning
 while performing Trick Play/Seek in linear services. While analyzing
 the pts value of the segments, it is noticed that the restamped pts
 value of some of the segments in the main content is a very high value
 compared to the previous segment. This is happening because a wrong
 timescale is used for pts re calculation. RecalculatePTS function is
 using the timescale value saved in PrivateInstanceAAMP. When the downloads
 progresses from content to ad the timescale of the segment can change.
 When this happens, if the pay rate is not normal play rate, the new timescale
 is not set to PrivateInstanceAAMP and thus causing the issue.

Changes:
* CacheFragment function is modified to set the timescale to PrivateInstanceAAMP irrespective of the play rate.

Risks: low

### DIFF
--- a/MediaStreamContext.cpp
+++ b/MediaStreamContext.cpp
@@ -146,6 +146,88 @@ bool MediaStreamContext::CacheFragment(std::string fragmentUrl, unsigned int cur
 			}
 		}
 
+		if ((actualType == eMEDIATYPE_INIT_VIDEO || actualType == eMEDIATYPE_INIT_AUDIO || actualType == eMEDIATYPE_INIT_SUBTITLE) && ret) // Only if init fragment successful or available from cache
+		{
+			//To read track_id from the init fragments to check if there any mismatch.
+			//A mismatch in track_id is not handled in the gstreamer version 1.10.4
+			//But is handled in the latest version (1.18.5),
+			//so upon upgrade to it or introduced a patch in qtdemux,
+			//this portion can be reverted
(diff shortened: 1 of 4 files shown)


MediaStreamContext.cpp
```diff
@@ -146,99 +146,97 @@ bool MediaStreamContext::CacheFragment(std::string fragmentUrl, unsigned int cur
...
-				// Not overwriting for subtitles, as subtecmp4transform never read trackId from init fragments
+				if(!trackIdUpdated)
+				{
+					aamp->mCurrentAudioTrackId = track_id;
+				}
+			}
+			// Not overwriting for subtitles, as subtecmp4transform never read trackId from init fragments
+		}
+
+		if (iCurrentRate != AAMP_NORMAL_PLAY_RATE)
```
cpc005, line 231: So this block of code has been moved further down the func and actualType is subsequently not used for anything significant.

anjali-syna: Yes. Since this block of code is moved down, video time scale update will happen inside CacheFragment function even when the rate is not NORMAL_PLAY_RATE and 'actualType' is used only for UpdateVideoEndMetrics now.

PR #645, opened 2025-11-01
VPLAY-11545 OSX runtime failure running AampTSBSessionManager in Xcod…

pstroffolino:
VPLAY-11545 OSX runtime failure running AampTSBSessionManager in Xcode - "the comparison function used by std::sort is not a valid strict-weak ordering function"

Reason for Change: added special case if ad types are identical, to avoid strict-weak ordering exception

Test Guidance: l1 AampTsbSessionManager passing both with run.sh and when run in xcode

Risk: Unknown

### DIFF
--- a/AampTSBSessionManager.cpp
+++ b/AampTSBSessionManager.cpp
@@ -1281,7 +1281,7 @@ std::vector<std::shared_ptr<AampTsbAdMetaData>> AampTSBSessionManager::MergeAndS
 			auto bAdType = b->GetAdType();
 			if( aType == bType && aAdType == bAdType )
 			{ // tie-breaker: avoid not a valid strict-weak ordering exception
-				maintainOrder (a<b);
+				maintainOrder = (a<b);
 			}
 			if ( ((aAdType == bAdType) && (aType == AampTsbAdMetaData::EventType::START)) ||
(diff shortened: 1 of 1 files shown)


AampTSBSessionManager.cpp
```diff
@@ -1273,19 +1271,22 @@ std::vector<std::shared_ptr<AampTsbAdMetaData>> AampTSBSessionManager::MergeAndS
...
 			auto bType = b->GetEventType();
 			auto aAdType = a->GetAdType();
 			auto bAdType = b->GetAdType();
-
-			if ( ((aAdType == bAdType) && (aType == AampTsbAdMetaData::EventType::START)) ||
-				 ((aAdType == AampTsbAdMetaData::AdType::RESERVATION) && bType == AampTsbAdMetaData::EventType::END) ||
-				 ((bAdType == AampTsbAdMetaData::AdType::RESERVATION) && aType == AampTsbAdMetaData::EventType::START) )
+			if( aType == bType && aAdType == bAdType )
+			{ // tie-breaker: avoid not a valid strict-weak ordering exception
+				maintainOrder = a<b;
```
anshephe, line 1282: Could this be simply maintainOrder = false, rather than comparing pointers?
If ( aType == bType && aAdType == bAdType ) is true they are equivalent (in reality shouldn't happen that you have identical advert metadata at the same position)

PR #652, opened 2025-11-04

lashmintha:
Risks: Low
Priority: P1

### DIFF
--- a/fragmentcollector_mpd.cpp
+++ b/fragmentcollector_mpd.cpp
@@ -4074,7 +4074,7 @@ AAMPStatusType StreamAbstractionAAMP_MPD::Init(TuneType tuneType)
 			}
 			// Update track with update in stream info
 			mUpdateStreamInfo = true;
-			ret = UpdateTrackInfo(!newTune, true);
+			ret = UpdateTrackInfo(!newTune, true, true);
 
 			if(eAAMPSTATUS_OK != ret)
 			{
@@ -4442,7 +4442,6 @@ AAMPStatusType StreamAbstractionAAMP_MPD::IndexNewMPDDocument(bool updateTrackIn
 				if(((AdState::IN_ADBREAK_AD_PLAYING != mCdaiObject->mAdState) && (AdState::IN_ADBREAK_WAIT2CATCHUP != mCdaiObject->mAdState))
 				   || (AdState::IN_ADBREAK_AD_PLAYING == mCdaiObject->mAdState && mUpdateStreamInfo))
 				{
-					AAMPLOG_TRACE("Indexing new mpd doc");
 					ret = UpdateTrackInfo(true, true);
 					if(ret != eAAMPSTATUS_OK)
 					{
@@ -7462,7 +7461,7 @@ static bool IsWebmVideoCodec(const std::string &codec )
 /**
  * @brief Updates track information based on current state
  */
-AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW, bool resetTimeLineIndex)
+AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW, bool resetTimeLineIndex, bool isUpdateDuringInit)
 {
 	AAMPStatusType ret = eAAMPSTATUS_OK;
 	long defaultBitrate = aamp->GetDefaultBitrate();
@@ -7610,8 +7609,10 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW, 
 				{
 					mUpdateStreamInfo = false;
 					int representationCount = 0;
+					bool repCountChanged = false;
 					std::set<uint32_t> blAdaptationIdxs;
 					bool bSeenNonWebmCodec = false;
+					std::map<int, long> iBandwidthMap;
 
 					const auto &blacklistedProfiles = aamp->GetBlacklistedProfiles();
 					// Filter the blacklisted profiles in the period
@@ -7648,10 +7649,22 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW, 
 							// chosenAdaptationIdxs.insert(0);
 						}
 					}
-					if ((representationCount != GetProfileCount()) && mStreamInfo)
+					bool shouldResetRepresentationIndex = (representationCount != GetProfileCount()) && (mStreamInfo != nullptr);
+					int currentRepIndex = pMediaStreamContext->representationIndex;
+					// On period change, check if current profile is present in new period
+					if( false == shouldResetRepresentationIndex && mStreamInfo != nullptr)
+					{
+						if (periodChanged && i == eMEDIATYPE_VIDEO && !ISCONFIGSET(eAAMPConfig_AudioOnlyPlayback))
+						{
+							shouldResetRepresentationIndex = ShoudResetRepresentationIndex(chosenAdaptationIdxs);
+							AAMPLOG_WARN("Period changed - representationIndex check: %s, currentProfileIndex to be maintained: %d",
+								shouldResetRepresentationIndex ? "true" : "false", currentProfileIndex);
+						}
+					}
+					if (shouldResetRepresentationIndex)
 					{
 						SAFE_DELETE_ARRAY(mStreamInfo);
-
+                        repCountChanged = (representationCount != GetProfileCount());
 						// reset representationIndex to -1 to allow updating the currentProfileIndex for period change.
 						pMediaStreamContext->representationIndex = -1;
 						AAMPLOG_WARN("representationIndex set to (-1) to find currentProfileIndex");
@@ -7823,6 +7836,14 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW, 
 							}
 						}
 					}
+					//To get the desired profile Index after period change
+					if( repCountChanged == true )
+					{
+						//Getting the desired profile to avoid crash while accessing invalid profile index
+						currentProfileIndex = GetProfileIndexForBandwidth(pMediaStreamContext->fragmentDescriptor.Bandwidth);
+						AAMPLOG_INFO("Current profile index after period change: %d Bandwidth %u", currentProfileIndex, pMediaStreamContext->fragmentDescriptor.Bandwidth);
+					}
+
 					if(adaptationSets.size() > 0)
 					{
 						// Skipping blacklist check for audio only track at the moment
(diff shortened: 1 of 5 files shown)


support/aampabr/ABRManager.cpp
```diff
@@ -730,3 +739,67 @@ int  ABRManager::getProfileIndexForLowestBandwidth()
...
+ * This function compares the incoming iBandwidthMap (from manifest representations) 
+ * with the current mProfiles in ABR to detect if the available representations have changed.
+ * This is typically needed during manifest refresh to detect profile changes.
+ * 
+ * @param iBandwidthMap Map of index to bandwidth from manifest representations
+ * @param currentProfileIndex Current profile index being used
+ * @param currentProfileIndexToBeMaintained Pointer to output - index to maintain if representations changed
+ * @return bool True if representation change detected, false otherwise
+ */
+bool ABRManager::CheckRepresentationChange(const std::map<int, long>& iBandwidthMap,
```
nu641001: Make this API name more generic, representation is more of a mpeg-dash term in this context, isn't it?

support/aampabr/ABRManager.h
```diff
@@ -350,6 +348,10 @@ class ABRManager {
    * @brief Default iframe bitrate
    */
   long mDefaultIframeBitrate;
+  /**
+   * @brief Desired profile index maintenance flag
+   */
+  bool isDesiredProfileNeeded;
```
nu641001: Please remove unused member

fragmentcollector_mpd.cpp
```diff
@@ -7666,6 +7697,8 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 					int addedProfiles = 0;
 					int idx = 0;
 					std::map<int, struct ProfileInfo> iProfileMaps;
+
```
nu641001: Remove new line

fragmentcollector_mpd.cpp
```diff
@@ -7760,6 +7793,7 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 						}
 					}
 					mProfileCount = idx;
+
```
nu641001: Remove this

fragmentcollector_mpd.cpp
```diff
@@ -7936,7 +7980,24 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 							}
 							UpdateIframeTracks();
 						}
-						currentProfileIndex = GetDesiredProfile(false);
+						//currentProfileIndex = GetDesiredProfile(false);
```
nu641001: Remove this, Document both cases in the same area

nu641001:
Could you please also check coverity issues?

fragmentcollector_mpd.cpp
```diff
@@ -7823,6 +7864,16 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 							}
 						}
 					}
+					//To get the desired profile Index after period change
+					if( repCountChanged == true )
+					{
+						//Getting the desired profile to avoid crash while accessing invalid profile index
+						currentProfileIndex = GetProfileIndexForBandwidth(pMediaStreamContext->fragmentDescriptor.Bandwidth);
+						AAMPLOG_INFO("Current profile index after period change: %d Bandwidth %u", currentProfileIndex, pMediaStreamContext->fragmentDescriptor.Bandwidth);
+						//To rampup to desiredprofile from the ABR set the below flag to true
```
Vinish100: What is the use of this?

fragmentcollector_mpd.cpp
```diff
@@ -7936,7 +7987,23 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
 							UpdateIframeTracks();
 						}
-						currentProfileIndex = GetDesiredProfile(false);
+						/* During initialization its better to call the GetDesiredProfile(), which picks up the
+						* desired profile index corresponding to default bitrate.
+						* During subsequent 'period' change its better to call the  UpdateProfileBasedOnFragmentCache()
+						* which will Rampup and pick the highest profile index if the network bandwidth is high.
+						* and also RampsDown to the lowest profile if the network bandwidth is low.
+						*/
+						if(mFirstTuneDone == false)
```
Vinish100: Isn't it better to pass as an argument to UpdateTrackInfo, rather than a variable which is only used for this specific logic

fragmentcollector_mpd.cpp
```diff
@@ -7648,10 +7651,48 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
+							* in the new period.
+							*/					
+							for (const uint32_t &adaptIdx : chosenAdaptationIdxs)
+							{
+								IAdaptationSet* adaptationSet = adaptationSets.at(adaptIdx);
+								AAMPLOG_INFO("Processing AdaptationSet index:%d", adaptIdx);
+								const auto &representations = adaptationSet->GetRepresentation();
+								for (int reprIdx = 0; reprIdx < representations.size(); reprIdx++)
+								{
+									if( representations.at(reprIdx)->GetBandwidth() == mStreamInfo[idx].bandwidthBitsPerSecond )
```
Vinish100: Why idx and reprIdx are both incremented here? Are you checking if the representation are in the same order in the new adaptations?

As per the example at line 7665, you should be checking single value in all new adaptations right?

fragmentcollector_mpd.cpp
```diff
@@ -7648,10 +7651,48 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
...
 					}
-					if ((representationCount != GetProfileCount()) && mStreamInfo)
+					bool representationChanged = true;
+					bool shouldResetRepresentationIndex = (representationCount != GetProfileCount()) && (mStreamInfo != nullptr);
+					// On period change, check if current profile is present in new period
+					if( false == shouldResetRepresentationIndex && mStreamInfo != nullptr)
+					{
+						if (periodChanged && i == eMEDIATYPE_VIDEO && !ISCONFIGSET(eAAMPConfig_AudioOnlyPlayback))
+						{
+							int idx = 0;
```
Vinish100: Considering the complexity of UpdateTrackInfo function, I would recommend keeping this as a separate member function for better test coverage and readability

fragmentcollector_mpd.cpp
lashmintha: For some streams there is case of representation split across multiple Adaptation sets,

Eg :
        <Adaptation 1>
<Rep 1>
<Rep 2>
        <Adaptation 2>
<Rep 3>
<Rep 4>

In the mstreamInfo[4], we have cummulative set of all the 4 representation count, in order to achieve the better comparison I am using like this

fragmentcollector_mpd.cpp
lashmintha: When the representation count changes across periods, 

Eg 1:
<Period 0>
<Rep 512k>
<Rep 1M>

<Period 1>
<Rep 512k>
<Rep 1M>
<Rep 1.5M>
<Rep 2M>

While switching to <period 1>, even when the network bandwidth is good enough to play <Rep 2M>, it won't ramping up to the desired profile index due to that "mNwConsistencyBypass= false", its set as false, so during that case only I am setting that flag to true

fragmentcollector_mpd.h
```diff
@@ -573,6 +573,13 @@ class StreamAbstractionAAMP_MPD : public StreamAbstractionAAMP
 	 * @brief Clears the mFirstPTS value to trigger update of first PTS
 	 */
 	void clearFirstPTS(void) override;
+	/**
```
pstroffolino, line 576: doxygen comments describe two parameters, but method only has one

fragmentcollector_mpd.cpp
```diff
@@ -14400,3 +14440,47 @@ void StreamAbstractionAAMP_MPD::clearFirstPTS(void)
...
+	* Period 2: AdaptationSet 0 : Representations : 1000, 3000
+	*           AdaptationSet 1 : Representations : 2000, 4000
+	* Here if current profile is 2000, we need to check both the adaptation sets to find the current profile
+	* in the new period.
+	*/	
+	bool representationChanged = false;
+	int repsize	= 0, reprIdx = 0;
+	for (const uint32_t &adaptIdx : chosenAdaptationIdxs)
+	{
+		IAdaptationSet* adaptationSet = mCurrentPeriod->GetAdaptationSets().at(adaptIdx);
```
pstroffolino, line 14466: can be const or const reference?

fragmentcollector_mpd.cpp
```diff
@@ -14400,3 +14440,47 @@ void StreamAbstractionAAMP_MPD::clearFirstPTS(void)
...
+		const auto &representations = adaptationSet->GetRepresentation();
+		
+		for (reprIdx = 0; reprIdx < representations.size(); reprIdx++)
+		{
+			if( ( representations.at(reprIdx)->GetBandwidth() != mStreamInfo[repsize+reprIdx].bandwidthBitsPerSecond )  )
```
pstroffolino, line 14471: this looking for exact bandwidth matches.  Overkill?

fragmentcollector_mpd.cpp
```diff
@@ -9431,6 +9467,8 @@ bool StreamAbstractionAAMP_MPD::SelectSourceOrAdPeriod(bool &periodChanged, bool
 {
 	bool ret = false;
 	waitForAdBreakCatchup = false; // reset this flag
+	AAMPLOG_INFO("SelectSourceOdAdPeriod: mIterPeriodIndex=%d, mNumberOfPeriods=%d, mCurrentPeriodIdx=%d, mBasePeriodId=%s, mPlayRate=%f",
```
pstroffolino, line 9470: typo: Od = Or

fragmentcollector_mpd.cpp
```diff
@@ -4074,7 +4074,7 @@ AAMPStatusType StreamAbstractionAAMP_MPD::Init(TuneType tuneType)
 			}
 			// Update track with update in stream info
 			mUpdateStreamInfo = true;
-			ret = UpdateTrackInfo(!newTune, true);
+			ret = UpdateTrackInfo(!newTune, true, true);
```
pstroffolino, line 4077: might be worth adding comment after each parameter to UpdateTrackInfo to document name/purpose

fragmentcollector_mpd.cpp
```diff
@@ -7823,6 +7836,14 @@ AAMPStatusType StreamAbstractionAAMP_MPD::UpdateTrackInfo(bool modifyDefaultBW,
 							}
 						}
 					}
+					//To get the desired profile Index after period change
+					if( repCountChanged == true )
+					{
+						//Getting the desired profile to avoid crash while accessing invalid profile index
+						currentProfileIndex = GetProfileIndexForBandwidth(pMediaStreamContext->fragmentDescriptor.Bandwidth);
```
pstroffolino, line 7843: wouldn't this be equivalent and safe to do always?  do we really need preceding logic and repCountChanged==true check?

fragmentcollector_mpd.cpp
```diff
@@ -14400,3 +14440,47 @@ void StreamAbstractionAAMP_MPD::clearFirstPTS(void)
...
+	* for finding the current profile in the new period.
+	* Example:
+	* Period 1: AdaptationSet 0 : Representations : 1000, 2000, 3000
+	* Period 2: AdaptationSet 0 : Representations : 1000, 3000
+	*           AdaptationSet 1 : Representations : 2000, 4000
+	* Here if current profile is 2000, we need to check both the adaptation sets to find the current profile
+	* in the new period.
+	*/	
+	bool representationChanged = false;
+	int repsize	= 0, reprIdx = 0;
```
pstroffolino, line 14463: "repsize" is being flagged as a possible typo by godspell tool
If you rename to repSize or representationSize, issue should be avoided without having to add to Godspell's whitelist

fragmentcollector_mpd.cpp
lashmintha: Consider the below case,
<period 0>
     <Rep 0>
     <Rep 1>
     <Rep 2>
     <Rep 3>
    
<Period 1>
     <Rep 0>
     <Rep 1>
     
     As the currentProfileIndex=3, sets for the period 0, so it try to access the same index after switching to the new <period 1> so facing array out of boundry issue, in order to avoid this resetting the currentProfileIndex..!

PR #666, opened 2025-11-11
VPLAY-11709 l2 test 2073 regression

pstroffolino:
VPLAY-11709 l2 test 2073 regression

Reason for Change: workaround for regression from VPLAY-11588 impacting some HLS playback (audio-only hls playlist using aac in this case)

Test Guidance: l2 test 2073 passing again

Risk: High


pstroffolino:
redundant with PR#664

PR #670, opened 2025-11-13
VPLAY-11143: AAMP mp4Demux Integration

Vinish100:
Reason for change: Integrate mp4Demux formally into AAMP
Test Procedure: Test with useMp4Demux=true and validate all types of Linear, VOD content in all apps.
Risks: Medium


mp4demux/AampMp4Demuxer.h
```diff
@@ -0,0 +1,132 @@
...
+	 */
+	void setThrottleEnable(bool enable) override { }
+
+	/**
+	 * @brief Set frame rate for trickmode
+	 *
+	 * @param[in] frameRate - rate per second
+	 * @return void
+	 */
+	void setFrameRateForTM (int frameRate) override { }
```
pstroffolino, line 93: should spell out "TrickMode" rather than using "TM" abbreviation

mp4demux/AampMp4Demuxer.h
```diff
@@ -0,0 +1,132 @@
...
+
+	/**
+	 * @brief given TS media segment (not yet injected), extract and report first PTS
+	 */
+	double getFirstPts( AampGrowableBuffer* pBuffer ) override { return 0.0; };
+
+	/**
+	 * @brief optionally specify new pts offset to apply for subsequently injected TS media segments
+	 */
+	void setPtsOffset( double ptsOffset ) override { };
```
pstroffolino, line 52: assuming this for non-invasive pts restamping?

mp4demux/AampMp4Demuxer.h
Vinish100: Will take it in a different ticket. Too many changes in this one as-is

mp4demux/AampMp4Demuxer.h
Vinish100: Yes. We need to implement in Phase 2

middleware/InterfacePlayerRDK.h
```diff
@@ -147,6 +146,100 @@ enum class InterfaceCB
 	startNewSubtitleStream // Add more events here if needed
 };
 
+// Codec information structure
```
DomSyna: This looks like duplicate definition of AampCodecInfo. This adds both a maintenance burden, potential for inconsistency.

Would it be better to to put things like this in a common header file ?

middleware/GstUtils.h
```diff
@@ -66,6 +66,7 @@ enum GstStreamOutputFormat
         GST_FORMAT_ISO_BMFF,        /**< ISO Base Media File format */
         GST_FORMAT_AUDIO_ES_MP3,    /**< MP3 Audio Elementary Stream */
         GST_FORMAT_AUDIO_ES_AAC,    /**< AAC Audio Elementary Stream */
+        GST_FORMAT_AUDIO_ES_AAC_RAW,/**< AAC Raw Audio Elementary Stream */
```
jfagunde: Would it be possible to move adding this new stream output format to its own pull request? It seems fairly self-contained and it would reduce this PR.

aampgstplayer.cpp
```diff
@@ -1312,3 +1329,76 @@ void AAMPGstPlayer::StopMonitorAvTimer()
...
+
+/**
+ * @brief Set stream capabilities based on codec info
+ *
+ * @param[in] type - Media type
+ * @param[in] codecInfo - Codec information
+ */
+void AAMPGstPlayer::SetStreamCaps(AampMediaType type, AampCodecInfo &&codecInfo)
+{
+	CodecInfo gstCodecInfo;
```
DomSyna: If aamp used the midlewares CodecInfo throughout we wouldn't need methods like this as its basically doing a move from AAMPCodecInfo to CodecInfo

If we can't use the same struct we could use static_assert to validate the structs are identical at compile time. Then in here we can just

std::memcpy(&gstCodecInfo, &codecInfo, sizeof(CodecInfo));
// Move codecData last, after all other fields have been accessed
gstCodecInfo.codecData = std::move(codecInfo.mCodecData);
playerInstance->SetStreamCaps(static_cast<GstMediaType>(type), gstCodecInfo);

AampDemuxDataTypes.h
```diff
@@ -0,0 +1,257 @@
...
+	// Delete copy constructor and copy assignment to prevent accidental copies
+	AampDrmMetadata(const AampDrmMetadata&) = delete;
+	AampDrmMetadata& operator=(const AampDrmMetadata&) = delete;
+};
+
+/*
+ * @struct AampMediaSample
+ * @brief Media sample structure
+ */
+struct AampMediaSample
```
jfagunde, line 34: How about moving the introduction of this structure to its own pull request, separate from the mp4demux code, to split this pull request into more manageable size ones?

middleware/InterfacePlayerRDK.cpp
```diff
@@ -5248,3 +5248,197 @@ double InterfacePlayerRDK::FlushTrack(int mediaType, double pos, double audioDel
...
+ * @brief Decorate a GstBuffer with DRM metadata
+ * @param[in] buffer The GstBuffer to decorate
+ * @param[in] drmMetadata The DRM metadata
+ */
+static void DecorateGstBufferWithDrmMetadata(GstBuffer *buffer, const MediaDrmMetadata &drmMetadata)
+{
+	GstStructure *metadata = NULL;
+	if (drmMetadata.isEncrypted)
+	{
+		metadata = gst_structure_new(
```
DomSyna, line 5365: It looks like there is a lot or repetition here, could you refactor to use a helper function ?

maybe something like
static bool AddBufferFieldToStructure(GstStructure *metadata, const char *fieldName, const void *data, gsize size)
{
GstBuffer *buffer = CreateGstBufferWithData(data, size);
if (!buffer) return false;
gst_structure_set(metadata, fieldName, GST_TYPE_BUFFER, buffer, NULL);
gst_buffer_unref(buffer);
return true;
}

middleware/InterfacePlayerRDK.h
DomSyna: oh AAMPCodecInfo has more fields in the audio / video is this intentional ?

aampgstplayer.cpp
DomSyna: Just noticed the codecInfo structs are actually different.

mp4demux/MP4Demux.cpp
```diff
@@ -0,0 +1,1285 @@
...
+ * - dec3: Enhanced AC3 audio
+ * 
+ * @param fourCC Four character code from MP4 container
+ * @return StreamOutputFormat corresponding to the codec type
+ */
+StreamOutputFormat Mp4Demux::GetStreamOutputFormatFromFourCC(const uint32_t fourCC)
+{
+	switch (fourCC)
+	{
+		case MultiChar_Constant("avcC"):
```
DomSyna: Maybe iterate over
constexpr FourCCMapping codecMappings[] = {
	{ MultiChar_Constant("avcC"), FORMAT_VIDEO_ES_H264 },
	{ MultiChar_Constant("hvcC"), FORMAT_VIDEO_ES_HEVC },
	{ MultiChar_Constant("esds"), FORMAT_AUDIO_ES_AAC_RAW },
	{ MultiChar_Constant("dec3"), FORMAT_AUDIO_ES_EC3 }
};
to find the format ?
better from a SOLID pov

mp4demux/MP4Demux.cpp
```diff
@@ -0,0 +1,1285 @@
...
+	}
+	uint64_t dts = baseMediaDecodeTime;
+	for (auto i = 0u; i < sampleCount; i++)
+	{
+		samples.emplace_back();
+		// Get reference to newly added sample
+		AampMediaSample& newSample = samples.back();
+		uint32_t sampleLen = defaultSampleSize;
+		uint32_t sampleDuration = defaultSampleDuration;
+		if (flags & 0x0100)
```
DomSyna: These flags would be better as defines and not magic numbers. Maybe copilot can sort it out ?

mediaprocessor.h
DomSyna, line 55: violates ISP as it forces implementations to provide many methods they don't need
Probably not part of this work, but the code would benefit from breaking the class up into functional block

mp4demux/MP4Demux.cpp
```diff
@@ -0,0 +1,1285 @@
...
+/**
+ * @brief Get DRM protection system data
+ * Returns all PSSH (Protection System Specific Header) data
+ * extracted from the MP4 container for DRM license acquisition.
+ * 
+ * @return Protection data vector with ownership transferred to caller
+ */
+std::vector<AampPsshData> Mp4Demux::GetProtectionEvents()
+{
+	return std::move(protectionData);
```
DomSyna, line 1273: std::vector<AampPsshData> result;
result.swap(protectionData);
return result;

Don't be tempted to return std::move(result); as it will stop the compiler doing NRVO and move is the natural fallback if NVRO isn't possible

middleware/InterfacePlayerRDK.h
Vinish100: No, that will be cleaned up

middleware/GstUtils.h
Vinish100: Done. VPLAY-12257

mp4demux/MP4Demux.cpp
Vinish100: As these are member variables, compiler doesn't attempt NVRO as per copilot. So a explicit std::move is required. Using rvalue references for return value is not suggested as well..

aampgstplayer.cpp
Vinish100: Done

middleware/InterfacePlayerRDK.h
Vinish100: Done

mp4demux/MP4Demux.cpp
Vinish100: Done

middleware/InterfacePlayerRDK.cpp
Vinish100: Done

PR #677, opened 2025-11-14
RDKEMW-6162 : Update aamp with libds client library API's

varatharajan568:
Reason for change:Update aamp related DS MGR client to use libds client call
Risks: Low
Test Procedure: Need to ensure the HDMI Hotplug, HDCP status, Res post scenario's
Priority: P1


middleware/externals/rdk/PlayerExternalsRdkInterface.h
```diff
@@ -35,7 +35,8 @@
 #include "dsAudio.h"
 
 #include <memory>
-
+#ifdef USE_DS_EVENT_SUPPORTED
```
Vinish100: What is this for?

pstroffolino (approved):

PR #684, opened 2025-11-17
VPLAY-11776: Fix for crash due to curl operations

Abhi-jith-S:
Reason for change:Improvements in curl operations
Test Procedure:Refer ticket
Priority: P2
Risks:Low

### DIFF
--- a/downloader/AampCurlDownloader.cpp
+++ b/downloader/AampCurlDownloader.cpp
@@ -137,6 +137,7 @@ AampCurlDownloader::AampCurlDownloader() : mCurlMutex(),m_threadName(""),mDownlo
 
 AampCurlDownloader::~AampCurlDownloader()
 {
+	std::lock_guard<std::mutex> lock(mCurlMutex);
 	mDownloadActive = false;
 	
 	if(mCreatedNewFd && mCurl)
@@ -163,6 +164,7 @@ int AampCurlDownloader::Download(const std::string &urlStr, std::shared_ptr<Down
 	if(urlStr.size() == 0 || dnldData == nullptr)
 	{
 		AAMPLOG_ERR("Invalid inputs provided for download . Check the arguments. Url[%s] dnldData is Null[%d]", urlStr.c_str(), (dnldData == nullptr));
+		return CURLE_BAD_FUNCTION_ARGUMENT; //return an appropriate error code
 	}
 	else if(mCurl)
 	{
@@ -174,15 +176,21 @@ int AampCurlDownloader::Download(const std::string &urlStr, std::shared_ptr<Down
 				mDownloadResponse = dnldData;
 				mDownloadResponse->sEffectiveUrl = urlStr;
 				CURL_EASY_SETOPT_STRING(mCurl, CURLOPT_URL, urlStr.c_str());
+				CURL_EASY_SETOPT_LONG(mCurl, CURLOPT_TIMEOUT, 30L);        // total timeout
+				CURL_EASY_SETOPT_LONG(mCurl, CURLOPT_CONNECTTIMEOUT, 10L); // connection timeout 
 			}
 			bool loopAgain = false;
+			int backoffMs = 1000; // start with 1s
 			do{
 				mDownloadStartTime = mDownloadUpdatedTime = NOW_STEADY_TS_MS;
 				if( mDnldCfg && mDnldCfg->bCurlThroughput )
 				{
 					AAMPLOG_MIL( "curl-begin type=%d", eMEDIATYPE_MANIFEST);
 				}
+				{
+				std::lock_guard<std::mutex> lock(mCurlMutex);
 				httpRetVal = curl_easy_perform(mCurl);
+				}
 				loopAgain = false;
 				numDownloadAttempts++;
 				if(httpRetVal == CURLE_OK)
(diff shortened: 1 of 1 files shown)


downloader/AampCurlDownloader.cpp
```diff
@@ -174,15 +177,21 @@ int AampCurlDownloader::Download(const std::string &urlStr, std::shared_ptr<Down
 				mDownloadResponse = dnldData;
 				mDownloadResponse->sEffectiveUrl = urlStr;
 				CURL_EASY_SETOPT_STRING(mCurl, CURLOPT_URL, urlStr.c_str());
+				CURL_EASY_SETOPT_LONG(mCurl, CURLOPT_TIMEOUT, 30L);        // total timeout
```
vinodkadungoth, line 180: Please avoid hardcoding. If there is an existing init configuration or macro for CURLOPT_TIMEOUT, we should use it; otherwise, we need to define a new one.I believe we should have existing one

downloader/AampCurlDownloader.cpp
```diff
@@ -174,15 +177,21 @@ int AampCurlDownloader::Download(const std::string &urlStr, std::shared_ptr<Down
...
+				CURL_EASY_SETOPT_LONG(mCurl, CURLOPT_CONNECTTIMEOUT, 10L); // connection timeout
```
vinodkadungoth, line 181: Please avoid hardcoding. If there is an existing init configuration or macro for CURLOPT_TIMEOUT, we should use it. otherwise, we need to define a new one.I believe we should have existing one

downloader/AampCurlDownloader.cpp
```diff
@@ -174,15 +177,21 @@ int AampCurlDownloader::Download(const std::string &urlStr, std::shared_ptr<Down
 				mDownloadResponse = dnldData;
 				mDownloadResponse->sEffectiveUrl = urlStr;
 				CURL_EASY_SETOPT_STRING(mCurl, CURLOPT_URL, urlStr.c_str());
+				CURL_EASY_SETOPT_LONG(mCurl, CURLOPT_TIMEOUT, 30L);        // total timeout
+				CURL_EASY_SETOPT_LONG(mCurl, CURLOPT_CONNECTTIMEOUT, 10L); // connection timeout 
 			}
 			bool loopAgain = false;
+			int backoffMs = 1000; // start with 1s
```
vinodkadungoth, line 184: Please avoid hardcoding. I suggest to add a macro.

downloader/AampCurlDownloader.cpp
```diff
@@ -235,6 +252,9 @@ int AampCurlDownloader::Download(const std::string &urlStr, std::shared_ptr<Down
 						if (httpRetVal == CURLE_COULDNT_CONNECT || IsCurlTimeoutFailure (httpRetVal) )
 						{
 							loopAgain = true;
+							std::this_thread::sleep_for(std::chrono::milliseconds(backoffMs));
+    						backoffMs = std::min(backoffMs * 2, 16000); // cap at 16s
+							AAMPLOG_WARN("Testing-->BackOffMs due curl errors %d = %d", httpRetVal, backoffMs);
```
vinodkadungoth, line 257: Remove "Testing-->"

downloader/AampCurlDownloader.cpp
```diff
@@ -159,10 +160,12 @@ int AampCurlDownloader::Download(const std::string &urlStr, std::shared_ptr<Down
 {
 	int httpRetVal=0;
 	int numDownloadAttempts=0;
-	int numRetriesAllowed = mDnldCfg?mDnldCfg->iDownloadRetryCount:0;
+	//int numRetriesAllowed = mDnldCfg?mDnldCfg->iDownloadRetryCount:0; commented out for validation
+	int numRetriesAllowed = 2;
```
vinodkadungoth, line 164: it should be configurable.

downloader/AampCurlDownloader.cpp
```diff
@@ -218,8 +230,13 @@ int AampCurlDownloader::Download(const std::string &urlStr, std::shared_ptr<Down
 								mDownloadResponse->iHttpRetValue >= 500 ))
 						{
 							AAMPLOG_WARN("Download failed due to Server error http-%d numDownloadAttempts %d numRetriesAllowed %d",mDownloadResponse->iHttpRetValue,numDownloadAttempts,numRetriesAllowed);
-							int retryDelayMs = (mDownloadResponse->iHttpRetValue == 502) ? MIN_DELAY_BETWEEN_MANIFEST_UPDATE_FOR_502_MS : mDnldCfg->iDownloadRetryWaitMs;
+							int retryDelayMs = (mDownloadResponse->iHttpRetValue == 502) ? MIN_DELAY_BETWEEN_MANIFEST_UPDATE_FOR_502_MS : backoffMs;
 							std::this_thread::sleep_for(std::chrono::milliseconds(retryDelayMs));
+							// Update backoff for next retry (only if not using fixed 502 delay)
+                			if (mDownloadResponse->iHttpRetValue != 502) {
+								backoffMs = std::min(backoffMs * 2, 16000);
```
vinodkadungoth, line 237: backoffMs cap is 16 seconds is high value right? , pls check with @Vinish100

downloader/AampCurlDownloader.cpp
```diff
@@ -235,6 +252,9 @@ int AampCurlDownloader::Download(const std::string &urlStr, std::shared_ptr<Down
 						if (httpRetVal == CURLE_COULDNT_CONNECT || IsCurlTimeoutFailure (httpRetVal) )
 						{
 							loopAgain = true;
+							std::this_thread::sleep_for(std::chrono::milliseconds(backoffMs));
+    						backoffMs = std::min(backoffMs * 2, 16000); // cap at 16s
```
cpc005, line 256: backoffMs  is re-calculated in 2 locations with a const value. It would be better if it was updated in just 1 location.

here in this order: sleep for backoff, recalculate backoff, log backoff value
 It would seem better to:  log backoff, sleep backoff, recalculate

Abhi-jith-S:
Closing this PR as this fix is not applicable to the exact crash issue.

PR #690, opened 2025-11-19
VPLAY-11618 Handle empty language in HLS manifest

DomSyna:
Reason for Change: When HLS CC track is missing the language the current behaviour is to create a language based on the track name which is wrong.

Summary of Changes:

Change the manifest parsing to not populate the language for the CC track if it is not present in the manifest.
Change the subtitle track selection to support a list of of languages provided via the config, or setPreferredTextLanguage()  i.e ["eng",""] allowing the matching of the empty string for non specified language

Risk: Low

### DIFF
--- a/fragmentcollector_hls.cpp
+++ b/fragmentcollector_hls.cpp
@@ -7582,23 +7582,24 @@ bool StreamAbstractionAAMP_HLS::SelectPreferredTextTrack(TextTrackInfo &selected
 
 	unsigned long long bestScore = 0;
 	bool bestTrackFound = false;
+	auto languageVectorToCheck = (aamp->preferredTextLanguagesList.empty()) ? aamp->preferredSubtitleLanguageVctr : aamp->preferredTextLanguagesList;
 
 	for (const auto &track : availableTracks)
 	{
 		unsigned long long score = 1; // Base score for any track
 
 		// Score language preference (higher priority = higher score)
-		if (!aamp->preferredTextLanguagesList.empty())
+		if (!languageVectorToCheck.empty())
 		{
 			std::string normalizedTrackLanguage =
 				(track.language.empty()) ? track.language : Getiso639map_NormalizeLanguageCode(track.language, aamp->GetLangCodePreference());
-			auto iter = std::find(aamp->preferredTextLanguagesList.cbegin(),
-								  aamp->preferredTextLanguagesList.cend(),
+			auto iter = std::find(languageVectorToCheck.cbegin(),
+								  languageVectorToCheck.cend(),
 								  normalizedTrackLanguage);
-			if (iter != aamp->preferredTextLanguagesList.cend())
+			if (iter != languageVectorToCheck.cend())
 			{
-				size_t position = std::distance(aamp->preferredTextLanguagesList.cbegin(), iter);
-				size_t priorityMultiplier = aamp->preferredTextLanguagesList.size() - position;
+				size_t position = std::distance(languageVectorToCheck.cbegin(), iter);
+				size_t priorityMultiplier = languageVectorToCheck.size() - position;
 				score += priorityMultiplier * AAMP_LANGUAGE_SCORE;
 
 				AAMPLOG_TRACE("Track '%s' lang='%s' (normalized='%s') matches position %zu (bonus: %llu)",
--- a/fragmentcollector_hls.cpp
+++ b/fragmentcollector_hls.cpp
@@ -7590,7 +7590,8 @@ bool StreamAbstractionAAMP_HLS::SelectPreferredTextTrack(TextTrackInfo &selected
 		// Score language preference (higher priority = higher score)
 		if (!aamp->preferredTextLanguagesList.empty())
 		{
-			std::string normalizedTrackLanguage = Getiso639map_NormalizeLanguageCode(track.language, aamp->GetLangCodePreference());
+			std::string normalizedTrackLanguage =
+				(track.language.empty()) ? track.language : Getiso639map_NormalizeLanguageCode(track.language, aamp->GetLangCodePreference());
 			auto iter = std::find(aamp->preferredTextLanguagesList.cbegin(),
 								  aamp->preferredTextLanguagesList.cend(),
 								  normalizedTrackLanguage);
--- a/fragmentcollector_hls.cpp
+++ b/fragmentcollector_hls.cpp
@@ -7605,7 +7605,7 @@ bool StreamAbstractionAAMP_HLS::SelectPreferredTextTrack(TextTrackInfo &selected
 		if (!languageVectorToCheck.empty())
 		{
 			std::string normalizedTrackLanguage =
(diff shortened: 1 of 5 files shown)


fragmentcollector_hls.cpp
```diff
@@ -580,13 +580,19 @@ AAMPStatusType StreamAbstractionAAMP_HLS::ParseMainManifest()
...
+						if (audioMedia.language.empty())
+						{ // handle non-compliant manifest missing language attribute
+							AAMPLOG_WARN("Audio track %d missing language, using name '%s' as fallback",
+										 mMediaCount, audioMedia.name.c_str());
+							audioMedia.language = audioMedia.name;
+						}
+						if (!audioMedia.language.empty())
+						{
+							mLangList.insert(GetLanguageCode(mMediaCount));
+						}
```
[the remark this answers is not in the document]
DomSyna: Not required as this is obvious from the warning since it outputs the name

priv_aamp.cpp
```diff
@@ -12522,7 +12546,23 @@ void PrivateInstanceAAMP::SetPreferredTextLanguages(const char *param)
...
+							
+							if (closedCaptionTrackId >= 0)
+							{
+								SetPreferredTextTrack(std::move(selectedTextTrack));
+								AAMPLOG_INFO("Selected text track at index %d (lang=%s)", 
+								             closedCaptionTrackId, selectedTextTrack.language.c_str());
+							}
+							else
+							{
+								AAMPLOG_ERR("Selected text track not found in trackInfo vector");
```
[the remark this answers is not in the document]
DomSyna: Not expected to happen just here for defensive coding

fragmentcollector_hls.cpp
```diff
@@ -580,13 +580,19 @@ AAMPStatusType StreamAbstractionAAMP_HLS::ParseMainManifest()
...
-					{ // handle non-compliant manifest missing language attribute
-						mediaInfoStore[mMediaCount].language =  mediaInfoStore[mMediaCount].name;
-					}
-					if (mediaInfoStore[mMediaCount].type == eMEDIATYPE_AUDIO && !mediaInfoStore[mMediaCount].language.empty() )
+					if (mediaInfoStore[mMediaCount].type == eMEDIATYPE_AUDIO)
 					{
-						mLangList.insert(GetLanguageCode(mMediaCount));
+						auto& audioMedia = mediaInfoStore[mMediaCount];
+						if (audioMedia.language.empty())
+						{ // handle non-compliant manifest missing language attribute
```
dnfaulkner, line 587: Perhaps add a comment:
"but only for eMEDIATYPE_AUDIO and intentionally not for eMEDIATYPE_SUBTITLES"

dnfaulkner (commented):
Just one small comment

fragmentcollector_hls.cpp
DomSyna: Isn't that clear from 
if (mediaInfoStore[mMediaCount].type == eMEDIATYPE_AUDIO)
and the variable name audioMedia ?

fragmentcollector_hls.cpp
```diff
@@ -7538,40 +7559,86 @@ void StreamAbstractionAAMP_HLS::SelectSubtitleTrack()
...
+		unsigned long long score = 1; // Base score for any track
 
-		// Check for language match
-		if (!aamp->preferredTextLanguagesString.empty() && track.language == aamp->preferredTextLanguagesString)
+		// Score language preference (higher priority = higher score)
+		if (!aamp->preferredTextLanguagesList.empty())
 		{
-			score += AAMP_LANGUAGE_SCORE; // Add score for language match
+			std::string normalizedTrackLanguage =
+				(track.language.empty()) ? track.language : Getiso639map_NormalizeLanguageCode(track.language, aamp->GetLangCodePreference());
```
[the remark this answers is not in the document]
DomSyna: We may want to match the empty string in the preferred list with an empty string from the manifest (i.e the language wasn't specified)

fragmentcollector_hls.cpp
```diff
@@ -7538,40 +7573,89 @@ void StreamAbstractionAAMP_HLS::SelectSubtitleTrack()
...
+			bestTrackFound = true;
 			selectedTextTrack = track;
+
+			AAMPLOG_INFO("New best text track: lang=%s, rendition=%s, name=%s, score=%llu",
+						 track.language.c_str(), track.rendition.c_str(),
+						 track.name.c_str(), score);
 		}
 	}
+
+	if (!bestTrackFound)
```
cpc005: Could eliminate bestTrackFound and use bestScore!=0

fragmentcollector_hls.cpp
```diff
@@ -7538,40 +7573,89 @@ void StreamAbstractionAAMP_HLS::SelectSubtitleTrack()
...
+	{
+		AAMPLOG_WARN("No suitable text track found");
```
cpc005, line 7659: Do we need this WARN log msg and also the one given by the code that calls this func if no track is found?

fragmentcollector_hls.cpp
DomSyna: Done

fragmentcollector_hls.cpp
DomSyna: Strictly speaking no, but I want to leave it as CC on a whole needs a proper refactor and at that point it will probably prove useful

cpc005 (approved):

fragmentcollector_hls.cpp
```diff
@@ -580,13 +580,19 @@ AAMPStatusType StreamAbstractionAAMP_HLS::ParseMainManifest()
...
-						mediaInfoStore[mMediaCount].language =  mediaInfoStore[mMediaCount].name;
-					}
-					if (mediaInfoStore[mMediaCount].type == eMEDIATYPE_AUDIO && !mediaInfoStore[mMediaCount].language.empty() )
+					if (mediaInfoStore[mMediaCount].type == eMEDIATYPE_AUDIO)
 					{
-						mLangList.insert(GetLanguageCode(mMediaCount));
+						auto& audioMedia = mediaInfoStore[mMediaCount];
+						if (audioMedia.language.empty())
+						{ // handle non-compliant manifest missing language attribute
+							AAMPLOG_WARN("Audio track %d missing language, using name '%s' as fallback",
```
pstroffolino, line 588: do we actually need/want this?
if no language, why not just leave empty instead of falling back to name?

Isn't this one of the very things we were hoping to correct with this review?

dnfaulkner: We tried to minimise the change in behaviour to just fix the closed captions issue, just in case there was still a possible scenario needed for audio but will follow your guidance if you think that we should correct it for audio as well?

DomSyna: If we want to investigate / change this it should probably be done as a separate ticket as there is a good chance it may impact higher layers as a value they could have been checking may now become an empty string, or cease to exist.

pstroffolino (approved):

PR #708, opened 2025-11-26
VPLAY-11814 Content type index discrepancy in IP_AAMP_TUNETIME

sayanurag:
Reason for change: Moving gstdecode parameter to end for the metrics dashboard.
Risks: Low


pstroffolino (changes requested):
I don't want to merge this.  The impacted tool should ideally be fixed taking IP_AAMP_TUNETIME version field like every other tool.

pstroffolino:
won't be merged - ticket closed working as designed

PR #710, opened 2025-11-27
VPLA-11918 aamp l2 test framework support to collect high watermark for concurrent threads

cpc005:
Reason for change: L2 tests to check number of threads created by AAMP
Output PID when aamp-cli starts so L2 framework can read this information
Risk: Low

### DIFF
--- a/test/aampcli/Aampcli.cpp
+++ b/test/aampcli/Aampcli.cpp
@@ -21,7 +21,7 @@
  * @file Aampcli.cpp
  * @brief Stand alone AAMP player with command line interface.
  */
-
+#include <unistd.h>
 #include "Aampcli.h"
 #include "scte35/AampSCTE35.h"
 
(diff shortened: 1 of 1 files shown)


test/aampcli/Aampcli.cpp
```diff
@@ -371,7 +371,7 @@ static int main_func(int argc, char **argv)
 	AAMPCLI_PRINTF("**************************************************************************\n");
 	AAMPCLI_PRINTF("** ADVANCED ADAPTIVE MEDIA PLAYER (AAMP) - COMMAND LINE INTERFACE (CLI) **\n");
 	AAMPCLI_PRINTF("**************************************************************************\n");
-
+	AAMPCLI_PRINTF("process: %d\n", getpid()); //Used by L2 tests
```
jfagunde, line 374: I would add [AAMPCLI], to be consistent with other messages printed and to make it clearer that this is the process in which aamp-cli is running. I.e.
```
AAMPCLI_PRINTF("[AAMPCLI] process: %d\n", getpid());
```

cpc005: It is not universal, no action

jfagunde (approved):

pstroffolino (approved):

PR #675, opened 2025-11-14
VPLAY-11362:Playback starts at the actual position the user requested

srikanthreddybijjam-comcast:
Reason for change: Adjusted Detla to match PTO and StartTime
Test Procedure: Refer jira ticket VPLAY-11362
Priority: P1

PR #716, opened 2025-11-29
Revert "VPLAY-11362:Playback starts at the actual position the user requested"

pstroffolino:
Reverts rdkcentral/aamp#675

l2 tests 9001, 9002 failing


pstroffolino:
Temporary revert only, used to confirm this was reason for regression.

PR #721, opened 2025-12-01
VPLAY-11951 AampTime fixed point calculation conversion gives wrong result with some input

pstroffolino:
VPLAY-11951 AampTime fixed point calculation conversion gives wrong result with some input

Reason for Change: introduced overflow sanity check and fallback to floating point arithmetic

Test Guidance: see new L1 test

Risk: Medium


pstroffolino:
closed in favor of https://github.com/rdkcentral/aamp/pull/786

PR #723, opened 2025-12-02
Feature/vplay 11795 memset

DomSyna:
VPLAY-11795: Replace unsafe memset with RAII patterns

Fixes undefined behaviour from memset on non-POD types and modernizes POD initialization.

Key changes:

Non-POD types: Use proper cleanup methods instead of memset
POD types: Use = {} initialization at declaration
Code cleanup: Remove dead code, simplify logic

Impact: 16 memset calls replaced across 6 files. No functional changes.

Risk: low

### DIFF
--- a/fragmentcollector_mpd.cpp
+++ b/fragmentcollector_mpd.cpp
@@ -160,6 +160,7 @@ StreamAbstractionAAMP_MPD::StreamAbstractionAAMP_MPD(class PrivateInstanceAAMP *
 	,mIsSegmentTimelineEnabled(false)
 	,mSeekedInPeriod(false)
 	,mIsFinalFirstPTS(false)
+	,mMediaStreamContext{}
 {
 	this->aamp = aamp;
 	if (aamp->mDRMLicenseManager)
@@ -167,7 +168,6 @@ StreamAbstractionAAMP_MPD::StreamAbstractionAAMP_MPD(class PrivateInstanceAAMP *
 		AampDRMLicenseManager *licenseManager = aamp->mDRMLicenseManager;
 		licenseManager->SetLicenseFetcher(this);
 	}
-	memset(&mMediaStreamContext, 0, sizeof(mMediaStreamContext));
 	GetABRManager().clearProfiles();
 	mLastPlaylistDownloadTimeMs = aamp_GetCurrentTimeMS();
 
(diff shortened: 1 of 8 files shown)


fragmentcollector_mpd.cpp
```diff
@@ -10982,7 +10985,6 @@ static void indexThumbnails(dash::mpd::IMPD *mpd, int thumbIndexValue, std::vect
 												{
 													std::string tmedia = media;
 													TileInfo tileInfo;
```
jfagunde, line 10985: Initialise tileInfo to its default value? I.e.
`TileInfo tileInfo{};`
I think it is not needed, but you are doing it in other places, like `std::tm tmTime{};`, why not here and in line 11034?

tsprocessor.cpp
```diff
@@ -237,10 +237,6 @@ TSProcessor::TSProcessor(class PrivateInstanceAAMP *aamp,StreamOperation streamO
 		}
 		m_demux = true;
 	}
-
-	int compBufLen = MAX_PIDS*sizeof(RecordingComponent);
-	memset(videoComponents, 0, compBufLen);
```
jfagunde, line 242: videoComponents and audioComponents should also be included in the initialization list.

fragmentcollector_mpd.cpp
DomSyna: tileInfo calls the default constructor
tileInfo{} calls the default constructor (value initialization)

We could have the {} here but its redundant. We should use the {} for:
POD types without constructors for 0 initialisation
Arrays to 0 init all elements
Aggregate types to force init for all members

std::tm is the first type POD without constructor so its needed for 0 init.

tsprocessor.cpp
DomSyna: No. 
RecordingComponents have a constructor, and since videoComponents are an array of these, each element in the array is automatically initialised with this constructor.

DomSyna: I was tempted to remove the constructor from RecordingComponent and provide the member vars with initial vals, but can't fix everything in one review

jfagunde (approved):

p-bond (approved):

pstroffolino (approved):

PR #724, opened 2025-12-02
VPLAY-11886: Rate correction is enabled during RW on LLD

psiva01:
Reason for change: Modified the LLD LatencyMonitor condition to disable rate change during rewind as well.
Test Procedure: Refer Jira
Risks: Low


pstroffolino (approved):

fragmentcollector_mpd.cpp
```diff
@@ -3503,7 +3503,7 @@ AAMPStatusType StreamAbstractionAAMP_MPD::Init(TuneType tuneType)
 				liveAdjust = true;
 				notifyEnteringLive = true;
 			}
-			else if (((eTUNETYPE_SEEK == tuneType) || (eTUNETYPE_RETUNE == tuneType || eTUNETYPE_NEW_SEEK == tuneType)) && (currentRate > AAMP_RATE_PAUSE))
+			else if (((eTUNETYPE_SEEK == tuneType) || (eTUNETYPE_RETUNE == tuneType || eTUNETYPE_NEW_SEEK == tuneType)) && (currentRate != AAMP_RATE_PAUSE))
```
nu641001, line 3503: Previously, the logic used aamp->rate, but after the localTSB integration, it shifted to currentRate which sets PAUSE to stream abstraction object if pipelinePaused, but in actual PAUSE, aamp->rate remains AAMP_NORMAL_RATE and the pipelinePaused flag is set. Is it acceptable to keep using aamp->rate > AAMP_RATE_PAUSE? With the recent fix, the scope of this else if now also includes rewind scenarios—correct?

jfagunde: Hi @nu641001 ,
currentRate is a local variable set to aamp->rate in line 3198
Before merging https://github.com/rdkcentral/aamp/pull/649/files#diff-9412049d3685e68289ccd0b5e0a718c64d26b9a783882ba14700abd57e136be3, this function was using mPlayRate (which is always 1 when using AAMP TSB, see [VPLAY-11583](https://ccp.sys.comcast.net/browse/VPLAY-11583)). Is that what you mean?

nu641001: Yes @jfagunde Now I see where the confusion came from. The rate in StreamAbstraction essentially mirrors the value in aamp->rate. Since aamp->rate never transitions to a PAUSED state, the check for rate != PAUSE doesn’t really apply here. We can safely remove that condition to avoid any future ambiguity.

pstroffolino (approved):

PR #735, opened 2025-12-04
VPLAY-11719 Honor subtitles/CC "sub-type" preference

kd-syna:
Reason for Change: When an app specifies the CC/subtitles "sub-type" preference, use that to choose the most appropriate text track.

Summary of Changes:

SavePreferredTextLanguage() will cache the value of the "sub-type" preference from the JSON preferences document.
SelectPreferredTextTrack() will then prefer an in-band CC track when "sub-type" is "CLOSED-CAPTIONS" and prefer an out-of-band subtitle track when "sub-type" is "SUBTITLES". If the "sub-type" is neither of those values, it is treated as if it was not specified. 

Risk: Low


fragmentcollector_hls.cpp
```diff
@@ -7642,15 +7647,25 @@ bool StreamAbstractionAAMP_HLS::SelectPreferredTextTrack(TextTrackInfo &selected
...
 
+		// Only if it IS specified AND matches, score sub-type match.
+		// If a preference is not set (or is not recognised) it doesn't match.
+		if ( (!track.isCC && aamp->preferredTextSubTypeString == "SUBTITLES")
+		  || ( track.isCC && aamp->preferredTextSubTypeString == "CLOSED-CAPTIONS"))
+		{
+			// This is intentional re-use of the AAMP_TYPE_SCORE weighting, as
+			// specifying name AND sub-type in the same preference profile is not supported.
+			score += AAMP_TYPE_SCORE;
+		}
```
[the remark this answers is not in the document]
DomSyna: Done

fragmentcollector_hls.cpp
```diff
@@ -7642,15 +7647,26 @@ bool StreamAbstractionAAMP_HLS::SelectPreferredTextTrack(TextTrackInfo &selected
...
+
 		// Update best if this score is higher
 		if (score > bestScore)
 		{
 			bestScore = score;
 			selectedTextTrack = track;
 
-			AAMPLOG_INFO("New best text track: lang=%s, rendition=%s, name=%s, score=%llu",
+			AAMPLOG_INFO("New best text track: lang=%s, rendition=%s, name=%s, CC=%s, score=%llu",
```
dnfaulkner, line 7497: I guess that this has been copied from the DASH implementation but wondering if it would be clearer to log the sub-type instead, like this as it aligns with lines 7652/3:
AAMPLOG_INFO("New best text track: lang=%s, rendition=%s, name=%s, sub-type=%s, score=%llu",
						 track.language.c_str(), track.rendition.c_str(),
						 track.name.c_str(), track.isCC ? "CLOSED-CAPTIONS" : "SUBTITLES", score);

priv_aamp.h
```diff
@@ -1064,7 +1064,8 @@ class PrivateInstanceAAMP : public DrmCallbacks, public std::enable_shared_from_
 	std::vector<std::string> preferredTextLanguagesList;	/**< list of preferred text languages from most-preferred to the least*/
 	std::string preferredTextRenditionString; 		/**< String value for rendition */
 	std::string preferredTextTypeString; 			/**< String value for text type */
-	std::string preferredTextLabelString; 			/**< String value for text type */
+	std::string preferredTextLabelString; 			/**< String value for label */
+	std::string preferredTextSubTypeString; 		/**< String value for sub-type  (e.g., "SUBTITLES" or "CLOSED-CAPTIONS") */
```
dnfaulkner: "i.e." instead of "e.g."? sub-type is only either one of those two strings

dnfaulkner (commented):
Only minor comments

fragmentcollector_hls.cpp
DomSyna: That may lead to confusion when people are inspecting logs as there is an isCC flag in the track but no sub-type.

priv_aamp.h
DomSyna: That was CoPilot, I'll change it

fragmentcollector_hls.cpp
dnfaulkner: Hmmm. Put both in then :) I like logging the sub-type because that's what's being scored against in the track preferences logged on line 7605.

dnfaulkner (approved):

fragmentcollector_hls.cpp
```diff
@@ -7650,15 +7655,26 @@ bool StreamAbstractionAAMP_HLS::SelectPreferredTextTrack(TextTrackInfo &selected
...
 		if (score > bestScore)
 		{
 			bestScore = score;
 			selectedTextTrack = track;
 
-			AAMPLOG_INFO("New best text track: lang=%s, rendition=%s, name=%s, score=%llu",
+			AAMPLOG_INFO("New best text track: lang=%s, rendition=%s, name=%s, CC=%s, sub-type=%s, score=%llu",
 						 track.language.c_str(), track.rendition.c_str(),
-						 track.name.c_str(), score);
+						 track.name.c_str(), track.isCC ? "True" : "False", track.isCC ? "CLOSED-CAPTIONS" : "SUBTITLES", score);
```
pstroffolino: Here we're logging two parameters both based on track.isCC
Wrong or just redundant?

fragmentcollector_hls.cpp
```diff
@@ -7650,15 +7655,26 @@ bool StreamAbstractionAAMP_HLS::SelectPreferredTextTrack(TextTrackInfo &selected
 			score += AAMP_TYPE_SCORE;
 		}
 
+		// Only if it IS specified AND matches, score sub-type match.
+		// If a preference is not set (or is not recognised) it doesn't match.
+		if ( (!track.isCC && aamp->preferredTextSubTypeString == "SUBTITLES")
+		  || ( track.isCC && aamp->preferredTextSubTypeString == "CLOSED-CAPTIONS"))
+		{
+			// This is intentional re-use of the AAMP_TYPE_SCORE weighting.
```
pstroffolino: This is fine, but if these were distinct, which should have priority?
We may eventually deprecate  the use of weights and replace with a comparator function.

fragmentcollector_hls.cpp
DomSyna: @dnfaulkner asked for the sub-type to be added as well, see outdated comment above. Happy to remove again

fragmentcollector_hls.cpp
```diff
@@ -7610,6 +7610,11 @@ bool StreamAbstractionAAMP_HLS::SelectPreferredTextTrack(TextTrackInfo &selected
 	unsigned long long bestScore = 0;
 	const auto& languageVectorToCheck = (aamp->preferredTextLanguagesList.empty()) ? aamp->preferredSubtitleLanguageVctr : aamp->preferredTextLanguagesList;
 
+	AAMPLOG_INFO("Preferences: rendition='%s' name='%s' sub-type='%s'",
```
pstroffolino: Could remove the "Preferences:" prefix given that function name SelectPreferredTextTrack is automatically logged

pstroffolino (approved):

fragmentcollector_hls.cpp
DomSyna: done

fragmentcollector_hls.cpp
DomSyna: done: removed sub type

fragmentcollector_hls.cpp
DomSyna: Done, changed name priority to be higher based on architect advise

pstroffolino (approved):

PR #768, opened 2025-12-16
VPLAY-12164 Http 404 is reported for previous channel mpd file while …

sayanurag:
…channel change

Reason for change: Moving the Teardown operation before the mpd update release cause some delay and results in more Http 404 errors.
Risks: Low


sayanurag:
Not needed

PR #845, opened 2026-01-13
VPLAY-12164 Http 404 for mpd while channel change

sayanurag:
Reason for change: Moving the Teardown operation before the mpd update release cause some delay and results in more Http 404 errors.
Risks: Low

### DIFF
--- a/priv_aamp.cpp
+++ b/priv_aamp.cpp
@@ -8359,6 +8359,10 @@ void PrivateInstanceAAMP::Stop( bool sendStateChangeEvent )
 	//Moved the tsb delete request from XRE to AAMP to avoid the HTTP-404 erros
 	if(IsFogTSBSupported())
 	{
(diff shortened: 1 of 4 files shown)


priv_aamp.cpp
```diff
@@ -5118,6 +5118,15 @@ void PrivateInstanceAAMP::TeardownStream(bool newTune, bool disableDownloads)
 		}
 		ReleaseStreamLock();
 	}
+
+	if(disableDownloads)
+	{
+		// stop the mpd update immediately after Stream abstraction delete
+		if(mMPDDownloaderInstance != nullptr)
+		{
+			mMPDDownloaderInstance->Release();
```
jfagunde: Can this call be removed from other places since it is done here? I.e. from PrivateInstanceAAMP::detach() and PrivateInstanceAAMP::Stop()

PR #900, opened 2026-01-23
VPLAY-11582 ProcessFragmentChunk() uses timescale of segment downloaded

jfagunde:
Reason for Change: Avoid potential edge case and make code correct

Summary of Changes:
* Use timescale of fragment injected in ProcessFragmentChunk()
* Verify timescale used by InjectFragment() in L1 tests

Test Procedure: L1 tests should pass and no AAMP TSB + CDA regressions

Risk: Low


streamabstraction.cpp
```diff
@@ -1031,18 +1031,33 @@ bool MediaTrack::ProcessFragmentChunk()
...
-		timeScale = aamp->GetAudTimeScale();
-	}
-	else if (type == eTRACK_SUBTITLE)
+
+	// Use the timescale stored in the cached fragment, which represents the timescale
+	// of the segment being injected. This is critical when using TSB, as the segment
+	// being downloaded at the live edge may have a different timescale (e.g., an ad)
+	// than the segment being injected from TSB (e.g., base content).
+	uint32_t timeScale = cachedFragment->timeScale;
+	//uint32_t timeScale = 0;		// Jose
```
DomSyna: delete ?

streamabstraction.cpp
```diff
@@ -1031,18 +1031,33 @@ bool MediaTrack::ProcessFragmentChunk()
...
+	if(!timeScale)
 	{
-		timeScale = aamp->GetSubTimeScale();
+		AAMPLOG_WARN("[%s] Cached fragment timescale is 0, fragment URI: %s", name, cachedFragment->uri.c_str());
```
DomSyna: How can the timescale in the cached fragment be 0 ?

DomSyna (commented):
Not looked at tests yet, just thought better to iron out code changes as they may impact tests

streamabstraction.cpp
jfagunde: Thanks, I forgot to remove that.

streamabstraction.cpp
jfagunde: I asked that same question to Copilot. Here is the answer:
Yes, [cachedFragment->timeScale](vscode-file://vscode-app/snap/code/219/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) being 0 is a valid real-world scenario. Here's why:

MPD without timescale: If the DASH manifest (MPD) doesn't specify a timescale attribute in either the representation-level or adaptation-set-level SegmentTemplate, GetTimescale() returns 0.

libdash default behavior: The comment says "if timescale missing in template, GetTimeScale returns 1" - but if both templates are null or neither has a timescale, it returns 0.

Non-TSB paths: Looking at [MediaStreamContext.cpp:74-80](vscode-file://vscode-app/snap/code/219/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), if [mActiveDownloadInfo](vscode-file://vscode-app/snap/code/219/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is NULL (which triggers a warning at line 84), the timescale comes from [fragmentDescriptor.TimeScale](vscode-file://vscode-app/snap/code/219/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), which itself could be 0 if not initialized properly or during certain edge cases.

Legacy content: Some older or malformed DASH streams may not have proper timescale values.

HLS streams: For non-DASH content (HLS), the [timeScale](vscode-file://vscode-app/snap/code/219/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) field might not be populated at all since HLS doesn't have an explicit timescale concept in the same way.

The fallback logic in [ProcessFragmentChunk()](vscode-file://vscode-app/snap/code/219/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is defensive programming - it ensures the code works even in edge cases where the timescale isn't properly propagated through the system. The fallback to global timescales (and then to a default value) prevents division-by-zero errors and ensures playback continues even with missing metadata.

However, you could consider removing the test cases for [timeScale == 0](vscode-file://vscode-app/snap/code/219/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) if you want the test to strictly verify only the TSB fix behavior. The important test cases are the ones where [cachedFragment->timeScale != globalTimeScale](vscode-file://vscode-app/snap/code/219/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - those would definitively fail with the old code.

jfagunde: After discussing with @DomSyna , we don't believe this is a valid case, so I have deleted the code that tries to recover from timescale = 0.

streamabstraction.cpp
```diff
@@ -1031,36 +1031,16 @@ bool MediaTrack::ProcessFragmentChunk()
...
-				AAMPLOG_WARN("[%s] Empty timeScale!!! Using default timeScale=%d", name, timeScale);
-			}
-		}
-		else
-		{
-			timeScale = 1000.0;
-			AAMPLOG_WARN("[%s] Invalid play context maybe test setup, timeScale=%d", name, timeScale);
-		}
+		AAMPLOG_ERR("[%s] Cached fragment timescale is 0, fragment URI: %s", name, cachedFragment->uri.c_str());
+		return true;
```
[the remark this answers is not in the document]
jfagunde: That is not true. ProcessFragmentChunk() returns true when the fragment should be removed from the cached fragment chunk buffer and false otherwise. I am adding a comment to the declaration of the function to clarify that.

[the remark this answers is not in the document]
jfagunde: Removing the fallback options is intentional. If the timescale of a fragment is 0, it should not be injected. That can do more harm than good. With this change, it will print an ERROR and discard the fragment.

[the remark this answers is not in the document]
jfagunde: I have updated the tests.

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
```diff
@@ -820,3 +820,113 @@ TEST_F(MediaTrackTests, MediaTrackConstructorChunkModeTest)
...
+
+	// Configure low-latency mode since ProcessFragmentChunk is used for chunk-based injection
+	SetLowLatencyMode(true);
+	mPrivateInstanceAAMP->rate = AAMP_NORMAL_PLAY_RATE;
+	mPrivateInstanceAAMP->mMediaFormat = eMEDIAFORMAT_DASH;
+	mStreamAbstractionAAMP_MPD->trickplayMode = false;
+
+	// Set up mock to return the global timescale (used as fallback when cached fragment timescale is 0)
+	EXPECT_CALL(*g_mockPrivateInstanceAAMP, GetVidTimeScale())
+		.WillRepeatedly(Return(testData.globalTimeScale));
```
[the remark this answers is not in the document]
jfagunde: Test modified.

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
```diff
@@ -820,3 +820,113 @@ TEST_F(MediaTrackTests, MediaTrackConstructorChunkModeTest)
...
+}
+
+ChunkTimescaleTestData chunkTimescaleTestData[] = {
+	// When cached fragment has a valid timescale, it should be used
+	{90000, 48000, 90000},    // Fragment timescale 90000 (typical video), global 48000 -> use 90000
+	{48000, 90000, 48000},    // Fragment timescale 48000 (typical audio), global 90000 -> use 48000
+	{10000000, 90000, 10000000},  // Fragment timescale 10000000 (100ns units), global 90000 -> use 10000000
+	// When cached fragment timescale is 0, fall back to global timescale
+	{0, 90000, 90000},        // Fragment timescale 0, global 90000 -> use global
+	{0, 48000, 48000},        // Fragment timescale 0, global 48000 -> use global
```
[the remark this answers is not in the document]
jfagunde: Tests updated.

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
```diff
@@ -820,3 +820,113 @@ TEST_F(MediaTrackTests, MediaTrackConstructorChunkModeTest)
...
+ *
+ * This test verifies that when processing fragment chunks, the timescale from
+ * the cached fragment is used (if set), rather than falling back to global
+ * timescales. This is critical for TSB (Time-Shift Buffer) scenarios where:
+ * - The segment being downloaded at the live edge may be from an ad with one timescale
+ * - The segment being injected from TSB may be from base content with a different timescale
+ *
+ * The test would fail if ProcessFragmentChunk used global timescales instead of the
+ * cached fragment's timescale, as the ParseChunkData mock expects the exact timescale
+ * value from the cached fragment.
```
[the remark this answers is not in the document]
jfagunde: Test updated

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
```diff
@@ -820,3 +820,113 @@ TEST_F(MediaTrackTests, MediaTrackConstructorChunkModeTest)
...
+ */
+TEST_P(MediaTrackProcessFragmentChunkTimescaleTests, UseCachedFragmentTimescaleInParseChunkData)
```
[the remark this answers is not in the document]
jfagunde: Test updated

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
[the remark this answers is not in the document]
jfagunde: Test updated

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
```diff
@@ -820,3 +820,113 @@ TEST_F(MediaTrackTests, MediaTrackConstructorChunkModeTest)
...
+	testFragment.uri = "test_segment.m4s";
+	testFragment.fragment.AppendBytes(FRAGMENT_TEST_DATA, strlen(FRAGMENT_TEST_DATA));
+
+	bufferedFragment = videoTrack.GetFetchChunkBuffer(true);
+	videoTrack.numberOfFragmentChunksCached = 1;
+	bufferedFragment->Copy(&testFragment, testFragment.fragment.GetLen());
+
+	// Key assertion: ParseChunkData should be called with the expected timescale
+	// This tests that ProcessFragmentChunk uses cachedFragment->timeScale when set,
+	// or falls back to global timescale when cachedFragment->timeScale is 0
```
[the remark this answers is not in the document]
jfagunde: Test updated

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
```diff
@@ -820,3 +824,133 @@ TEST_F(MediaTrackTests, MediaTrackConstructorChunkModeTest)
...
+	constexpr uint32_t kFragmentTimeScale{90000};
+	CachedFragment* bufferedFragment{nullptr};
+
+	// Configure low-latency mode since ProcessFragmentChunk is used for chunk-based injection
+	SetLowLatencyMode(true);
+	mPrivateInstanceAAMP->rate = AAMP_NORMAL_PLAY_RATE;
+	mPrivateInstanceAAMP->mMediaFormat = eMEDIAFORMAT_DASH;
+	mStreamAbstractionAAMP_MPD->trickplayMode = false;
+
+	EXPECT_CALL(*g_mockAampConfig, IsConfigSet(eAAMPConfig_OverrideMediaHeaderDuration))
```
DomSyna: These could use _ as they all return false or 1. Maybe even move them out into a helper fn as they are generic setup and not really part of the test

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
```diff
@@ -820,3 +824,133 @@ TEST_F(MediaTrackTests, MediaTrackConstructorChunkModeTest)
...
+	EXPECT_CALL(*g_mockAampConfig, GetConfigValue(eAAMPConfig_MaxFragmentChunkCached))
+		.WillRepeatedly(Return(1));
+	EXPECT_CALL(*g_mockIsoBmffBuffer, parseBuffer(_, _)).WillRepeatedly(Return(true));
+	EXPECT_CALL(*g_mockPrivateInstanceAAMP, GetLLDashChunkMode()).WillRepeatedly(Return(true));
+
+	TestableMediaTrack videoTrack{eTRACK_VIDEO, mPrivateInstanceAAMP, "video",
+								  mStreamAbstractionAAMP_MPD};
+
+	// First inject an init fragment (required before media fragments)
+	CachedFragment initFragment;
```
DomSyna: {}

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
```diff
@@ -820,3 +824,133 @@ TEST_F(MediaTrackTests, MediaTrackConstructorChunkModeTest)
...
+	initFragment.initFragment = true;
+	initFragment.fragment.AppendBytes(FRAGMENT_TEST_DATA, strlen(FRAGMENT_TEST_DATA));
```
DomSyna: better to have a const string at top and use that
const std::string testData = FRAGMENT_TEST_DATA;
then here 
initFragment.fragment.AppendBytes(testData.data(), testData.length());

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
```diff
@@ -820,3 +824,133 @@ TEST_F(MediaTrackTests, MediaTrackConstructorChunkModeTest)
...
+	bufferedFragment = videoTrack.GetFetchChunkBuffer(true);
+	videoTrack.numberOfFragmentChunksCached = 1;
+	bufferedFragment->Copy(&initFragment, initFragment.fragment.GetLen());
+	ASSERT_TRUE(videoTrack.InjectFragment());
```
DomSyna: To avoid duplication create a helper, though a lambda is probably more cpp 17 here and at bottom of test
ASSERT_TRUE(InjectChunk(initFragment));

where above you define the inject chunk lambda, something like
auto InjectChunk = [&](const CachedFragment& fragmentSource) {
        auto* dest = videoTrack.GetFetchChunkBuffer(true);
        ASSERT_NE(dest, nullptr); // Defensive check
        dest->Copy(&fragmentSource, fragmentSource.fragment.GetLen());
        videoTrack.numberOfFragmentChunksCached = 1;
        return videoTrack.InjectFragment();
    };
think thats about right

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
```diff
@@ -820,3 +824,133 @@ TEST_F(MediaTrackTests, MediaTrackConstructorChunkModeTest)
...
+	constexpr uint32_t kZeroTimeScale{0};
+	CachedFragment* bufferedFragment{nullptr};
+
+	// Configure low-latency mode since ProcessFragmentChunk is used for chunk-based injection
+	SetLowLatencyMode(true);
+	mPrivateInstanceAAMP->rate = AAMP_NORMAL_PLAY_RATE;
+	mPrivateInstanceAAMP->mMediaFormat = eMEDIAFORMAT_DASH;
+	mStreamAbstractionAAMP_MPD->trickplayMode = false;
+
+	EXPECT_CALL(*g_mockAampConfig, IsConfigSet(eAAMPConfig_OverrideMediaHeaderDuration))
```
DomSyna: As above... looks like the same setup, so a helper setup fn would be good, probably add it to the fixture as both tests use it, maybe others can to, not looked

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
```diff
@@ -820,3 +824,133 @@ TEST_F(MediaTrackTests, MediaTrackConstructorChunkModeTest)
...
+ *
+ * This test verifies the behaviour when a cached fragment has a zero timescale.
+ * This should never happen in real playback scenarios. When it does occur,
+ * ProcessFragmentChunk returns early without calling ParseChunkData, as the
+ * timescale is required for correct PTS calculation.
+ */
+TEST_F(MediaTrackTests, ProcessFragmentChunkWithZeroTimescale)
+{
+	constexpr uint32_t kZeroTimeScale{0};
+	CachedFragment* bufferedFragment{nullptr};
```
DomSyna: const string FRAGMENT_TEST_DATA

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
```diff
@@ -820,3 +824,133 @@ TEST_F(MediaTrackTests, MediaTrackConstructorChunkModeTest)
...
+								  mStreamAbstractionAAMP_MPD};
+
+	// First inject an init fragment (required before media fragments)
+	CachedFragment initFragment;
+	initFragment.initFragment = true;
+	initFragment.fragment.AppendBytes(FRAGMENT_TEST_DATA, strlen(FRAGMENT_TEST_DATA));
+	bufferedFragment = videoTrack.GetFetchChunkBuffer(true);
+	videoTrack.numberOfFragmentChunksCached = 1;
+	bufferedFragment->Copy(&initFragment, initFragment.fragment.GetLen());
+	ASSERT_TRUE(videoTrack.InjectFragment());
```
DomSyna: use lambda as suggested above, although given its here as well in different test maybe add a method to the test fixture

streamabstraction.cpp
DomSyna, line 1051: Took a while to see why it was returning true, from github code review its not at all obvious

jfagunde: I agree. The return value of this function is not at all obvious. I added the function comment to the header file. I will add another comment here to make it clearer.

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
jfagunde: Done

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
jfagunde: FRAGMENT_TEST_DATA is already a const char used in multiple tests in this file. I rather leave it as it is for consistency with the rest of the tests.

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
jfagunde: This is the pattern used in existing tests. Most of the values are always the same, but some might change in different tests, like eAAMPConfig_EnablePTSReStamp. I rather not do the suggested refactoring as part of this Jira.

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
jfagunde: Added helper method to test fixture, as suggested, to avoid code duplication.

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
jfagunde: Added helper method to test fixture, as suggested, to avoid code duplication.

test/utests/tests/MediaTrackTests/MediaTrackTests.cpp
jfagunde: Setup is very similar for the different tests, not exactly the same. And I prefer to follow the existing pattern and not refactor existing tests as part of this Jira.

DomSyna (approved):
Tests could be restructured better to honour DRY and SOLID principles

test/utests/tests/TrackInjectTests/TrackInjectTests.cpp
```diff
@@ -126,6 +126,7 @@ class MediaTrackTest : public MediaTrack
 			this->mCachedFragment = new CachedFragment[3];
 			cachFragment = &this->mCachedFragment[fragmentIdxToFetch];
 		}
+		cachFragment->timeScale = 1;
 		cachFragment->initFragment = isInit;
```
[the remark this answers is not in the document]
jfagunde: Implemented the suggestion, as done in MediaTrackTests.

DomSyna (approved):

Vinish100 (commented):
Looks good to me 👍

easy8software (approved):

PR #904, opened 2026-01-23
VPLAY-12061:Closed captioning issues after introducing webvTT along …

Gnanesha:
VPLAY-12061:Closed captioning issues after introducing webvTT along with inBand CC tracks

Reason for change: filter out the webvtt tracks before sending the list of tracks to application
Test Procedure: see ticket
Risks: Low


Gnanesha:
will push new version

PR #916, opened 2026-01-27
VPLAY-12294:underflow detection - avoid dependency on SoC signaling

varshnie:
Reason for change:Aamp detect and control the underflow Test Procedure:Refer Jira Ticket
Risks: Medium


AampUnderflowMonitor.h
```diff
@@ -0,0 +1,110 @@
...
+     * @brief Destructor. Ensures monitoring has been stopped.
+     */
+    ~UnderflowMonitor();
+
+    /**
+     * @fn start
+     * @brief Start the monitoring thread; no-op if already running.
+     * @return void
+     */
+    void start();
```
Vinish100: Use pascal case for function names

priv_aamp.cpp
```diff
@@ -8923,6 +8950,27 @@ void PrivateInstanceAAMP::SetState(AAMPPlayerState state)
 	}
 
 	mScheduler->SetState(mState);
+
+	// Start or stop UnderflowMonitor in response to player state changes
+	if (mpStreamAbstractionAAMP)
```
nu641001: Is it a good idea to access streamabstraction object without a stream lock? ie AcquireStreamLock and ReleaseStreamLock

AampDefine.h
```diff
@@ -148,6 +148,15 @@
 
 #define DEFAULT_EARLY_ABORT_PROFILE_BANDWIDTH_PERCENT 80 /**< By default, let's abort early if bps is less than 80% of profile bandwidth */
 
+#define DEFAULT_UNDERFLOW_LOW_BUFFER_POLL_MS 500		/**< Polling interval when buffer is low in milliseconds */
```
pstroffolino, line 151: this seems needlessly complicated
we can advance and follow-up, but the idea of polling to detect underflow seems inefficient and error prone

If we know we have buffered filled with N seconds of content, and we know the current time, we should be able to predict precisely when underflow will occur.

that said, polling will be more robust if play speed is inaccurate, or there is jumping across gaps

PR #917, opened 2026-01-27
VPLAY-12490 Use CPP 17 to build AAMP

DomSyna:
Move the CPP version forward to 17

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ function(xcode_define_schema new_schema)
 endfunction()
 
 project (AAMP)
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror=format -Wno-multichar -Wno-non-virtual-dtor -Wno-psabi")
--- a/middleware/gst-plugins/CMakeLists.txt
+++ b/middleware/gst-plugins/CMakeLists.txt
@@ -53,7 +53,7 @@ include_directories(${MW_ROOT}/)
 
 set(GST_COMMON_DEPENDENCIES ${OS_LD_FLAGS} ${GSTREAMERBASE_LIBRARIES} ${GSTREAMER_LIBRARIES} ${CURL_LIBRARIES} ${CLI_LD_FLAGS} -ldl -luuid ${SEC_CLIENT_LIB})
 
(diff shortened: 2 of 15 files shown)


CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ function(xcode_define_schema new_schema)
 endfunction()
 
 project (AAMP)
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
cpc005, line 30: Seems that we are still setting the C++ version in a lot of places. 
Do the tests need to specify the version or should they just pick it up from the version set for the code being tested.
Perhaps some common include file where the version is set.
Do the full stack builds still work.

DomSyna: I left it in the submodules that look like they can be built independently from makefile inspection. Didn't want to break them by being overly zealous

middleware/test/utests/CMakeLists.txt
```diff
@@ -43,8 +43,6 @@ if (NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
   find_path (JSC_INCDIR JavaScriptCore/JavaScript.h HINTS ${JSCORE_INCLUDE_DIRS} REQUIRED)
 endif()
 
-set(CMAKE_CXX_STANDARD 17)
```
jfagunde, line 46: Why did you remove these lines? Are they not needed?

DomSyna: Good spot

jfagunde (approved):

middleware/gst-plugins/CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ include_directories(${MW_ROOT}/)
 
 set(GST_COMMON_DEPENDENCIES ${OS_LD_FLAGS} ${GSTREAMERBASE_LIBRARIES} ${GSTREAMER_LIBRARIES} ${CURL_LIBRARIES} ${CLI_LD_FLAGS} -ldl -luuid ${SEC_CLIENT_LIB})
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-multichar -std=c++11")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-multichar")
```
dnfaulkner, line 56: Why don't you need -std=c++17 on this one? I guess it doesn't need to be a specific version?

dnfaulkner (commented):
.

middleware/gst-plugins/CMakeLists.txt
DomSyna: should inherit from the parent makefile

easy8software (approved):

pstroffolino (approved):

PR #932, opened 2026-01-28
VPLAY-12431 : App Suspending when user try to Play video

Gnanesha:
VPLAY-12431 : App Suspending when user try to Play video
    
    Reason for change : Do not send state change event when destruction of object is in progress
    Risks: Low
    Test Procedure: Please see the ticket
    Priority: P0


paulpandian25:
Looks good.

PR #936, opened 2026-01-29
VPLAY-12482 use of TLS1.3 for CDN downloads

pstroffolino:
VPLAY-12482 use of TLS1.3 for CDN downloads

Reason for Change: bug fix for CurlDownloader config propogation
Changed default eAAMPConfig_TLSVersion  back to CURL_SSLVERSION_TLSv1_2.  This is a minimum TLS version; the best supported TLS version between client and server will still be negotiated and used, typically TLSv1.3

Test Guidance: with verbose curl logging, confirm both segment and manifest downloads using TLSv1.3


dnfaulkner (commented):
Looks good to me. As discussed on Slack, I agree that the L1 tests are of limited value - they're just testing the config setting/getting interface. I agree that it would be much more useful to confirm the configured setting gets used in the mocked media/manifest download requests, which then propagate into the curl requests.

PR #950, opened 2026-01-31
VPLAY-12334 update gstttestharness to use common newer, hardened mp4demux implementation integrated with aamp

pstroffolino:
VPLAY-12334 update gstttestharness to use common newer, hardened mp4demux implementation integrated with aamp

Reason for Change: consolidate two versions of mp4demux


jfagunde (changes requested):
AAMP fails to build with this PR on ubuntu.

PR #951, opened 2026-02-02
Feature/vplay 12344

Reshma-JO07:


middleware/InterfacePlayerRDK.cpp
```diff
@@ -1908,6 +1946,236 @@ static GstPadProbeReturn InterfacePlayerRDK_DemuxPadProbeCallbackAny(GstPad *pad
...
+		gst_bin_remove(GST_BIN(stream->sinkbin), decryptor);
+		gst_element_set_state(queue, GST_STATE_NULL);
+		gst_bin_remove(GST_BIN(stream->sinkbin), queue);
+		return;
+	}
+	
+	MW_LOG_WARN("Successfully linked queue to decryptor via manual pad linking");
+
+	// Get the pre-created rialto sink (created in SetupStream)
+	GstElement* realSink = nullptr;
```
Vinish100, line 2423: These can be a helper class to identify the sink

middleware/InterfacePlayerRDK.cpp
```diff
@@ -1908,6 +1946,236 @@ static GstPadProbeReturn InterfacePlayerRDK_DemuxPadProbeCallbackAny(GstPad *pad
...
+	MW_LOG_MIL("Linking elements in static pipeline for %s", gstGetMediaTypeName(mediaType));
+
+	// Manual pad linking to bypass custom element checks while respecting caps
+	// gst_element_link() fails on Xi One UK widevinedecryptor despite compatible template caps
+	GstPad* queueSrcPad = gst_element_get_static_pad(queue, "src");
+	GstPad* decryptorSinkPad = gst_element_get_static_pad(decryptor, "sink");
+	
+	if (!queueSrcPad || !decryptorSinkPad)
+	{
+		MW_LOG_ERR("Failed to get pads for manual linking");
```
Vinish100: See, if some of these codes could be brought into a generic function for linking 2 elements. We can avoid code duplication

middleware/InterfacePlayerRDK.cpp
```diff
@@ -1908,6 +1946,236 @@ static GstPadProbeReturn InterfacePlayerRDK_DemuxPadProbeCallbackAny(GstPad *pad
 	}
 	return rtn;
 }
+static void GstPlayer_OnDemuxPadAddedCb_Static(GstElement* demux, GstPad* newPad, void* _this)
```
Vinish100: As a optimization, would recommend moving this function implementation into a helper function in InterfacePlayerRDK, so that we can add some unit tests for same

middleware/InterfacePlayerRDK.cpp
Reshma-JO07: This statement is not clear. You mean need to create a helper class to identify the sink after sink gets created in SetupStream function?

Reshma-JO07: Done

middleware/InterfacePlayerRDK.cpp
Reshma-JO07: Done

middleware/InterfacePlayerRDK.cpp
Reshma-JO07: Done

PR #956, opened 2026-02-02
VPLAY-12061:Closed captioning issues after introducing webvTT along w…

Gnanesha:
…ith inBand CC tracks

Reason for change: filter out the webvtt tracks before sending the list of tracks to application
Test Procedure: see ticket
Risks: Low


priv_aamp.cpp
```diff
@@ -6450,6 +6450,8 @@ void PrivateInstanceAAMP::Tune(const char *mainManifestUrl,
 			SETCONFIGVALUE_PRIV(AAMP_DEFAULT_SETTING, eAAMPConfig_EnableLiveLatencyCorrection, true);
 		}
 		SETCONFIGVALUE_PRIV(AAMP_DEFAULT_SETTING, eAAMPConfig_EnablePTSReStamp, SocUtils::EnablePTSRestamp());
+		SETCONFIGVALUE_PRIV(AAMP_DEFAULT_SETTING, eAAMPConfig_DisableWebVTT, true);
```
Gnanesha, line 6453: “This suggestion is incorrect. The setting is applied in a controlled manner only where relevant for Peacock, and Peacock does not use WebVTT.

PR #959, opened 2026-02-03
Feature/vplay 12147 memory leak in aamp tsb data manager.cpp

DomSyna:
The prev and next pointers in TsbFragmentData class create a cyclic dependency on the shared pointers so they can't be automatically released as the usage count will never reach 0. To resolve this I have changed prev to a weak pointer.

### DIFF
--- a/AampTSBSessionManager.cpp
+++ b/AampTSBSessionManager.cpp
@@ -837,12 +837,12 @@ bool AampTSBSessionManager::NavigateToNextFragment(TsbFragmentDataPtr& fragment,
 	}
 	else							// Rewind
 	{
-		if (fragment->prev)
+		if (auto prevShared = fragment->prev.lock()) 
 		{
-			fragment = fragment->prev;
+			fragment = prevShared;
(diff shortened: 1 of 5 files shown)


AampTSBSessionManager.cpp
```diff
@@ -837,12 +837,12 @@ bool AampTSBSessionManager::NavigateToNextFragment(TsbFragmentDataPtr& fragment,
...
 	{
-		if (fragment->prev)
+		if (auto prevShared = fragment->prev.lock()) 
 		{
-			fragment = fragment->prev;
+			fragment = prevShared;
 			success = true;
-		}
-		if (!(fragment->prev))
+		}
```
jfagunde: Trailing tab

jfagunde (approved):

AampTSBSessionManager.cpp
DomSyna: done

pstroffolino (approved):

PR #960, opened 2026-02-03
VPLAY-12061 webvtt fix for peacock - follow-up

pstroffolino:
VPLAY-12061 webvtt fix for peacock - follow-up

Reason for Change: refined comment text and tab consistency


pstroffolino:
abandoning

PR #961, opened 2026-02-03
VPLAY-12576  Add precision to AAMP totalPerformRequest HttpRequestEnd

rajat698:
Reason for change: To avoid rounding of AAMP segment's downloadTimeMS
Risks: Low
Priority: P1


drm/AampDRMLicManager.cpp
```diff
@@ -566,7 +568,7 @@ KeyState AampDRMLicenseManager::processLicenseResponse(std::shared_ptr<DrmHelper
  * @param[in] eventHandle - DRM Metadata event handle
  * @param[in] respData - download response data
  */
-void AampDRMLicenseManager::UpdateLicenseMetrics(DrmRequestType requestType, int32_t statusCode, std::string licenseRequestUrl, long long downloadTimeMS, DrmMetaDataEventPtr eventHandle, DownloadResponsePtr respData )
+void AampDRMLicenseManager::UpdateLicenseMetrics(DrmRequestType requestType, int32_t statusCode, std::string licenseRequestUrl, double downloadTimeMS, DrmMetaDataEventPtr eventHandle, DownloadResponsePtr respData )
```
jfagunde: It would be better to use AampTime instead, to avoid floating point calculations.

drm/AampDRMLicManager.cpp
```diff
@@ -35,6 +35,8 @@
 #include "PlayerUtils.h"
 #include "PlayerSecInterface.h"
 
+#include <chrono>
+
```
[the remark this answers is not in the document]
easy8software: let's do this

drm/AampDRMLicManager.cpp
easy8software, line 570: why this change?  do we really need sub-ms downloatTimeMs reporting?  1ms is very tiny
in general, can make sense to use double for seconds, and integer types for ms.

drm/AampDRMLicManager.cpp
```diff
@@ -956,17 +958,18 @@ DrmData* AampDRMLicenseManager::getLicense(LicenseRequest &licenseRequest,
 	AAMPLOG_WARN(" Sending license request to server : %s ", licenseRequest.url.c_str());
 
 	unsigned int attemptCount = 0;
-	long long tStartTimeWithRetry = NOW_STEADY_TS_MS;
+	const auto startTimeWithRetry = std::chrono::steady_clock::now();
```
easy8software, line 960: why this change from NOW_STEADY_TS_MS macro to std::chrono::steady_clock::now()?

AampMPDDownloader.cpp
```diff
@@ -589,11 +593,11 @@ void AampMPDDownloader::stichToCachedManifest(ManifestDownloadResponsePtr mpdToA
 * @params DownloadResponse pointer ,totalPerformanceTime
 * @return void
 */
-void AampMPDDownloader::showDownloadMetrics(DownloadResponsePtr dnldPtr, int totalPerformanceTime)
+void AampMPDDownloader::showDownloadMetrics(DownloadResponsePtr dnldPtr, double totalPerformanceTimeMs)
 {
 	int http_code			=	dnldPtr->iHttpRetValue;
 	double total			=	dnldPtr->downloadCompleteMetrics.total;
-	double totalPerformRequest	= (double)(totalPerformanceTime)/1000;	// in sec
+	double totalPerformRequest	= totalPerformanceTimeMs / 1000.0;
```
pstroffolino, line 596: if we're going to make change here, probably good idea to add suffix to make units clear (i.e. Ms or S)

priv_aamp.cpp
```diff
@@ -4970,7 +4977,22 @@ bool PrivateInstanceAAMP::GetFile( std::string remoteUrl, AampMediaType mediaTyp
 					}
 
 					AAMPLOG(reqEndLogLevel, "HttpRequestEnd: %s%d,%d,%d%s,%2.4f,%2.4f,%2.4f,%2.4f,%2.4f,%2.4f,%2.4f,%2.4f,%g,%ld,%ld,%" BITSPERSECOND_FORMAT ",%.500s%s%s",
```
pstroffolino, line 5150: Here is where the functional change is happening, but not clear how?  understanding is there was some inconsistent rounding interfering with calculations.  But the floating point formatters above aren't changed.

drm/AampDRMLicManager.cpp
rajat698: We're doing so to match curl's precision. Before this PR (tip of the sprint), we were doing using ms and then dividing by 1000 for conversion to seconds for logging. So if actual time taken was for example 93.4 ms, or even 93.9 ms it rounded off to its floor integer (93) and then was divided by 1000 = 0.0930 seconds, instead of the actual 0.0934 or 0.0939. So this change is to bring this precision.

This precision helps us for more accurate measurements and to avoid mistakes like for example if we are calculating (total time - latency) and latency logged today is more precise. For example latency logged is 0.0936 and total time logged is 0.0930 (when it's actually 0.0939), total time - latency will be negative, which is wrong.

drm/AampDRMLicManager.cpp
rajat698: NOW_STEADY_TS_MS gives us integer milliseconds and not precise milliseconds with decimals which is needed for this change

priv_aamp.cpp
rajat698: Here it always was the floating data typed logged (time in ms/1000), but less precise. 0.0939 seconds was logged as 0.0930. We have just added that precision without changing the data type

AampMPDDownloader.cpp
```diff
@@ -589,11 +593,11 @@ void AampMPDDownloader::stichToCachedManifest(ManifestDownloadResponsePtr mpdToA
 * @params DownloadResponse pointer ,totalPerformanceTime
 * @return void
```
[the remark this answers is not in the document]
rajat698: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/rdkcentral/aamp/pull/961#discussion_r2760781945)

drm/AampDRMLicManager.cpp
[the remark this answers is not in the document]
rajat698: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/rdkcentral/aamp/pull/961#discussion_r2760781993)

PR #969, opened 2026-02-04
Feature/vplay 12577 vector like wrappers

DomSyna:
Added vector like wrappers into AampGrowableBuffer for the expected vector methods that will be needed, to aid with migration to vector from AampGrowableBuffer.

Replaced the the simple pass through calls like GetLen, GetAvail etc with the new equivalents

Fixed the allocation tracking accounting that was missed in the copy constructor, and aampCache as its useful to keep for now as one more way to validate further growable buffer changes.


test/utests/tests/AampGrowableBuffer/FunctionalTests.cpp
```diff
@@ -255,13 +255,13 @@ TEST_F(FunctionalTests, Reserve32KBytesTest)
 
 	// Act: Call the ReserveBytes function
 	buffer.ReserveBytes(numBytesToReserve);
-
+
```
jfagunde: Trailing space

AampGrowableBuffer.h
```diff
@@ -36,41 +36,46 @@
...
+			NETMEMORY_PLUS();
+		}
 	}
-	
+
 	// Copy assignment
-	AampGrowableBuffer& operator=(const AampGrowableBuffer & other)
-	{ // never reached/used?
-		buffer = other.buffer;
+	AampGrowableBuffer &operator=(AampGrowableBuffer other)
```
jfagunde, line 55: Is it intentional that the parameter other is no longer a reference?

AampGrowableBuffer.h
```diff
@@ -87,26 +92,78 @@ class AampGrowableBuffer
...
+	// Vector-like convenience wrappers (lower-case names to match std::vector)
+	bool empty() const { return buffer.empty(); }
+	void shrink_to_fit() { buffer.shrink_to_fit(); } // If used may need to call NETMEMORY_MINUS()
+	void reserve(size_t n) { buffer.reserve(n); }	 // If used may need to call NETMEMORY_PLUS()
+	void resize(size_t n)
+	{
+		const size_t prevCap = buffer.capacity();
+
+		buffer.resize(n);
+		if (prevCap == 0 && buffer.capacity() > 0)
```
jfagunde, line 117: If the previous capacity was not 0, should NETMEMORY_PLUS() be called? And same question for line 127

AampGrowableBuffer.h
DomSyna: Yes, by making it a value is will use the copy constructor above and set PLUS if appropriate. There is minimal overhead though as the swap basically does optimised ptr magic

AampGrowableBuffer.h
DomSyna: No only when we "allocate". The macro is very basic and tracks allocates vs Frees

AampGrowableBuffer.h
```diff
@@ -87,26 +92,78 @@ class AampGrowableBuffer
...
+	{
+		if (prevCap == 0 && newCap > 0)
+		{
+			NETMEMORY_PLUS();
+		}
+		else if (prevCap > 0 && newCap == 0)
+		{
+			NETMEMORY_MINUS();
+		}
+	}
```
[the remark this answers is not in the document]
DomSyna: intentional

AampGrowableBuffer.h
```diff
@@ -87,26 +92,78 @@ class AampGrowableBuffer
...
+	uint8_t *data() { return buffer.data(); }
+	const uint8_t *data() const { return buffer.data(); }
+
```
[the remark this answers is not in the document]
DomSyna: part of the migration to vector

jfagunde (approved):

PR #997, opened 2026-02-09
Feature/vplay 12623 if get ptr null checks

DomSyna:
Replace the use of GtePtr() in arithmetic, and selection statements.
Optimize self-referential AppendBytes to use memmove() + resize() instead of temporary buffers.


fragmentcollector_hls.cpp
jfagunde: I would update this log line.

MediaStreamContext.cpp
```diff
@@ -490,18 +490,18 @@ bool MediaStreamContext::CacheTsbFragment(std::shared_ptr<CachedFragment> fragme
...
-		if(cachedFragment->fragment.GetPtr())
+		if(cachedFragment->fragment.capacity() != 0)
 		{
 			// If following log is coming, possible memory leak. Need to clear the data first before slot reuse.
 			AAMPLOG_WARN("Fetch buffer has junk data, Need to free this up");
 		}
 		cachedFragment->fragment.clear();
 		cachedFragment->Copy(fragment.get(), fragment->fragment.size());
-		if(cachedFragment->fragment.GetPtr() && cachedFragment->fragment.size() > 0)
+		if(cachedFragment->fragment.capacity() != 0 && cachedFragment->fragment.size() > 0)
```
jfagunde: How about `if (!cachedFragment->fragment.empty())` ?

CachedFragment.cpp
```diff
@@ -73,7 +73,7 @@ void CachedFragment::Copy(CachedFragment* other, size_t len)
 	this->discontinuityIndex = other->discontinuityIndex;
 	
 	// Copy fragment data up to specified length
-	if (other && other->fragment.GetPtr() && len > 0) {
+	if (other && other->fragment.capacity() != 0 && len > 0) {
```
cpc005, line 76: AppendBytes() checks for len==0 so don't really need a check here

MediaStreamContext.cpp
```diff
@@ -850,14 +850,14 @@ bool MediaStreamContext::DownloadFragment(DownloadInfoPtr dlInfo)
 	}
 
 	// Handle change in bandwidth for segmentBase streams, so need to load new range
-	if((dlInfo->bandwidth != fragmentDescriptor.Bandwidth) && IDX.GetPtr() && uriInfo.range.empty())
+	if((dlInfo->bandwidth != fragmentDescriptor.Bandwidth) && IDX.capacity() != 0 && uriInfo.range.empty())
```
jfagunde, line 853: Does it make more sense to check `!IDX.empty()` here and in the conditions below?

priv_aamp.cpp
```diff
@@ -5026,11 +5024,11 @@ bool PrivateInstanceAAMP::GetFile( std::string remoteUrl, AampMediaType mediaTyp
 					getDefaultHarvestPath(harvestPath);
 					AAMPLOG_WARN("Harvest path has not configured, taking default path %s", harvestPath.c_str());
 				}
-				if(buffer->GetPtr() )
+				if (buffer->capacity() != 0)
```
jfagunde, line 5174: How about `!buffer->empty()` or `buffer->size() > 0`?

CachedFragment.cpp
DomSyna: AppendBytes will go soon.

streamabstraction.cpp
```diff
@@ -810,7 +810,7 @@ bool MediaTrack::CheckForDiscontinuity(CachedFragment* cachedFragment, bool& fra
 	StreamAbstractionAAMP* context = GetContext();
 	double injectedDuration = GetTotalInjectedDuration();
 
-	if(cachedFragment->fragment.GetPtr())
+	if(cachedFragment->fragment.capacity() != 0)
```
cpc005, line 814: As I understand it may be possible for for the following to be true
v.empty() ==true && v.capacity() >0
I.E space is allocated but there is nothing in it. So there is potential for the use of v.capacity() to be confusing

priv_aamp.cpp
```diff
@@ -5179,7 +5177,7 @@ bool PrivateInstanceAAMP::GetFile( std::string remoteUrl, AampMediaType mediaTyp
 	}
 
 	// Strip downloaded chunked Iframes when ranged requests receives 200 as HTTP response for HLS MP4
-	if( mConfig->IsConfigSet(eAAMPConfig_RepairIframes) && NULL != range && '\0' != range[0] && 200 == http_code && NULL != buffer->GetPtr() && FORMAT_ISO_BMFF == this->mVideoFormat)
+	if( mConfig->IsConfigSet(eAAMPConfig_RepairIframes) && NULL != range && '\0' != range[0] && 200 == http_code && buffer->capacity() != 0 && FORMAT_ISO_BMFF == this->mVideoFormat)
```
jfagunde: Would `!buffer->empty()` be better?

streamabstraction.cpp
```diff
@@ -1865,7 +1865,7 @@ CachedFragment* MediaTrack::GetFetchChunkBuffer(bool initialize)
 
 	if(initialize && cachedFragment)
 	{
-		if (cachedFragment->fragment.GetPtr() )
+		if (cachedFragment->fragment.capacity() != 0)
```
cpc005, line 1869: Here we are testing if there is segment data stored
fragment.size() > 0 would be a correct test
but fragment. capacity>0 would be a misleading test since a preceding
fragment.clear()
may leave fragment.size() == 0 and fragment.capacity() >0

streamabstraction.cpp
jfagunde, line 814: Does it make more sense to check `if (!cachedFragment->empty())`

streamabstraction.cpp
```diff
@@ -942,7 +942,7 @@ bool MediaTrack::ProcessFragmentChunk()
 	class StreamAbstractionAAMP* pContext = GetContext();
 	//Get Cache buffer
 	CachedFragment* cachedFragment = &this->mCachedFragmentChunks[fragmentChunkIdxToInject];
-	if(cachedFragment != NULL && NULL == cachedFragment->fragment.GetPtr())
+	if(cachedFragment != NULL && cachedFragment->fragment.capacity() == 0)
```
jfagunde, line 946: Does it make more sense to change this condition to:
```
if (cachedFragment == nullptr)
{
	AAMPLOG_WARN("No fragment chunk available");
}
else if (cachedFragment->empty())
{
	if(!SignalIfEOSReached())
	{
		AAMPLOG_TRACE("[%s] Ignore NULL Chunk - cachedFragment->fragment.len %zu", name, cachedFragment->fragment.size());
	}
	return false;
}
```

streamabstraction.cpp
```diff
@@ -1523,9 +1523,9 @@ bool MediaTrack::InjectFragment()
 		}
 
 		AAMPLOG_TRACE("[%s] - fragmentIdxToInject %d cachedFragment %p ptr %p",
-					 name, fragmentIdxToInject, cachedFragment, cachedFragment->fragment.GetPtr());
+					  name, fragmentIdxToInject, cachedFragment, cachedFragment->fragment.GetPtr());
 
-		if (cachedFragment->fragment.GetPtr())
+		if (cachedFragment->fragment.capacity() != 0)
```
jfagunde, line 1529: Does it make more sense to check `if (cachedFragment->fragment.empty())`?

fragmentcollector_hls.cpp
DomSyna: Changed to
AAMPLOG_WARN("%s cachedFragment->fragment is empty", name);

More correct for now, and will be correct once the growable buffer has gone totally

MediaStreamContext.cpp
DomSyna: I considered that, then backed away from the idea. I don't want to change any logic in this delivery as if I break it it will be time consuming to find which bit I broke. 

There are several places I think we can change the capacity check for empty(). I'm hoping that will be done easier as a general tidy up at the end, looking at each vectors life cycle, typing etc

priv_aamp.cpp
DomSyna: Thought about that, but if you have an empty buffer then I think the path and file will be created, and checking empty() will change this behaviour

priv_aamp.cpp
DomSyna: Done

streamabstraction.cpp
DomSyna: Yes, so we just need to check a growable buffer has been allocated, as it currently does

DomSyna: Not without understanding the complete life cycle of the fragment. It may have allocated space and be empty. Which would currently enter this condition and potentially set state vars

streamabstraction.cpp
DomSyna: Quite possibly, but where I wasn't 100% I have kept the logic identical.

streamabstraction.cpp
DomSyna: Probably, but without understanding if you could have an empty growable buffer at this point I'm hesitant to change the logical flow.

streamabstraction.cpp
DomSyna: No we are testing if a buffer has been allocated, and if so providing a warning before clearing it.

jfagunde (approved):

priv_aamp.cpp
```diff
@@ -5191,13 +5189,12 @@ bool PrivateInstanceAAMP::GetFile( std::string remoteUrl, AampMediaType mediaTyp
 				size_t len = (end - start) + 1;
 				if( buffer->size() >= len)
 				{
-					buffer->clear();
-					buffer->AppendBytes(buffer->GetPtr() + start, len);
+					std::memmove(buffer->data(), buffer->data() + start, len);
+					buffer->resize(len);
 				}
```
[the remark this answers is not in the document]
DomSyna: Given to get here we have received HTTP 200 indicating the entire segment has been received and the existing code didn't perform these additional defensive checks I'm inclined to not waste cycles performing redundant checking

priv_aamp.cpp
```diff
@@ -5191,13 +5189,12 @@ bool PrivateInstanceAAMP::GetFile( std::string remoteUrl, AampMediaType mediaTyp
 				size_t len = (end - start) + 1;
 				if( buffer->size() >= len)
 				{
-					buffer->clear();
-					buffer->AppendBytes(buffer->GetPtr() + start, len);
+					std::memmove(buffer->data(), buffer->data() + start, len);
+					buffer->resize(len);
```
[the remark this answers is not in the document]
DomSyna: done

CachedFragment.cpp
```diff
@@ -73,7 +73,7 @@ void CachedFragment::Copy(CachedFragment* other, size_t len)
...
 		this->fragment.AppendBytes(other->fragment.GetPtr(), len);
 	}
```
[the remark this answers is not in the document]
DomSyna: other can't be NULL. This will be resolved with the removal of AppendBytes

priv_aamp.cpp
```diff
@@ -5311,13 +5309,12 @@ bool PrivateInstanceAAMP::GetFile( std::string remoteUrl, AampMediaType mediaTyp
...
```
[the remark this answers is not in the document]
DomSyna: Previously commented on this

pstroffolino (approved):

PR #1012, opened 2026-02-11
Feature/vplay 12671 replace append bytes calls

DomSyna:
### Summary

Replacing `AppendBytes()` method with vector like alternatives `assign()` and `insert()`. 

**Key Changes:**
- **Replaced `AppendBytes()` calls** with `assign()` (for replacement) or `insert()` (for appending) across production and test code
- **Removed redundant `clear()` calls** before `assign()` operations (assign already clears)
- **Eliminated unnecessary casts** by using `data()` instead of `GetPtr()` with reinterpret_cast in buffer operations
- **Updated `CachedFragment::Copy()`** to use `assign()` without explicit length parameter

All changes maintain backward compatibility and test coverage.


mp4demux/MP4Demux.cpp
```diff
@@ -717,7 +717,9 @@ void Mp4Demux::ParseTrackRun()
 		{
 			throw Mp4ParseException(MP4_PARSE_ERROR_DATA_BOUNDARY_MISMATCH, "trun: sample payload OOB");
 		}
-		newSample.mData.AppendBytes(dataPtr, sampleLen);
+		newSample.mData.insert(newSample.mData.GetVector().end(),
+				reinterpret_cast<const uint8_t*>(dataPtr),
```
jfagunde: Why do you need the cast if dataPtr is already a const uint8_t*? Here and other places.

test/utests/tests/AampTSBSessionManager/FunctionalTests.cpp
```diff
@@ -170,7 +170,7 @@ TEST_F(FunctionalTests, TSBWriteTests)
 	cachedFragment->duration = 0;
 	cachedFragment->position = 0;
 	cachedFragment->absPosition = 1234.0;
-	cachedFragment->fragment.AppendBytes(TEST_DATA, strlen(TEST_DATA));
+	cachedFragment->fragment.assign(reinterpret_cast<const uint8_t*>(TEST_DATA), reinterpret_cast<const uint8_t*>(TEST_DATA) + strlen(TEST_DATA));
```
jfagunde: How about defining TEST_DATA as const uint8_t* instead of casting it everywhere? And the same comment applies to other tests.

test/utests/tests/AdFallbackTests/FunctionalTests.cpp
```diff
@@ -268,7 +268,8 @@ class AdFallbackTests : public ::testing::Test
 		{
 			/* Setup fake AampGrowableBuffer contents. */
 			buffer->clear();
-			buffer->AppendBytes((char *)mAdManifest, strlen(mAdManifest));
+			buffer->assign(reinterpret_cast<const uint8_t*>(mAdManifest),
```
jfagunde: Same comment, why not define mAdManifest as const uint8_t* instead of casting it?

AampGrowableBuffer.h
```diff
@@ -129,7 +129,21 @@ class AampGrowableBuffer
 			NETMEMORY_PLUS();
 		}
 	}
+	void assign(const uint8_t *first, const uint8_t *last)
```
jfagunde: There is a version of assign that takes a size and a value: https://en.cppreference.com/w/cpp/container/vector/assign.html
`void assign( size_type count, const T& value );`
This is closer of the prototype of AppendBytes() that is being replaced and would mean less casting.

tsDemuxer.cpp
```diff
@@ -457,7 +457,9 @@ void Demuxer::processPacket(const unsigned char * packetStart, bool &basePtsUpda
 						bytes_to_read = size;
 					}
 					AAMPLOG_DEBUG("PES_STATE_GETTING_HEADER. size = %d, bytes_to_read =%d", size, bytes_to_read);
-					pes_header.AppendBytes( data, bytes_to_read);
+					pes_header.insert(pes_header.GetVector().end(),
```
jfagunde, line 461: How about implementing other versions of insert() that only take the position and object to insert or the position, object and size?

CachedFragment.cpp
```diff
@@ -72,9 +72,10 @@ void CachedFragment::Copy(CachedFragment* other, size_t len)
 	this->isDummy = other->isDummy;
 	this->discontinuityIndex = other->discontinuityIndex;
 	
-	// Copy fragment data up to specified length
-	if (other && other->fragment.GetPtr() && len > 0) {
-		this->fragment.AppendBytes(other->fragment.GetPtr(), len);
+	// Copy fragment data vector from other to this using assign method
+	if (!other->fragment.GetVector().empty()) {
```
jfagunde: Open brace should be in a new line

CachedFragment.h
```diff
@@ -125,9 +125,8 @@ class CachedFragment
 	/**
 	 * @brief Copy content from another CachedFragment
 	 * @param other Source CachedFragment to copy from
-	 * @param len Length of data to copy from the fragment buffer
 	 */
-	void Copy(CachedFragment* other, size_t len);
+	void Copy(CachedFragment* other);
```
jfagunde, line 129: Are we always copy the whole of the other cached fragment?

AampGrowableBuffer.h
jfagunde: This suggestion is not valid. The suggested version of assign() takes one item and a counter, not a size (the item would be repeated n times).

tsDemuxer.cpp
jfagunde: The same as with assign(), this suggestion is not valid. The version of insert() mention takes one item and a counter, not a size (the item would be repeated n times).

isobmff/isobmffprocessor.cpp
```diff
@@ -1183,7 +1183,8 @@ void IsoBmffProcessor::cacheRestampInitSegment(AampMediaType type,char *segment,
 	stInitRestampSegment *pSt = new stInitRestampSegment;
 	memset(pSt,0,sizeof(stInitRestampSegment));
 	pSt->buffer =  new AampGrowableBuffer("cached-restamp-init-segment");
-	pSt->buffer->AppendBytes(segment, size);
+	pSt->buffer->assign(reinterpret_cast<const uint8_t *>(segment),
+						reinterpret_cast<const uint8_t *>(segment) + size);
```
jfagunde: Is the last element `segment + size` or `segment + size - 1`?

CachedFragment.h
DomSyna: I think so

AampGrowableBuffer.h
```diff
@@ -119,17 +119,35 @@ class AampGrowableBuffer
 			NETMEMORY_PLUS();
 		}
 	}
-	void insert(typename std::vector<uint8_t>::const_iterator pos, const uint8_t *first, const uint8_t *last)
+	void insert(typename std::vector<uint8_t>::const_iterator pos, const void *first, const void *last)
 	{
+		const uint8_t* start = static_cast<const uint8_t*>(first);
+		const uint8_t* end = static_cast<const uint8_t*>(last);
```
jfagunde, line 125: Are these casts needed or would buffer.insert() cope with void* parameters? Same question for assign() method.

DomSyna: No as the vector methods are templated (uint8_t in this case). So since the buffer is a vector of uint8_t and the compiler doesn't know the size of void we have to cast so it knows

MediaStreamContext.cpp
```diff
@@ -222,7 +222,7 @@ bool MediaStreamContext::CacheFragmentChunk(AampMediaType actualType, const char
 		cachedFragment->absPosition = 0;
 		cachedFragment->type = actualType;
 		cachedFragment->downloadStartTime = dnldStartTime;
-		cachedFragment->fragment.AppendBytes(ptr, size);
+		cachedFragment->fragment.assign(ptr, ptr + size);
 		cachedFragment->timeScale = fragmentDescriptor.TimeScale;
```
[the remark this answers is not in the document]
DomSyna: If ptr is nullptr and size is 0, the result of ptr + size is nullptr
Given vector.assign() does nothing if the range is empty (first == last) so it should be fine 

if ptr is nullptr and size > 0 that would be an issue but that shouldn't be possible.

AampGrowableBuffer.h
```diff
@@ -119,17 +119,35 @@ class AampGrowableBuffer
...
+		buffer.assign(start, end);
+
+		const size_t newCap = buffer.capacity();
 
+		// Honour the AAMP memory tracking logic
+		if (oldCap == 0 && newCap > 0)
+		{
+			NETMEMORY_PLUS();
+		}
+	}
```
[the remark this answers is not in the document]
DomSyna: From the spec:
"When an expression that has integral type is added to or subtracted from a pointer, the result has the type of the pointer operand. If the expression P has type “pointer to $cv$ $T$” and points to an element of an array object x with $n$ elements, the expressions P + J and J + P (where J has value $j$) point to an element of x if $0 \le j \le n$ [...] **If the value 0 is added to or subtracted from a pointer value, the result is the original pointer value**."

So as nullptr is a ptr + 0 does nothing.

jfagunde (approved):

pstroffolino (approved):

PR #1043, opened 2026-02-17
VPLAY-12637 [Rogers] [VIPA] AAMP does not return all available CC tracks

anjali-syna:
Reason for change: AAMP does not return all the available CEA-708
 Closed Captions tracks in a Multiview stream

Changes:
* Added support to return all the available CEA-708 Closed Captions tracks in a Multiview stream

Risks: low


dnfaulkner (commented):
Looks OK to me

PR #1044, opened 2026-02-17
VPLAY-12581 : [Rogers-RTK]Seamless audio transition with PTS Restamp …

molakalapalliharipriya:
…is not working

Reason for change:
1. For PTS Restamp case, position passed to FlushTrack() was incorrect; converted to relative position using PTS offset adjustments - this correction resolved the complete audio loss issue
2. In AAMP TSB mode, culledSeconds is set by the AAMP TSB Manager based on actual fragment position; changed offsetFromStart calculation to be relative to what's actually available in the manifest, not relative to the TSB window start - this resolved the audio loss issue on audio switch
3. diffInInjectedDuration - Fixed timeline domain mismatch calculation

Test Procedure: Test with audio track switching with PTS Restamp enabled

Risks: Low
priority : p1


fragmentcollector_mpd.cpp
Vinish100: Can we remove newInjectedPosition altogether as its not used now.??

PR #1073, opened 2026-02-23
VPLAY-12760: AAMP Trackworker Q causes early EOS and L2 2025 unreliable

nu641001:
Reason for change: During VOD trickplay, it can cause early EOS due to missing check in ChekForEndOfStream
Risks: Low
Test Procedure: Test with L2 2025 multiple times
Priority: P1
Signed-off-by: Nandakishor U M <nu641001@gmail.com>


Vinish100 (commented):
Looks good to me 👍

PR #1084, opened 2026-02-26
RDKEMW-13801 : MTK SoC-Interface support

narenr94:
RDKEMW-13801 : MTK SoC-Interface support
Reason for change : MTK doesnt have dedicated SoC-Interface implemented, leading to patches for mtk specific changes
Test Steps : mentioned in ticket
Signed-off by : R.Naren <naren_ramesh@comcast.com>


middleware/vendor/mtk/MtkSocInterface.cpp
```diff
@@ -0,0 +1,273 @@
...
+/**
+ * @brief Check if AppSrc should be used for progressive playback.
+ *
+ * Determines whether the AppSrc element should be used in the current context.
+ *
+ * @return True if AppSrc should be used, false otherwise.
+ */
+bool MtkSocInterface::UseAppSrc()
+{
+#if defined (__APPLE__)
```
pstroffolino, line 38: why #if defined (__APPLE) in several places within MtkSocInterface?

middleware/vendor/mtk/MtkSocInterface.cpp
```diff
@@ -0,0 +1,273 @@
...
+{
+#if !defined(UBUNTU)
+	return false;
+#endif
+	return true;
+}
+
+void MtkSocInterface::ConfigurePluginPriority()
+{
+#ifdef UBUNTU
```
pstroffolino, line 147: why UBUNTU-specific ifdefs in MtkSocInterface?

pstroffolino (approved):

middleware/vendor/mtk/MtkSocInterface.cpp
narenr94: MTK prior to this change was using DefaultSoCInterface and mods and patches are based off that ... hence for template we used the same one which is also being used by the simulators ... main goal of this ticket is to introduce Interface for MTK .. we will cleanup and streamline it for mtk in upcoming tasks

middleware/vendor/mtk/MtkSocInterface.cpp
narenr94: MTK prior to this change was using DefaultSoCInterface and mods and patches are based off that ... hence for template we used the same one which is also being used by the simulators ... main goal of this ticket is to introduce Interface for MTK .. we will cleanup and streamline it for mtk in upcoming tasks

PR #1087, opened 2026-02-26
VPLAY-12785 Playback of some linear test streams freezes after a few seconds using AAMP (regression

cpc005:
Reason for change: Some test streams could not be played by aamp due to a regression.

- Rename mLocalUtcTime to mServerUtcTime to be clearer what it holds
- Change FindServerUTCTime() to give a valid mServerUtcTime even when it does not fetch time from the network.
- Update L1 tests to verify the value of mServerUtcTime
- Improve cohesion: move call to SetLocalTimeDelta(..) next to SetHasServerUtcTime(...)


test/utests/tests/StreamAbstractionAAMP_MPD/FunctionalTests.cpp
```diff
@@ -464,6 +464,11 @@ class StreamAbstractionAAMP_MPDTest :   public FunctionalTestsBase,
 	class TestableStreamAbstractionAAMP_MPD : public StreamAbstractionAAMP_MPD
 	{
 	public:
+	    // Add this getter for testing
+    	double GetServerUtcTime() const
```
DomSyna: tabs not spaces ?

test/utests/tests/StreamAbstractionAAMP_MPD/FunctionalTests.cpp
```diff
@@ -3128,18 +3135,20 @@ TEST_F(StreamAbstractionAAMP_MPDTest, FindServerUTCTime_SkipSyncBeforeInterval)
 	xmlTextReaderPtr reader = xmlReaderForMemory(manifestXml, strlen(manifestXml), NULL, NULL, 0);
 	ASSERT_NE(reader, nullptr);
 	ASSERT_TRUE(xmlTextReaderRead(reader));
-	
+
 	Node *rootNode = MPDProcessNode(&reader, "http://example.com/manifest.mpd");
 	ASSERT_NE(rootNode, nullptr);
 
 	// First call - should sync
```
DomSyna: Maybe the comment should explain you are expecting a call to GetNetworkTime here, and the next call will be calcuating the tie instead of querying the Netork. when we look at this again in 6 months I 'm concerned we won;t know what the test is testing

test/utests/tests/StreamAbstractionAAMP_MPD/FunctionalTests.cpp
cpc005: done

test/utests/tests/StreamAbstractionAAMP_MPD/FunctionalTests.cpp
cpc005: done

DomSyna (approved):

pstroffolino (approved):

PR #1088, opened 2026-02-26
VLAY-12809 calls to GetPtr() with data() ready for removal of AampGrowableBuffer

DomSyna:
Replace remaining calls to GetPtr() with data() ready for removal of AampGrowableBuffer

**Changes:**

- Replaces GetPtr() with data() with type casting as appropriate
- Changed the typing an a few localised module to make coherent and improve the interfaces


isobmff/isobmffprocessor.cpp
```diff
@@ -589,7 +589,7 @@ void IsoBmffProcessor::restampPTSAndSendSegment(AampGrowableBuffer *pBuffer,doub
 			}
 
 			//Step 6.Now time to restamp the PTS
-			buffer.restampPTS(sumPTS,currentPTS,(uint8_t *)(pBuffer->GetPtr()),(uint32_t)(pBuffer->size()));
+			buffer.restampPTS(sumPTS,currentPTS,(uint8_t *)(pBuffer->data()),(uint32_t)(pBuffer->size()));
```
jfagunde: Can we remove the (uint8_t*) cast?

ElementaryProcessor.cpp
```diff
@@ -45,20 +45,18 @@ ElementaryProcessor::~ElementaryProcessor()
...
 	if (ret)
 	{
 		AAMPLOG_INFO("IsoBmffProcessor:: eMEDIATYPE_SUBTITLE sending segment at pos:%f dur:%f", position, duration);
 		sendStream(pBuffer, position, duration, fragmentPTSoffset, discontinuous, isInit);
 	}
 	return true;
-}
-
-/**
+}/**
```
jfagunde: This change seems unintentional. I would revert it (move comment to its own line).

MediaStreamContext.cpp
```diff
@@ -245,7 +245,7 @@ bool MediaStreamContext::CacheFragmentChunk(AampMediaType actualType, const char
 		 */
 		cachedFragment->PTSOffsetSec = GetContext()->mPTSOffset.inSeconds();
 
-		AAMPLOG_TRACE("[%s] cachedFragment %p ptr %p", name, cachedFragment, cachedFragment->fragment.GetPtr());
+		AAMPLOG_TRACE("[%s] cachedFragment %p ptr %p", name, cachedFragment, static_cast<const void*>(cachedFragment->fragment.data()));
```
jfagunde: Since we are not aware of any issues using %p with uint8_t* and AAMP code is full of that, do we need the casting? The code seems more complex (even if it is more correct). The same comment applies to other changes like the one to line 641.

streamabstraction.cpp
```diff
@@ -1531,10 +1531,8 @@ bool MediaTrack::InjectFragment()
 			AAMPLOG_TRACE("[%s] fragmentIdxToInject : %d Discontinuity %d ", name, fragmentIdxToInject, cachedFragment->discontinuity);
 		}
 
-		AAMPLOG_TRACE("[%s] - fragmentIdxToInject %d cachedFragment %p ptr %p",
-					  name, fragmentIdxToInject, cachedFragment, cachedFragment->fragment.GetPtr());
-
-		if (cachedFragment->fragment.capacity() != 0)
+	AAMPLOG_TRACE("[%s] - fragmentIdxToInject %d cachedFragment %p ptr %p",
+				  name, fragmentIdxToInject, cachedFragment, cachedFragment->fragment.data());		if (cachedFragment->fragment.capacity() != 0)
```
jfagunde: Moving the if to the same line as the TRACE seems like an unintentional change. I'd revert it back.

jfagunde:
Should the method GetPtr() be deleted from AampGrowableBuffer?

isobmff/isobmffprocessor.cpp
DomSyna: Done

MediaStreamContext.cpp
DomSyna: Done

streamabstraction.cpp
DomSyna: Done

ElementaryProcessor.cpp
DomSyna: Done

DomSyna:
> Should the method GetPtr() be deleted from AampGrowableBuffer?

Not yet, that will go with the removal of growable buffer, but for now ....

test/utests/tests/StreamAbstractionAAMP_HLS/FunctionalTests.cpp
```diff
@@ -1565,7 +1565,7 @@ TEST_F(TrackStateTests, FindTimedMetadata_New)
     EXPECT_CALL(*g_mockAampConfig, IsConfigSet(eAAMPConfig_EnableSubscribedTags)).WillOnce(Return(true));
     AampGrowableBuffer buffer("tsProcessor PAT/PMT test");
         //buffer.AppendBytes(10);
-    buffer.GetPtr();
+    buffer.data();
```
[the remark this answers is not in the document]
DomSyna: This looks like a half completed test, so removed the unised vars

fragmentcollector_hls.cpp
```diff
@@ -3299,7 +3299,7 @@ AAMPStatusType StreamAbstractionAAMP_HLS::Init(TuneType tuneType)
 	{
 		if( AampLogManager::isLogLevelAllowed(eLOGLEVEL_TRACE) )
 		{ // use printf to avoid 2048 char syslog limitation
-			printf("***Main Manifest***:\n\n%.*s\n************\n", (int)this->mainManifest.size(), this->mainManifest.GetPtr());
+			printf("***Main Manifest***:\n\n%.*s\n************\n", (int)this->mainManifest.size(), reinterpret_cast<const char*>(this->mainManifest.data()));
```
jfagunde, line 3302: [Existing code] This seems wrong: the format specifier '.*' specifies the precision of a number. I don't think it makes sense to use it with a string. Should it be '%*s' to specify the width instead? The same comment applies to other calls to printf() in this file. Maybe a separate Jira should be raised for this?

jfagunde (approved):

fragmentcollector_hls.cpp
DomSyna: Its probably correct. In this scenario its a magic way to deal with strings that may not be null terminated. The * indicates the variable length nature. I don't know how it works internally though, so will go read later ;)

jfagunde: Please, ignore this comment. The code is correct.

DomSyna: totally valid, the size comes from the size before the pointer in the params list

pstroffolino (approved):

tsprocessor.cpp
```diff
@@ -1868,11 +1868,11 @@ bool TSProcessor::sendSegment(AampGrowableBuffer* pBuffer, double position, doub
...
 	if (len % m_packetSize)
 	{
-		int discardAtEnd = len % m_packetSize;
-		AAMPLOG_INFO("Discarding %d bytes at end", discardAtEnd);
+		size_t discardAtEnd = len % m_packetSize;
+		AAMPLOG_INFO("Discarding %zu bytes at end", discardAtEnd);
 		len = len - discardAtEnd;
 	}
-	ret = processBuffer((unsigned char*)packetStart, len, insPatPmt, discontinuous);
+	ret = processBuffer(packetStart, static_cast<int>(len), insPatPmt, discontinuous);
```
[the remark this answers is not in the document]
pstroffolino: maybe later

PR #1101, opened 2026-02-27
VPLAY-12834 ParseSegmentIndexBox incorrect index

DomSyna:
Turns out it wasn't a bug, just confusing code.
Since 
const char **f = &start;
So *f and start are effectively the same thing.

Refactored to bring it forward to modern cpp, and remove confusing code

Added L1 tests

### DIFF
--- a/AampMPDUtils.cpp
+++ b/AampMPDUtils.cpp
@@ -296,6 +296,7 @@ bool ParseSegmentIndexBox( const uint8_t *start, size_t size, int segmentIndex, 
 {
 	if ((!start) || (size < 4))
 	{
+		AAMPLOG_WARN("Invalid parameters in ParseSegmentIndexBox: start=%p, size=%zu", start, size);
 		return false;
 	}
 
@@ -342,7 +343,10 @@ bool ParseSegmentIndexBox( const uint8_t *start, size_t size, int segmentIndex, 
 	else
 	{
 		AAMPLOG_WARN("Unsupported version in ParseSegmentIndexBox %u found, 0 or 1 expected", version);
-		if (firstOffset) *firstOffset = 0;
+		if (firstOffset) 
+		{
+			*firstOffset = 0;
+		}
 		return false;
 	}
 
--- a/AampMPDUtils.cpp
+++ b/AampMPDUtils.cpp
@@ -252,7 +252,8 @@ bool ParseSegmentIndexBox( const uint8_t *start, size_t size, int segmentIndex, 
 		return false;
 	}
 
-	auto version   = reader.Read<uint32_t>();    // version (8b) + flags (24b)
+	// version is the top 8 bits; lower 24 bits are flags (ignored)
+	const auto version = static_cast<uint8_t>(reader.Read<uint32_t>() >> 24);
 	reader.Skip<uint32_t>();                     // reference_ID
 	auto timescale = reader.Read<uint32_t>();    // timescale
 
@@ -277,7 +278,7 @@ bool ParseSegmentIndexBox( const uint8_t *start, size_t size, int segmentIndex, 
 		return true;
 	}
 
-	if (segmentIndex < static_cast<int>(reference_count))
+	if (segmentIndex >= 0 && segmentIndex < static_cast<int>(reference_count))
 	{
 		reader.Skip(SIDX_ENTRY_SIZE * segmentIndex);
 
--- a/AampMPDUtils.cpp
+++ b/AampMPDUtils.cpp
@@ -19,6 +19,7 @@
 
 #include "AampMPDUtils.h"
 #include <inttypes.h>
+#include <type_traits>
 
 /**
  * @brief Get xml node form reader
@@ -211,6 +212,76 @@ double ComputeFragmentDuration( uint32_t duration, uint32_t timeScale )
 	return newduration;
 }
 
+namespace
+{
+/**
+ * @brief Lightweight cursor for reading big-endian ISOBMFF box fields.
+ *
+ * Provides typed Read<T>() and Skip<T>() operations.
+ * The caller is responsible for validating the box size field before
+ * reading individual fields.
+ */
+class BoxReader final
+{
+public:
+	constexpr explicit BoxReader(const uint8_t *data) noexcept
+		: mCursor{data}
+	{
+	}
+
+	/**
+	 * @brief Read a big-endian integer field and advance the cursor.
+	 * @tparam T  Integral type whose sizeof determines the field width.
+	 * @return    The value read in host byte order.
+	 */
+	template <typename T>
+	[[nodiscard]] T Read() noexcept
+	{
+		static_assert(std::is_integral_v<T>,
+					  "Read<T> requires an integral type");
+		static_assert(sizeof(T) == 1 || sizeof(T) == 2 ||
+					  sizeof(T) == 4 || sizeof(T) == 8,
+					  "Read<T> only supports 1/2/4/8-byte types");
+
+		uint64_t val{0};
+		for (size_t i = 0; i < sizeof(T); ++i)
+		{
+			val = (val << 8) | static_cast<uint8_t>(mCursor[i]);
+		}
+		mCursor += sizeof(T);
+		return static_cast<T>(val);
+	}
+
+	/**
+	 * @brief Skip a field without reading it.
+	 * @tparam T  Integral type whose sizeof determines the skip width.
+	 */
+	template <typename T>
+	void Skip() noexcept
+	{
+		static_assert(std::is_integral_v<T>,
+					  "Skip<T> requires an integral type");
+		static_assert(sizeof(T) == 1 || sizeof(T) == 2 ||
+					  sizeof(T) == 4 || sizeof(T) == 8,
+					  "Skip<T> only supports 1/2/4/8-byte types");
+
+		mCursor += sizeof(T);
+	}
+
+	/**
+	 * @brief Skip a runtime number of bytes.
+	 * @param n  Number of bytes to advance.
+	 */
+	void Skip(size_t n) noexcept
+	{
+		mCursor += n;
+	}
+
+private:
+	const uint8_t *mCursor;
+};
+} // namespace
+
 /**
  * @brief Parse segment index box
  * @note The SegmentBase indexRange attribute points to Segment Index Box location with segments and random access points.
@@ -223,7 +294,7 @@ double ComputeFragmentDuration( uint32_t duration, uint32_t timeScale )
  */
 bool ParseSegmentIndexBox( const uint8_t *start, size_t size, int segmentIndex, unsigned int *referenced_size, float *referenced_duration, unsigned int *firstOffset)
 {
-	if (!start)
+	if ((!start) || (size < 4))
 	{
 		return false;
 	}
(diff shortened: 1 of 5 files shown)


AampMPDUtils.cpp
```diff
@@ -225,66 +225,72 @@ bool ParseSegmentIndexBox( const char *start, size_t size, int segmentIndex, uns
...
+		*referenced_duration = Read32(f) / static_cast<float>(timescale);
 
-		unsigned int flags = Read32(f);
-		(void)flags;
-		// starts_with_SAP (1 bit)
-		// SAP_type (3 bits)
-		// SAP_delta_time (28 bits)
+		// Skip SAP flags (4 bytes):
+		// starts_with_SAP (1 bit) | SAP_type (3 bits) | SAP_delta_time (28 bits)
+		*f += 4;
```
dnfaulkner: What's the point of doing this at all? Nothing happens (f isn't used) after this.

AampMPDUtils.cpp
```diff
@@ -225,66 +225,72 @@ bool ParseSegmentIndexBox( const char *start, size_t size, int segmentIndex, uns
...
+	*f += 4;
+	auto timescale = Read32(f);
+
+	uint64_t first_offset{0};
+	if (version == 0)
 	{
-		earliest_presentation_time = Read32(f);
-		(void)earliest_presentation_time; // unused
+		// Skip earliest_presentation_time (4 bytes)
+		*f += 4;
```
dnfaulkner: In harmony with Read32()/Read64(), I'd be tempted to use a helper method/macro:
Skip32(f);

AampMPDUtils.cpp
```diff
@@ -225,66 +225,72 @@ bool ParseSegmentIndexBox( const char *start, size_t size, int segmentIndex, uns
...
 		first_offset = Read32(f);
 	}
 	else
 	{
-		earliest_presentation_time = Read64(f);
-		(void)earliest_presentation_time; // unused
+		// Skip earliest_presentation_time (8 bytes)
+		*f += 8;
```
dnfaulkner: In harmony with Read32()/Read64(), I'd be tempted to use a helper method/macro:
Skip64(f);

dnfaulkner (commented):
.

test/utests/tests/AampMPDUtils/FunctionalTests.cpp
```diff
@@ -100,3 +103,303 @@ TEST(AampMPDUtils, ComputeFragmentDurationTest1)
...
+ */
+TEST(AampMPDUtils, ParseSegmentIndexBox_FirstOffset_ReturnsCorrectValue)
+{
+	constexpr uint32_t EXPECTED_OFFSET = 42;
+	auto sidx = BuildSidxBoxV0(1000, {{100, 1000}}, EXPECTED_OFFSET);
+	unsigned int firstOff = 0;
+	EXPECT_TRUE(ParseSegmentIndexBox(sidx.data(), sidx.size(), 0,
+									 nullptr, nullptr, &firstOff));
+	EXPECT_EQ(firstOff, EXPECTED_OFFSET);
+}
```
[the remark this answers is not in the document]
DomSyna: done

AampMPDUtils.h
```diff
@@ -96,33 +97,71 @@ void ConstructFragmentURL( std::string& fragmentUrl, const FragmentDescriptor *f
...
+	template <typename T>
+	[[nodiscard]] T Read() noexcept
+	{
+		static_assert(std::is_integral_v<T>,
+					  "Read<T> requires an integral type");
+		static_assert(sizeof(T) == 1 || sizeof(T) == 2 ||
+					  sizeof(T) == 4 || sizeof(T) == 8,
+					  "Read<T> only supports 1/2/4/8-byte types");
+
+		uint64_t val{0};
```
jfagunde: Could val be defined as a variable of type T and remove the casting from line 134?

AampMPDUtils.h
```diff
@@ -96,33 +97,71 @@ void ConstructFragmentURL( std::string& fragmentUrl, const FragmentDescriptor *f
...
- * @param[in] n word size in bytes
- * @retval 32 bit value
+ * @brief Lightweight cursor for reading big-endian ISOBMFF box fields.
+ *
+ * Provides typed Read<T>() and Skip<T>() operations.
+ * The caller is responsible for validating the box size field before
+ * reading individual fields.
  */
-uint64_t ReadWordHelper( const uint8_t **pptr, int n );
+class BoxReader final
```
jfagunde: How about moving this class definition to BoxReader.h file (or AampBoxReader.h)?

AampMPDUtils.cpp
```diff
@@ -225,123 +225,76 @@ bool ParseSegmentIndexBox( const uint8_t *start, size_t size, int segmentIndex,
...
+	auto timescale = reader.Read<uint32_t>();    // timescale
+
+	uint64_t first_offset{0};
+	if (version == 0)
 	{
-		earliest_presentation_time = Read32(f);
-		(void)earliest_presentation_time; // unused
-		first_offset = Read32(f);
+		reader.Skip<uint32_t>();                 // earliest_presentation_time
+		first_offset = reader.Read<uint32_t>();  // first_offset
```
jfagunde, line 266: Should we add a static cast here?

test/utests/tests/AampMPDUtils/FunctionalTests.cpp
```diff
@@ -100,3 +103,467 @@ TEST(AampMPDUtils, ComputeFragmentDurationTest1)
...
+
+/**
+ * @brief Negative segment index returns false without UB.
+ *
+ * A negative segmentIndex must be rejected before the skip arithmetic,
+ * which would otherwise convert a negative product to a huge size_t.
+ */
+TEST(AampMPDUtils, ParseSegmentIndexBox_NegativeIndex_ReturnsFalse)
+{
+	auto sidx = BuildSidxBoxV0(1000, {{100, 1000}});
```
jfagunde: Why not use constexpr, like it is done in the other tests, to clarify what each value is? The same comment applies to other following tests with magical numbers.

test/utests/tests/AampMPDUtils/FunctionalTests.cpp
```diff
@@ -100,3 +103,467 @@ TEST(AampMPDUtils, ComputeFragmentDurationTest1)
...
+ * The version byte (top 8 bits) is 0; the lower 24 bits carry arbitrary flags.
+ * A correct parser must isolate the version byte and not treat the flags as
+ * part of the version.
+ *
+ * @param[in] flags       24-bit flags value (lower 24 bits only).
+ * @param[in] timescale   Timescale for duration calculation.
+ * @param[in] entries     Vector of reference entries.
+ * @return                Byte vector containing the complete SIDX box.
+ */
+static std::vector<uint8_t> BuildSidxBoxV0WithFlags(
```
jfagunde: This function is very similar to the previous one. how about trying to reduce duplicated code? For example, this function could call the previous one and set the flags afterwards or call the same helper function from both.

AampMPDUtils.cpp
DomSyna: I don't think so, as you are widening compilers are normally happy

AampMPDUtils.h
DomSyna: We would then have to cast the (val << 8) | static_cast<uint8_t>(mCursor[i]); since if we were reading uint16_t the compiler will promote the values to int type, and then narrowing with the cast

AampMPDUtils.h
DomSyna: Its directly related to the code in the .cpp file so probably better to stay here for now. I'm thinking this should probably be moved into the ISOBmff code but maybe it doesn't fit in there nicely, and hence put here.

test/utests/tests/AampMPDUtils/FunctionalTests.cpp
DomSyna: agreed, will look at helper / merging

test/utests/tests/AampMPDUtils/FunctionalTests.cpp
DomSyna: yes, I'm going to blame the AI for that lol

AampMPDUtils.cpp
```diff
@@ -223,123 +294,82 @@ double ComputeFragmentDuration( uint32_t duration, uint32_t timeScale )
...
+		AAMPLOG_WARN("Wrong size in ParseSegmentIndexBox %u found, %zu expected", len, size);
 		if (firstOffset) *firstOffset = 0;
 		return false;
 	}
 
-	unsigned int type = Read32(f);
-	if (type != 'sidx')
+	auto type = reader.Read<uint32_t>();
+	if (type != SIDX_BOX_TYPE)
 	{
```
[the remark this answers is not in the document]
DomSyna: excessive checking on something we should be able to truse trust

AampMPDUtils.cpp
```diff
@@ -223,123 +294,82 @@ double ComputeFragmentDuration( uint32_t duration, uint32_t timeScale )
...
+		*referenced_size = reader.Read<uint32_t>() & 0x7fffffff;
+		// top bit is "reference_type"
 
-		unsigned int flags = Read32(f);
-		(void)flags;
-		// starts_with_SAP (1 bit)
-		// SAP_type (3 bits)
-		// SAP_delta_time (28 bits)
+		*referenced_duration = reader.Read<uint32_t>() /
+							   static_cast<float>(timescale);
```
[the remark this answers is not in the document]
DomSyna: We should be able to trust the sidx contents

test/utests/tests/AampMPDUtils/FunctionalTests.cpp
```diff
@@ -18,9 +18,12 @@
 */
 
 #include <cstdlib>
+#include <cstdint>
 #include <iostream>
 #include <string>
 #include <string.h>
+#include <vector>
+#include <utility>
```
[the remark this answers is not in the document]
DomSyna: done

AampMPDUtils.cpp
```diff
@@ -223,7 +294,7 @@ double ComputeFragmentDuration( uint32_t duration, uint32_t timeScale )
  */
 bool ParseSegmentIndexBox( const uint8_t *start, size_t size, int segmentIndex, unsigned int *referenced_size, float *referenced_duration, unsigned int *firstOffset)
 {
-	if (!start)
+	if ((!start) || (size < 4))
 	{
 		return false;
```
jfagunde, line 230: How about printing a WARN here?

AampMPDUtils.cpp
```diff
@@ -263,11 +334,17 @@ bool ParseSegmentIndexBox( const uint8_t *start, size_t size, int segmentIndex,
...
-	else
+	else if (version == 1)
 	{
 		reader.Skip<uint64_t>();                 // earliest_presentation_time
 		first_offset = reader.Read<uint64_t>();  // first_offset
 	}
+	else
+	{
+		AAMPLOG_WARN("Unsupported version in ParseSegmentIndexBox %u found, 0 or 1 expected", version);
+		if (firstOffset) *firstOffset = 0;
```
jfagunde: Following AAMP coding guidelines, this should be re-written as:
```
if (firstOffset)
{
	*firstOffset = 0;
}
```

DomSyna: Done, those guidelines are contrary to the AI instructions !

AampMPDUtils.cpp
DomSyna: done

jfagunde (approved):

pstroffolino (approved):

PR #1113, opened 2026-03-02
VPLAY-12854 getFile() to use std vector

DomSyna:
**Summary**
Part of the ongoing removal of AampGrowableBuffer from the AAMP codebase. This PR replaces the raw AampGrowableBuffer* parameter on PrivateInstanceAAMP::GetFile() and CurlCallbackContext::buffer with std::vector<uint8_t>&, eliminating pointer indirection and manual memory lifetime management at the download interface boundary.

**Changes**
Changes the signature, and internal use from GetFile() down into the Curl callbacks to replace AampGrowableBuffer with a std::vector.

Removed NETMEMORY_PLUS/NETMEMORY_MINUS tracking methods from AampGrowableBuffer as now the vector is being modified directly and the tracking would give false reports

Change the tests gated behind #ifdef AAMP_NET_TRACE in PrivAampTests.cpp, though I don't think they have ever been run as there are obvious compiler issues. I added the missing , but didn't address the forward declaration typing issues etc


admanager_mpd.cpp
```diff
@@ -914,7 +914,7 @@ MPD* PrivateCDAIObjectMPD::GetAdMPD(std::string &manifestUrl, bool &finalManifes
 	AampGrowableBuffer manifest("adMPD_CDN");
 	bool gotManifest = false;
 	std::string effectiveUrl;
-	gotManifest = mAamp->GetFile(manifestUrl, eMEDIATYPE_MANIFEST, &manifest, effectiveUrl, &http_error, &downloadTime, NULL, eCURLINSTANCE_DAI);
+	gotManifest = mAamp->GetFile(manifestUrl, eMEDIATYPE_MANIFEST, manifest.GetVector(), effectiveUrl, &http_error, &downloadTime, NULL, eCURLINSTANCE_DAI);
```
cpc005, line 917: Will AampGrowableBuffer  disappear completely so we loose the potential of tracking where some buffer got allocated via the name attribute?

cpc005 (approved):

admanager_mpd.cpp
DomSyna: Yes

priv_aamp.cpp
```diff
@@ -717,14 +717,14 @@ static int ReadConfigNumericHelper(std::string buf, const char* prefixPtr, T& va
...
-static bool IdentifyMp4ChunkBoundary(AampMediaType type, AampGrowableBuffer *buffer, size_t bufferOffset, size_t &chunkBoundaryOffset, uint64_t &chunkDurationInTicks)
+static bool IdentifyMp4ChunkBoundary(AampMediaType type, std::vector<uint8_t> &buffer, size_t bufferOffset, size_t &chunkBoundaryOffset, uint64_t &chunkDurationInTicks)
 {
 	bool found = false;
 	chunkBoundaryOffset = 0;
 	chunkDurationInTicks = 0;
 
 	IsoBmffBuffer isobmffBuffer;
-	isobmffBuffer.setBuffer(buffer->data() + bufferOffset, buffer->size() - bufferOffset);
+	isobmffBuffer.setBuffer(buffer.data() + bufferOffset, buffer.size() - bufferOffset);
```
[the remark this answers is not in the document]
DomSyna: No, prime example of "Const Overload Proliferation."

pstroffolino:
advanced with newer #1130

PR #1114, opened 2026-03-02
VPLAY-12819: Process crashed with WPEWebProcess while playback

Gnanesha:
Reason for change: check callback is present before even triggering timer
Test Procedure: play vod or live from application
Risks: None


test/utests/tests/InterfacePlayerRDKTests/InterfacePlayerRDKTests.cpp‎
```diff
@@ -0,0 +1,26 @@
+/*
```
mhughesacn, line 1: Please check the filename here too (cf. 1112)

pstroffolino (approved):

PR #1115, opened 2026-03-02
VPLAY-12819: Process crashed with WPEWebProcess while playback

Gnanesha:
Reason for change: check callback is present before even triggering timer
Test Procedure: play vod or live from application
Risks: None


mhughesacn:
Same as PR1112, PR1114

PR #1139, opened 2026-03-05
VPLAY-12921 Remove AAMP Playlist downloader thread ERROR

jfagunde:
Reason for Change: Remove ERROR from log when AAMP playing DASH

Summary of Changes: Remove ERROR log line

Test Procedure: Verify that the following ERROR is not in the log
                "Playlist downloader thread not started"

Risk: Low

### DIFF
--- a/streamabstraction.cpp
+++ b/streamabstraction.cpp
@@ -4265,12 +4265,6 @@ void MediaTrack::AbortWaitForPlaylistDownload()
 	{
 		plDownloadWait.notify_one();
 	}
-	else
-	{
-		// This is expected in some cases like playing DASH content, where the
-		// playlist downloader thread is not used.
-		AAMPLOG_TRACE("[%s] Playlist downloader thread not started", name);
-	}
 }
 
 /**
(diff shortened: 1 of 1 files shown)


streamabstraction.cpp
```diff
@@ -4267,7 +4267,9 @@ void MediaTrack::AbortWaitForPlaylistDownload()
 	}
 	else
 	{
-		AAMPLOG_ERR("[%s] Playlist downloader thread not started", name);
+		// This is expected in some cases like playing DASH content, where the
+		// playlist downloader thread is not used.
+		AAMPLOG_TRACE("[%s] Playlist downloader thread not started", name);
```
DomSyna: Should it be a WARN, if it happens in a non expected case it shouldn't be hidden ?

jfagunde: This is normal behaviour in many cases (playing DASH, playing VOD, thread has not yet started...). There is a WARN printed if starting the thread failed. I think it is wrong to print a WARN here in case the thread should have been running. In other similar functions, like StopPlaylistDownloaderThread(), nothing is printed if the thread has not been started.

DomSyna (approved):

streamabstraction.cpp
jfagunde: Removed log line, as suggested.

pstroffolino (approved):

PR #1146, opened 2026-03-06
[VPLAY-12715](https://ccp.sys.comcast.net/browse/VPLAY-12715) [osx-sp…

rajat698:
…ecific?] libdash ASAN crash when running L2 test 6002 (and some others)

Reason for change: Create inline osx required libdash patch
Risks: Low
Priority: P1


pstroffolino (approved):

pstroffolino (approved):

pstroffolino (changes requested):
this should be taken from meta-rdk-ext/recipes-multimedia/libdash/libdash like the other patches, so that it can be applied across all RDK builds (not just OSX/Ubuntu)

the surgical patch should be renamed with convention like the other patches, i.e.
0011-RDKAAMP-61-AAMP-low-latency-dash-stream-evaluation.patch
instead of generic libdash.patch

pstroffolino:
Using VPLAY-13061 to deliver fix for this ASAN crash.

PR #1169, opened 2026-03-11
VPLAY-12919 [VIPA] Black screen after rewinding VOD asset through ads

anjali-syna:
Reason for change: Rewinding from main content to ad and in the middle of
 rewinding in ad, got a black screen.
 The issue of the black screen appears to be because the calls to
 StopInternal and SetRateInternal get swapped in order,
 due to SetRateInternal being put on the scheduler whereas Stop doesn't.

Changes:
* Add an additional check to only allow the StreamSink to be stopped if there are no inactive aamp players that have a tune performed on them.

Risks: low


main_aamp.cpp
```diff
@@ -366,6 +366,13 @@ void PlayerInstanceAAMP::Tune(const char *mainManifestUrl,
 								)
 {
 	ManageAsyncTuneConfig(mainManifestUrl);
+	
+	// Set tuned flag before scheduling the tune task
+	if (aamp)
+	{
+		AampStreamSinkManager::GetInstance().SetTuned(aamp);
```
anshephe: Add a check that AampStreamSinkManager::GetInstance() isn't null

AampStreamSinkInactive.h
```diff
@@ -357,8 +357,25 @@ class AampStreamSinkInactive : public StreamSink
...
+	 *   @brief Returns the tuned status of this instance
+	 */
+	bool IsTuned() const
+	{
+		return mTuned;
+	}
 
 private:
 	id3_callback_t mId3HandlerCallback;		/**< Returns the id3 callback handle associated with this instance */
+	bool mTuned = false;					/**< Indicates if this instance has been tuned */
```
anshephe: Initialise in the constructor for consistency?

anjali-syna: done

main_aamp.cpp
anjali-syna: The code already has the appropriate null check with if (aamp) before calling SetTuned(). Note that AampStreamSinkManager::GetInstance() returns a reference (not a pointer), so it cannot be null by definition in C++. The singleton pattern ensures an instance is always returned.

AampStreamSinkManager.cpp
```diff
@@ -605,6 +630,22 @@ StreamSink *AampStreamSinkManager::GetStoppingStreamSink(PrivateInstanceAAMP *aa
...
 
+void AampStreamSinkManager::SetTuned(PrivateInstanceAAMP *aamp)
+{
+	std::lock_guard<std::mutex> lock(mStreamSinkMutex);
+
+	auto it = mInactiveGstPlayersMap.find(aamp);
+	if (it != mInactiveGstPlayersMap.end())
+	{
+		it->second->SetTuned(true);
+		AAMPLOG_INFO("AampStreamSinkManager(%p) Set tuned flag for PLAYER[%d]", this, aamp->mPlayerId);
```
jfagunde, line 642: The player ID should be printed in every log message, as part of the header. Does it make sense to also print it here? The same question applies to other added log lines.

AampStreamSinkManager.cpp
```diff
@@ -593,8 +593,33 @@ StreamSink *AampStreamSinkManager::GetStoppingStreamSink(PrivateInstanceAAMP *aa
...
+				if (inactivePlayer.second->IsTuned())
+				{
+					hasTunedInactivePlayer = true;
+				}
+			}
+		}
+
+		if (!hasTunedInactivePlayer)
+		{
+			AAMPLOG_WARN("AampStreamSinkManager(%p) No active player and no tuned inactive players, returning single-pipeline sink for PLAYER[%d]",
```
jfagunde: Should this be a WARN? I would think that this is a normal case. I wonder if both should be INFO (lines 613 and 619).

main_aamp.cpp
jfagunde: This method is only implemented for AampStreamSinkInactive class, is that correct? Should it not have a default implementation or be implemented for AAMPGstPlayer class?

AampStreamSinkManager.cpp
```diff
@@ -605,6 +630,22 @@ StreamSink *AampStreamSinkManager::GetStoppingStreamSink(PrivateInstanceAAMP *aa
...
+	auto it = mInactiveGstPlayersMap.find(aamp);
+	if (it != mInactiveGstPlayersMap.end())
+	{
+		it->second->SetTuned(true);
+		AAMPLOG_INFO("AampStreamSinkManager(%p) Set tuned flag for PLAYER[%d]", this, aamp->mPlayerId);
+	}
+	else
+	{
+		AAMPLOG_WARN("AampStreamSinkManager(%p) Could not find inactive stream sink for PLAYER[%d]", this, aamp->mPlayerId);
+	}
```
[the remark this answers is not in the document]
anjali-syna: AAMPMediaPlayerJS_stop() flow will perform DeleteStreamSink. Player/sink itself won't exist after this. So reset of this variable is not required

AampStreamSinkManager.cpp
anjali-syna: done

AampStreamSinkManager.h
```diff
@@ -149,6 +149,12 @@ class AampStreamSinkManager
 	 * 				  or the stream sink associated with the given player (may be nullptr if couldn't be found)
 	 */
 	virtual StreamSink* GetStoppingStreamSink(PrivateInstanceAAMP *aamp);
```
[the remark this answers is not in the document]
anjali-syna: done

AampStreamSinkManager.cpp
```diff
@@ -593,8 +593,33 @@ StreamSink *AampStreamSinkManager::GetStoppingStreamSink(PrivateInstanceAAMP *aa
...
+		for (const auto& inactivePlayer : mInactiveGstPlayersMap)
+		{
+			if (inactivePlayer.first != aamp)
+			{
+				AAMPLOG_INFO("AampStreamSinkManager(%p) Inactive PLAYER[%d], tuned=%s",
+						this, inactivePlayer.first->mPlayerId, inactivePlayer.second->IsTuned() ? "true" : "false");
+				if (inactivePlayer.second->IsTuned())
+				{
+					hasTunedInactivePlayer = true;
+				}
```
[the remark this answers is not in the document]
anjali-syna: implemented 'break' after finding a tuned inactive player

AampStreamSinkManager.cpp
```diff
@@ -605,6 +631,22 @@ StreamSink *AampStreamSinkManager::GetStoppingStreamSink(PrivateInstanceAAMP *aa
...
+
+	auto it = mInactiveGstPlayersMap.find(aamp);
+	if (it != mInactiveGstPlayersMap.end())
+	{
+		it->second->SetTuned(true);
+		AAMPLOG_INFO("AampStreamSinkManager(%p) Set tuned flag for PLAYER[%d]", this, aamp->mPlayerId);
+	}
+	else
+	{
+		AAMPLOG_WARN("AampStreamSinkManager(%p) Could not find inactive stream sink for PLAYER[%d]", this, aamp->mPlayerId);
```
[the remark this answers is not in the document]
anjali-syna: Inside if also it is printed the same way. This is not an issue

AampStreamSinkManager.cpp
```diff
@@ -593,8 +593,34 @@ StreamSink *AampStreamSinkManager::GetStoppingStreamSink(PrivateInstanceAAMP *aa
...
 	{
-		AAMPLOG_WARN("AampStreamSinkManager(%p) No active player, returning single-pipeline sink for PLAYER[%d]", this, aamp->mPlayerId);
-		sink_ptr = mGstPlayer;
+		// Check if there is any inactive player that has been tuned (excluding the calling aamp)
+		bool hasTunedInactivePlayer = false;
+		for (const auto& inactivePlayer : mInactiveGstPlayersMap)
+		{
+			if (inactivePlayer.first != aamp)
+			{
+				AAMPLOG_INFO("AampStreamSinkManager(%p) Inactive PLAYER[%d], tuned=%s",
```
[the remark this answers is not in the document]
anjali-syna: Info is not enabled by default in builds. Most of the times it is explicitly enabled for debugging purpose.

main_aamp.cpp
```diff
@@ -366,6 +366,13 @@ void PlayerInstanceAAMP::Tune(const char *mainManifestUrl,
 								)
 {
 	ManageAsyncTuneConfig(mainManifestUrl);
+	
+	// Set tuned flag before scheduling the tune task
+	if (aamp)
```
jfagunde: I don't think we need this check. ManageAsyncTuneConfig() would crash if aamp was NULL.

main_aamp.cpp
jfagunde, line 355: As discussed, I would call `UsingPlayerId(aamp->mPlayerId);` at the start of this function, to ensure that any AAMP log line contains the right AAMP instance.

anjali-syna: done

main_aamp.cpp
anjali-syna: done

test/utests/tests/PlayerInstanceAAMP/PlayerInstanceAAMPTestsMain.cpp
```diff
@@ -2774,3 +2774,49 @@ TEST_F(PlayerInstanceAAMPTests, SetRateTest_LocalTSB_TrickPlayWhenPausedFromTSB)
...
+
+	// Mock the necessary calls
+	EXPECT_CALL(*g_mockPrivateInstanceAAMP, GetState()).WillRepeatedly(Return(eSTATE_IDLE));
+
+	// Expect SetTuned to be called before the async task is scheduled
+	EXPECT_CALL(*g_mockAampStreamSinkManager, SetTuned(mPrivateInstanceAAMP)).Times(1);
+
+	// Call Tune in async mode
+	mPlayerInstance->Tune(testUrl, true, contentType, true, false, nullptr, true);
+}
```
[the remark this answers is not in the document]
anjali-syna: Change is common for both syn and async tune path. So the previous test itself is enough. Removed this test

main_aamp.cpp
anjali-syna: Yes it is only implemented for AampStreamSinkInactive class. Not needed for AAMPGstPlayer class .

main_aamp.cpp
```diff
@@ -366,6 +366,13 @@ void PlayerInstanceAAMP::Tune(const char *mainManifestUrl,
...
+	{
+		AampStreamSinkManager::GetInstance().SetTuned(aamp);
+	}
```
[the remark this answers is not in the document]
anjali-syna: This call is needed

AampStreamSinkManager.cpp
anjali-syna: sometimes player id printed with all log lines seems incorrect. So it is better to keep this.

jfagunde (approved):

PR #1177, opened 2026-03-12
VPLAY-12687 Detect EOS if pipeline is paused due to underflow

jfagunde:
Reason for Change: Fix Technical Fault when FF to live

Summary of Changes:
* In CacheFragment() detect EOS if pipeline is paused due to underflow
* Verify changes in L1 tests

Test Procedure: FF to live in underflow conditions

Risk: Low


test/utests/fakes/FakeStreamAbstractionAamp.cpp
```diff
@@ -332,15 +332,25 @@ bool MediaTrack::SignalIfEOSReached()
...
 
 bool MediaTrack::IsLocalTSBInjection()
 {
 	bool localTsbInjection = false;
 
-	// When using mock, delegate to mock implementation
-	if (auto* mock = dynamic_cast<MockMediaTrack*>(this)) {
-		localTsbInjection = mock->IsLocalTSBInjection();
+	MockMediaTrack *mockObj = dynamic_cast<MockMediaTrack*>(this);
+	if (mockObj != nullptr)
```
DomSyna: modern cpp 17 practice is 
if (auto* mockObj = dynamic_cast<MockMediaTrack*>(this))
{
    return mockObj->IsLocalTSBInjection();
}
DRY
this method might be a good candidate for std::variant ?

test/utests/fakes/FakeStreamAbstractionAamp.cpp
```diff
@@ -332,15 +332,25 @@ bool MediaTrack::SignalIfEOSReached()
...
+	{
+		localTsbInjection = mockObj->IsLocalTSBInjection();
+	}
+
+	if (g_mockMediaTrack != nullptr)
```
DomSyna, line 350: modern style for this is
if (g_mockMediaTrack) return g_mockMediaTrack->IsLocalTSBInjection();

The justification is clean and easier to scan.

test/utests/tests/MediaStreamContextTests/FragmentDownloadTests.cpp
```diff
@@ -61,10 +64,25 @@ class FragmentDownloadTests : public testing::Test
 		g_mockStreamAbstractionAAMP = new NiceMock<MockStreamAbstractionAAMP>(mPrivateInstanceAAMP);
 		g_mockPrivateInstanceAAMP = new StrictMock<MockPrivateInstanceAAMP>();
 		g_mockAampTimeBasedBufferManager = new StrictMock<aamp::MockAampTimeBasedBufferManager>();
+		mTsbSessionMgr = std::make_unique<AampTSBSessionManager>(mPrivateInstanceAAMP);
+		mMockTSBSessionMgr = std::make_unique<NiceMock<MockTSBSessionManager>>(mPrivateInstanceAAMP);
+		g_mockTSBSessionManager = mMockTSBSessionMgr.get();
```
DomSyna, line 71: auto& g_mockTSBSessionManager = *mMockTSBSessionMgr;
Is possibly better now we are using cpp 17

test/utests/fakes/FakeStreamAbstractionAamp.cpp
jfagunde: Reverted back closer to what it was, as you suggest.

test/utests/fakes/FakeStreamAbstractionAamp.cpp
jfagunde: I don't agree with that. I still don't like having multiple exit points in a function and I prefer to use braces for if conditions to avoid confusing code like this:
```
if (condition) func_a(); func_b();
```
But most importantly, I think we should follow the AAMP coding guidelines to standardize the code and avoid unnecessary changes to match the style each of us like, to reduce the risk of conflicts.
If you want to change the style, it should be agreed, the guidelines updated and then modify the code, not the other way around.
From the current version of [AAMP Coding Guidelines](https://etwiki.sys.comcast.net/display/RDKV/AAMP+Coding+Guidelines):

- **Implement at most one return statement from any function**. this makes code flow more straightforward, and helps avoid human error with cleanup of locally managed resources.
- **{} are required for blocks  ,** even if it is single line of code within if condition or for loop.

test/utests/tests/MediaStreamContextTests/FragmentDownloadTests.cpp
jfagunde: g_mockTSBSessionManager is defined elsewhere, not here.

But also, I am not very keen on using 'auto' everywhere. It can make it harder to know the type of a variable. According to [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html):
"Use type deduction only if it makes the code clearer to readers who aren't familiar with the project, or if it makes the code safer. Do not use it merely to avoid the inconvenience of writing an explicit type."

test/utests/fakes/FakeStreamAbstractionAamp.cpp
DomSyna: Following the CPP  Core Guidelines as per the instruction files:
if (condition) func_a(); func_b();
is bad and it should be
if (condition) 
{
       func_a(); 
        func_b();
}
But
if (condition) func_a();
is good

test/utests/tests/MediaStreamContextTests/FragmentDownloadTests.cpp
DomSyna: yes its not ideal how we do things :(

DomSyna (approved):
one minor outstanding, its style and taste so leave it to you if you follow up on it

test/utests/fakes/FakeStreamAbstractionAamp.cpp
jfagunde: I'll leave it as is, since it doesn't comply with the **{} are required for blocks** rule.

PR #1193, opened 2026-03-17
VPLAY-13053 Migrate AampGrowableBuffer in MediaTrack to std::vector

DomSyna:
Migrate the AampGrowableBuffer in MediaTrack to std::vector


streamabstraction.cpp
```diff
@@ -1111,20 +1109,16 @@ bool MediaTrack::ProcessFragmentChunk()
...
 	}
 	else
 	{
 		AAMPLOG_TRACE("[%s] Set Unparsed Buffer chunk Empty...", name);
-		unparsedBufferChunk.Free();
-		//memset(&unparsedBufferChunk, 0x00, sizeof(AampGrowableBuffer));
+		aamp_utils::ClearAndRelease(unparsedBufferChunk);
 	}
-	parsedBufferChunk.Free();
+	aamp_utils::ClearAndRelease(parsedBufferChunk);
```
cpc005, line 1121: We passed parsedBufferChunk into InjectFragmentChunkInternal(). Does that take ownership. Cannot se why we need to Free it here??

cpc005 (approved):

streamabstraction.cpp
DomSyna: No it takes a reference not ownership.

If it did take ownership via a move for example that would leave this version in a valid but undefined states so it would still be good to put it in a known state as it is reused.

Once the migration away from aampgrowable buffer is complete I want to take a  moment to examine each of these buffers. We free them regularly, only to be allocated again. I suspect some of their lifesycle could be extended to reduce the number of allocation and free we perform. reading manifests, fragment cache etc sprint to mind

PR #1200, opened 2026-03-18
VPLAY-13060 ABR rejecting outliers for bandwidth calculation L2 2034 intermittent

cpc005:
Reason for change: At start of tune with 2 samples the ABR will reject both if they differ by too much even when both are higher than the required bandwidth.

- Modify ABR to only reject higher than median outliers when it has 2 samples


abr/RollingMedianOutlierEstimator.cpp
```diff
@@ -20,7 +20,7 @@
 
 #include <algorithm>
 #include <sys/time.h>
-
+#include <cstdlib>
 /**
  * @brief Get current time in milliseconds.
```
DomSyna, line 28: Is it called often with 0 or 1 entries so it doesn't actually do anything ?
If so consider something like if (tmpData.empty()) return -1; for early return.

abr/RollingMedianOutlierEstimator.cpp
```diff
@@ -128,15 +128,16 @@ BitsPerSecond RollingMedianOutlierEstimator::UpdateABRBitrateDataBasedOnCacheOut
 	BitsPerSecond ret = -1;
 	BitsPerSecond medianbps = 0;
 
+	size_t initialSize = tmpData.size();
 	std::sort(tmpData.begin(), tmpData.end());
```
DomSyna, line 135: std::nth_element is better than sort in this instance, the cost is  O(n) vs O(n \log n) <- I goggled the complexity difference ;)

abr/RollingMedianOutlierEstimator.cpp
```diff
@@ -146,9 +147,17 @@ BitsPerSecond RollingMedianOutlierEstimator::UpdateABRBitrateDataBasedOnCacheOut
 	for (auto tmpDataIter = tmpData.begin();
```
DomSyna: switching this to tmpData.erase(std::remove_if ... is better. It will reduce the effort from O(n^2) to O(n)

abr/RollingMedianOutlierEstimator.cpp
```diff
@@ -128,15 +128,16 @@ BitsPerSecond RollingMedianOutlierEstimator::UpdateABRBitrateDataBasedOnCacheOut
 	BitsPerSecond ret = -1;
 	BitsPerSecond medianbps = 0;
 
+	size_t initialSize = tmpData.size();
 	std::sort(tmpData.begin(), tmpData.end());
-	if (tmpData.size() % 2)
+	if (initialSize % 2)
 	{
-		medianbps = tmpData.at(tmpData.size() / 2);
+		medianbps = tmpData.at(initialSize / 2);
```
DomSyna: only use .at() if you want bounds checking and exceptions. [] is much better if you know the data is safe to access that way

DomSyna (commented):
Functionally looks fine, though could move forward to some of the cpp 17 features ;)

abr/RollingMedianOutlierEstimator.cpp
cpc005: Not sure how this will work when there is an even number of elements hence we need to calculate:
		BitsPerSecond m1 = tmpData.at(initialSize / 2 - 1);
		BitsPerSecond m2 = tmpData.at(initialSize / 2);
so 2 elements need to be in the correct place

abr/RollingMedianOutlierEstimator.cpp
cpc005: And as noted in the jira, the func gets called multiple times when no samples have changed. The calculated bandwidth could get cached and only re-calculated if there is a change.

abr/RollingMedianOutlierEstimator.cpp
cpc005: I do not see why the current approach is O(n^2) It is not restarting the iteration when something gets erased

abr/RollingMedianOutlierEstimator.cpp
cpc005: Fixed

DomSyna (approved):
Don't mind if you don't want to take onboard any or all of the comments. What you have done works and fixes the issue.

abr/RollingMedianOutlierEstimator.cpp
cpc005: Provided with example, did not incldue

PR #1203, opened 2026-03-19
VPLAY-12908 - Code improvement in TsbReader

hthakkar-synamedia:
Reason for change: - A follow up change to VPLAY-11126. Some edge cases are addressed here.
Test Procedure: - Refer to Ticket
Risks: low


fragmentcollector_mpd.cpp
```diff
@@ -10048,15 +10066,9 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
-					// play cache is full , wait until cache is available to inject next, max wait of 1sec
-					int timeoutMs = MAX_WAIT_TIMEOUT_MS;
 					int trackIdx = (vEOS && !aEOS) ? eMEDIATYPE_AUDIO : eMEDIATYPE_VIDEO;
-					//AAMPLOG_INFO("Cache full state track(%d), no download until(%d) Time(%lld)",trackIdx,timeoutMs,aamp_GetCurrentTimeMS());
-					bool temp =  mMediaStreamContext[trackIdx]->WaitForCachedFragmentChunkInjected(timeoutMs);
-					if(temp == false)
-					{
-						//AAMPLOG_INFO("Waiting for FreeFragmentAvailable");  //CID:82355 - checked return
-					}
+					//// play cache is full , wait until cache is available to inject next, max wait of MAX_WAIT_TIMEOUT_MS
```
jfagunde: Too many '/'?

fragmentcollector_mpd.cpp
```diff
@@ -10048,15 +10066,9 @@ void StreamAbstractionAAMP_MPD::TsbReader()
...
+					mMediaStreamContext[trackIdx]->WaitForCachedFragmentChunkInjected(MAX_WAIT_TIMEOUT_MS);
```
jfagunde: I would void this call, to make it clear that it is intentional to ignore the return value.

test/utests/tests/AampTSBSessionManager/FunctionalTests.cpp
```diff
@@ -653,3 +664,66 @@ TEST_F(FunctionalTests, AdMetadataBoundaryTest)
...
+			.WillOnce(Return(static_cast<int>(TSB::LogLevel::TRACE)));
+		EXPECT_CALL(*g_mockAampTsbMetaDataManager, Initialize()).WillOnce(Return());
+		EXPECT_CALL(*g_mockAampTsbMetaDataManager,
+					RegisterMetaDataType(AampTsbMetaData::Type::AD_RESERVATION_METADATA_TYPE, false))
+			.WillOnce(Return(true));
+		EXPECT_CALL(*g_mockAampTsbMetaDataManager,
+					RegisterMetaDataType(AampTsbMetaData::Type::AD_PLACEMENT_METADATA_TYPE, true))
+			.WillOnce(Return(true));
+		partialMock.Init();
+		std::this_thread::sleep_for(std::chrono::milliseconds(25));
```
jfagunde: Is this sleep needed? If it is, I'd add a comment explaining why.

test/utests/tests/AampTSBSessionManager/FunctionalTests.cpp
```diff
@@ -653,3 +664,66 @@ TEST_F(FunctionalTests, AdMetadataBoundaryTest)
...
+		// Key assertion: the real Flush() must call NotifyVideoTsbWaiters() exactly once.
+		EXPECT_CALL(partialMock, NotifyVideoTsbWaiters()).Times(1);
+		EXPECT_CALL(*g_mockTSBStore, Flush()).Times(1);
+		partialMock.Flush();
+		// ~PartialMockTSBSessionManager() runs here, verifying the EXPECT_CALL above.
+		// ~AampTSBSessionManager() then calls Flush(), but mInitialized_ is false so
+		// NotifyVideoTsbWaiters() is not invoked.
+	}
+
+	// Re-create a plain instance so TearDown can perform its own Flush cleanly.
```
jfagunde: Is this needed? Why do we need TearDown to perform its own Flush cleanly?

AampTSBSessionManager.cpp
jfagunde, line 380: How about adding a TRACE when the cond var is notified?

jfagunde, line 388: How about printing a TRACE just before waiting and another one after wake up?
Previously, when running out of segments in the TSB, we were printing a TRACE every 50ms. This would only print two: one when starting to wait and another one when waking up. It would help identify problems in which the cond var is not notified for whatever reason.

fragmentcollector_mpd.cpp
hthakkar-synamedia: Removed extra

fragmentcollector_mpd.cpp
hthakkar-synamedia: Done

AampTSBSessionManager.cpp
hthakkar-synamedia: Done

hthakkar-synamedia: Done

test/utests/tests/AampTSBSessionManager/FunctionalTests.cpp
hthakkar-synamedia: Init() starts the write thread (mWriteThread) asynchronously. The sleep gives that thread time to reach its blocking wait() call inside ProcessWriteQueue(). Without it, Flush() could run before the thread is actually waiting on the condition variable, causing a race where notify_one() fires before the thread is ready to receive it — potentially causing mWriteThread.join() to hang or the thread to miss the stop signal and not exit cleanly.

test/utests/tests/AampTSBSessionManager/FunctionalTests.cpp
hthakkar-synamedia: Simplified code to remove this

jfagunde (approved):

test/utests/tests/AampTSBSessionManager/FunctionalTests.cpp
```diff
@@ -653,3 +658,64 @@ TEST_F(FunctionalTests, AdMetadataBoundaryTest)
 	EXPECT_TRUE(mAampTSBSessionManager->StartAdPlacement(
 		TEST_AD_ID, TEST_REL_POSITION, zeroPos, 0.0, TEST_OFFSET));
 }
+
+/**
+ * @brief Verify that Flush() calls NotifyVideoTsbWaiters().
+ *
+ * Uses a stack-allocated plain AampTSBSessionManager so the real Flush() and
```
jfagunde: Why? Can we not use mAampTSBSessionManager instead?

jfagunde (approved):

PR #1210, opened 2026-03-20
VPLAY-13077: Crash observed in LLD during trickplay on MediaStreamContext::ABRProfileChanged

srikanthreddybijjam-comcast:
Reason for change: Added proper lock to protect main mpd
Risks:Low
Test Procedure: Refer jira ticket VPLAY-13077
Priority: P0


nu641001:
Review looks good to me. Intention is to protect MPD updates, rather than protecting IndexNewMPDDocument for period iterator updates as we use internal libdash object everywhere in the code

pstroffolino (approved):

PR #1213, opened 2026-03-20
VPLAY-13091 Only set subtitles_muted in TuneHelper() for in-band CC

jfagunde:
Reason for Change: L3 test 2003 failure

Summary of Changes:
* Modified TuneHelper() to only set subtitles_muted for in-band CC if no AAMP API has been called to set CC/subtitles
* Add a flag to keep track if set CC/subtitles AAMP API has been called
* Verified changes in L1

Test Procedure: L3 test 2003

Risks: Medium


cpc005 (approved):

jfagunde:
@copilot review this pull request once and give me comments on any issues you find. Do not modify the pull request.

priv_aamp.cpp
```diff
@@ -11900,6 +11911,7 @@ void PrivateInstanceAAMP::SetCCStatus(bool enabled)
 	AAMPLOG_INFO("enabled %s", enabled?"true":"false");
 	{
 		std::lock_guard<std::recursive_mutex> lock(mStreamLock);
+		mCCStatusSetByApp = true;
```
DomSyna: how does this get cleared to false again ?

DomSyna: won't any internal call that goes through SetCCStatus result in this getting set to true ?

jfagunde: SetCCStatus() is only called by an external API, internal calls call SetCCStatusInternal() directly.

jfagunde: It doesn't. An app will call AAMP APIs to enable/disable CC or talk to CCManager directly, it cannot sometimes do one and other times the other. If this flag is set for an AAMP instance, it will not change after that.

DomSyna: But things like SetTextTrack call SetCCStatus()

DomSyna: StreamAbstractionAAMP_HLS::SelectSubtitleTrack

jfagunde: You are correct. I changed the implementation so the flag is set when the external APIs are called.

jfagunde: I extended the comment of the member variable to clarify this.

priv_aamp.cpp
```diff
@@ -1640,7 +1640,9 @@ PrivateInstanceAAMP::PrivateInstanceAAMP(AampConfig *config) : mReportProgressPo
 	mDownloadsEnabled(true), profiler(), licenceFromManifest(false), previousAudioType(eAUDIO_UNKNOWN),isPreferredDRMConfigured(false),
 	mbDownloadsBlocked(false), streamerIsActive(false), mFogTSBEnabled(false), mIscDVR(false), mLiveOffset(AAMP_LIVE_OFFSET),
 	seek_pos_seconds(-1), rate(0), mSinkPaused(false), mMaxLanguageCount(0), zoom_mode(VIDEO_ZOOM_NONE),
-	video_muted(false), subtitles_muted(true), audio_volume(100), subscribedTags(), manifestHeadersNeeded(), httpHeaderResponses(), timedMetadata(), timedMetadataNew(), IsTuneTypeNew(false), trickStartUTCMS(-1), durationSeconds(0.0), culledSeconds(0.0), culledOffset(0.0), maxRefreshPlaylistIntervalSecs(DEFAULT_INTERVAL_BETWEEN_PLAYLIST_UPDATES_MS/1000),
+	video_muted(false), subtitles_muted(true), audio_volume(100),
```
DomSyna: Since your not changing these lines it might be nice to put them back how they were to remove any distraction

priv_aamp.cpp
```diff
@@ -6291,8 +6294,15 @@ void PrivateInstanceAAMP::TuneHelper(TuneType tuneType, bool seekWhilePaused)
...
 
 		// Retrieve the current closed‑captioning state and log it along with the in‑band CC flag.
-		subtitles_muted = !PlayerCCManager::GetInstance()->GetStatus();
-		AAMPLOG_WARN("SubtitlesMuted:%d isCCinBand:%d", subtitles_muted.load(), mIsInbandCC);
+		previousCCEnabled = PlayerCCManager::GetInstance()->GetStatus();
+		if (mIsInbandCC && !mCCStatusSetByApp.load())
+		{
+			// No API to set CC or subtitles has been called, so AAMP assumes that the app talks to CCManager directly.
+			// Set subtitles_muted based on the current CC status to ensure correct subtitle state is applied later on.
+			subtitles_muted = !previousCCEnabled;
```
DomSyna, line 6350: Will this result in a potential change in behaviour for devices that were working before the issue was found ?
Previously subtitles_muted would have been set based on GetStatus() regardlesss of mIsInbandCC. Now it only gets set if that's set ?

Could it be false when it gets here a a result of SelectSubtitleTrack not finding a specified track, and using a default ?

main_aamp.cpp
```diff
@@ -2999,6 +3000,8 @@ int PlayerInstanceAAMP::GetTextTrack()
  */
 void PlayerInstanceAAMP::SetCCStatus(bool enabled)
 {
+	AAMPLOG_MIL("enabled %s", enabled?"true":"false");
```
DomSyna, line 2994: This can be called by the JS bindings, is that an issue ?

DomSyna (commented):
Seems OK, a little nervous though as I don't know how we test scenarios we haven't thought about

main_aamp.cpp
jfagunde: The app calls JS bindings, which calls this API. This is fine.

priv_aamp.cpp
jfagunde: This is the only behavioral change compared to the code pre-VPLAY-12917 (https://github.com/rdkcentral/aamp/pull/1166). The variable subtitles_muted is set if SetCCStatus() or SetSubtitleMute() are called (as it was before). Now, in addition, it is set here for apps that use CC (mIsInbandCC = true) and don't call the APIs above (mCCStatusSetByApp = false). For those apps, the behaviour should be the same as post-VPLAY-12917.

priv_aamp.cpp

PR #1226, opened 2026-03-23
VPLAY-13104 l2 fixes for tip of dev_sprint_25_2

pstroffolino:
VPLAY-13104 l2 fixes for tip of dev_sprint_25_2

Reason for Change: Several L2 tests have begun consistently failing in recent days, including 1019, 2032, and 6002.
2032 failure is caused by change in log level (info vs. trace)
6002 is failing because a configuration value that controls underflow detection was set to 0, effectively disabling it
1019 - there was a recent change in how the test framework handles simulated server responses

Test Guidance: L2 1019, 2032, 6002 passing again

Risk: Medium


pstroffolino:
abandoning

PR #1247, opened 2026-03-26
VPLAY-13112: CC fails to display on linear channel change, and

psiva01:
appears only after D-pad right key press


psiva01:
Fixed by https://github.com/rdkcentral/aamp/pull/1213

PR #1252, opened 2026-03-27
VPLAY-13176 underflow detection refactoring

pstroffolino:
Reason for Change: motivated by struggles getting L2 6002 to pass on OSX
Refactoring to eliminate polling loop, vendor specific code, dependencies on streamer-level signaling.

Test Guidance: underflow monitoring/rerporting must work well across devices and simulator variations.  All underflow-specific tests must work.

Risk: High - simplification, but big change.  That said, current implementation has some known issues.


Vinish100 (commented):
Looks good to me 👍 
L1 is failing as fakes are not updated

PR #1262, opened 2026-03-31
VPLAY-13038

nebutler963:


test/utests/drm/mocks/MockPrivateInstanceAAMP.h
```diff
@@ -17,21 +17,5 @@
  * limitations under the License.
  */
 
-#ifndef AAMP_MOCK_AAMP_PRIV_AAMP_H
```
p-bond, line 20: If the file's no longer needed (per deleting the content of it), why not just delete it and force inclusion of the mock header?

p-bond (approved):
Optional if you want to address the comments.  Copilot is wrong on the C++17 issue.

PR #1267, opened 2026-04-01
VPLAY-12490 Use CPP 17 to build AAMP (#917)

hthakkar-synamedia:
VPLAY-12490 Use CPP 17 to build AAMP
Reason for Change: globally migrate AAMP code based from C++11 and C++14 to C++17 Test Guidance: general regression
Risk: Low
---------


DomSyna (approved):
Looks sane

PR #1274, opened 2026-04-02
VPLAY-12919 [VIPA] Black screen after rewinding VOD asset through ads…

paulsm2-syna:
… (#1169)

* VPLAY-12919 [VIPA] Black screen after rewinding VOD asset through ads

Reason for change: Rewinding from main content to ad and in the middle of
 rewinding in ad, got a black screen.
 The issue of the black screen appears to be because the calls to
 StopInternal and SetRateInternal get swapped in order,
 due to SetRateInternal being put on the scheduler whereas Stop doesn't.

Changes:
* Add an additional check to only allow the StreamSink to be stopped if there are no inactive aamp players that have a tune performed on them.

Risks: low

---------

### DIFF
--- a/AampStreamSinkInactive.h
+++ b/AampStreamSinkInactive.h
@@ -30,7 +30,7 @@ class AampStreamSinkInactive : public StreamSink
 {
 
 public:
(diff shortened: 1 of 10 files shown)


DomSyna (approved):
Looks good to me

PR #1277, opened 2026-04-07
VPAAMP-22 [VIPA] AAMP returns duplicate CC text tracks on intra-asset encrypted assets

cpc005:
Reason for change: Multiple adaption sets each has same CC lang. Results in same CC lang being listed in EPG multiple times.

- Only add CC track to vector if it has not already been added.
- Add L1 test using manifest where problem initially seen.


cpc005:
@copilot Please review this code change.

jfagunde (approved):

PR #1284, opened 2026-04-07
VPAAMP-42: Fix ad break fragment scheduling for live stream

nu641001:
Reason for change: Fix ad break fragment scheduling for live stream if split period ad is placed and in progress.
Risks: Low
Test Procedure: Test with Early return stream
Priority: P1

Signed-off-by: Nandakishor U M <nu641001@gmail.com>


Vinish100 (approved):
Looks good to me 👍

PR #1294, opened 2026-04-09
VPAAMP-83 l2 8015 test failures (regression)

pstroffolino:
Reason for Change: fix race condition causing frequent L2 test 8015 failures, epecially on ubuntu gitrunners

Test Guidance: L2 8015 passing reliably again

Risk: Low


Vinish100 (approved):
Looks good to me 👍

pstroffolino:
No longer needed - alternate test-only PR in same review is proper fix.

PR #1302, opened 2026-04-13
Fixes potential race in teardown/restart flow observed under trickplay

p-bond:
Fix race condition in ProfileEventAAMP::TuneBegin and move mLldLowBuffObject null assignment before cJSON_Delete in both TuneBegin and GetTelemetryParam to prevent use-after-free (addresses double free in ticket)

Add discontinuityParamMutex lock_guard in TuneBegin to protect telemetryParam and mLldLowBuffObject from concurrent access

Move mLldLowBuffObject = NULL to before cJSON_Delete(telemetryParam) in both methods, ensuring the pointer is cleared before the underlying cJSON node it points to is freed


AampProfiler.cpp
```diff
@@ -216,13 +216,14 @@ void ProfileEventAAMP::TuneBegin(void)
...
 	bitrateChange = 0;
 	bufferChange = 0;
-	if(telemetryParam != NULL)
+	cJSON *newParam = cJSON_CreateObject();
 	{
+		std::lock_guard<std::mutex> lock(discontinuityParamMutex);
+		mLldLowBuffObject = NULL;
 		cJSON_Delete(telemetryParam);
+		// mLldLowBuffObject is a child of telemetryParam, so it's automatically deleted above
+		telemetryParam = newParam;
```
cpc005: Why not just 
telemetryParam = cJSON_CreateObject();
No reason to be different to GetTelemetryParam()

AampProfiler.cpp
```diff
@@ -216,13 +216,14 @@ void ProfileEventAAMP::TuneBegin(void)
 	rateCorrection = 0;
 	bitrateChange = 0;
 	bufferChange = 0;
-	if(telemetryParam != NULL)
+	cJSON *newParam = cJSON_CreateObject();
 	{
+		std::lock_guard<std::mutex> lock(discontinuityParamMutex);
+		mLldLowBuffObject = NULL;
 		cJSON_Delete(telemetryParam);
```
jfagunde: Could it be a problem calling cJSON_Delete() on telemetryParam if it is NULL? Could it cause a crash?
Should this be tested in L1?

p-bond: cJSON_Delete() handles NULL safely — it's a no-op. The original NULL guard was redundant. No L1 test needed; this is cJSON's documented API contract, not our logic.

AampProfiler.cpp
p-bond: Moving allocation outside the lock to minimise critical section, although the saving is minimal.

AampProfiler.cpp
```diff
@@ -216,13 +216,14 @@ void ProfileEventAAMP::TuneBegin(void)
 	rateCorrection = 0;
 	bitrateChange = 0;
 	bufferChange = 0;
-	if(telemetryParam != NULL)
+	cJSON *newParam = cJSON_CreateObject();
```
jfagunde, line 219: This code used to be very similar to  ProfileEventAAMP::GetTelemetryParam(), why make it different?
How about calling  ProfileEventAAMP::GetTelemetryParam() from here? The only difference is that it prints a MIL, and maybe that is a good thing?

p-bond: They have different intent. GetTelemetryParam() logs-then-resets at a reporting boundary. TuneBegin() discards-and-resets because the previous tune's telemetry is stale. Calling GetTelemetryParam() here would emit a spurious MIL log of old data on every tune start, which would pollute log analysis.

AampProfiler.cpp
```diff
@@ -216,13 +216,14 @@ void ProfileEventAAMP::TuneBegin(void)
...
 	{
+		std::lock_guard<std::mutex> lock(discontinuityParamMutex);
+		mLldLowBuffObject = NULL;
 		cJSON_Delete(telemetryParam);
+		// mLldLowBuffObject is a child of telemetryParam, so it's automatically deleted above
```
jfagunde: Is this comment still relevant? and the same in line 639

AampProfiler.cpp
jfagunde: Ok. Regardless of the MIL, could this piece of code in both functions be standardized? With this change...
TuneBegin()
```
cJSON *newParam = cJSON_CreateObject();
{
	std::lock_guard<std::mutex> lock(discontinuityParamMutex);
	mLldLowBuffObject = NULL;
	cJSON_Delete(telemetryParam);
	telemetryParam = newParam;
}
```
GetTelemetryParam()
```
std::lock_guard<std::mutex> lock(discontinuityParamMutex);
if(telemetryParam != NULL)
{
	mLldLowBuffObject = NULL;
	cJSON_Delete(telemetryParam);
	telemetryParam = cJSON_CreateObject();
}
```

p-bond: Already explained: different intent requires different structure. TuneBegin unconditionally replaces and minimises lock scope; GetTelemetryParam conditionally logs-then-resets. Forcing identical shape would compromise one or the other.

AampProfiler.cpp
p-bond: Yes. mLldLowBuffObject is a child of the telemetryParam cJSON tree — the comment documents why we NULL it before the delete. Same applies at line 639.

nebutler963:
Looks Good

nebutler963 (approved):

p-bond:
The code changes have gone through Copilot as a minimal code change.
The suggested review comments introduce a lot more potential for issues.

PR #1316, opened 2026-04-15
VPAAMP-100: fix stale headers when running install-aamp.sh

p-bond:
Minor changes to avoid stale header files in aamp repo - they were only being copied in if --player-interface-source=external
Less of a sledgehammer than -t, faster and doesn't force a full rebuild


scripts/install_middleware_interfaces.sh
```diff
@@ -27,6 +27,47 @@
...
+# found first on the include path and causes build errors.
+# This function compares every header already present in .libs/include against
+# its counterpart in middleware/ and overwrites any that differ.
+function sync_internal_middleware_headers_fn()
+{
+    local include_dir="${LOCAL_DEPS_BUILD_DIR}/include"
+    local middleware_dir="${AAMP_DIR}/middleware"
+    local updated=0
+
+    if [[ ! -d "${include_dir}" ]] || [[ ! -d "${middleware_dir}" ]]; then
```
cpc005, line 43: Perhaps you can do all of this with rsync

p-bond: Good thought — I considered rsync but the directory structures don't match. include is flat (middleware headers sit at the root) while middleware is nested (closedcaptions/subtec/, playerJsonObject/, vendor/brcm/, etc.). Since rsync matches by relative path, not basename, a simple rsync --existing wouldn't find the counterparts. We'd need either a temp staging dir or a per-file loop, which ends up no simpler than the current find + diff -q + cp approach. Happy to revisit if you see an angle I'm missing.

nebutler963:
Looks good

nebutler963 (approved):

pstroffolino (approved):

PR #1341, opened 2026-04-22
VPAAMP-157 Spurious PlaybackSpeedChanged event

nebutler963:
When rewinding Linear or Vod spurious speed (x1) events were being generated by AAMP.


p-bond (approved):
Looks good to me.

PR #1350, opened 2026-04-23
Replace CachedFragment::Copy with std::move in CacheTsbFragment

benmesander:
Eliminate unnecessary deep copy of fragment data (2-7 MB for UHD) in CacheTsbFragment by using move assignment instead of Copy(). The move assignment operator already exists on CachedFragment.

Also add null guard for GetFetchChunkBuffer return value to prevent potential null pointer dereference. (bug discovered while doing above)

LL-DASH chunk mode callers that need the fragment for both CacheTsbFragment and EnqueueWrite now create a separate copy before moving into the cache.

Add unit test validating move semantics transfer ownership correctly.

### DIFF
--- a/MediaStreamContext.cpp
+++ b/MediaStreamContext.cpp
@@ -628,8 +628,8 @@ bool MediaStreamContext::CacheTsbFragment(std::shared_ptr<CachedFragment> fragme
 		}
 		if(!cachedFragment->fragment.empty())
 		{
-			// If following log is coming, possible memory leak. Need to clear the data first before slot reuse.
-			AAMPLOG_WARN("Fetch buffer has junk data, Need to free this up");
+			// Slot was not cleared after previous use; the assignment below will overwrite and release the old data.
+			AAMPLOG_WARN("Fetch buffer has junk data; previous slot was not cleared after use");
 		}
 		*cachedFragment = std::move(*fragment);
 		if(!cachedFragment->fragment.empty())
--- a/MediaStreamContext.cpp
+++ b/MediaStreamContext.cpp
@@ -612,7 +612,7 @@ int MediaStreamContext::GetDefaultDurationBetweenPlaylistUpdates()
  * @param fragment TSB fragment pointer
  * @retval true on success
  */
-bool MediaStreamContext::CacheTsbFragment(std::shared_ptr<CachedFragment> fragment)
+bool MediaStreamContext::CacheTsbFragment(std::shared_ptr<CachedFragment>&& fragment)
 {
 	// FN_TRACE_F_MPD( __FUNCTION__ );
 	std::lock_guard<std::mutex> lock(fetchChunkBufferMutex);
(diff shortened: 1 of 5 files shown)


MediaStreamContext.cpp
```diff
@@ -730,7 +735,8 @@ void MediaStreamContext::OnFragmentDownloadSuccess(DownloadInfoPtr dlInfo)
 			// If reader is at EOS, inject the last data in AAMP TSB
 			if (aamp->GetLLDashChunkMode())
 			{
-				CacheTsbFragment(fragmentToTsbSessionMgr);
+				auto fragmentForChunkCache = std::make_shared<CachedFragment>(*fragmentToTsbSessionMgr);
```
DomSyna, line 738: Not sure I see the point here ? Isn't this still doing the deep copy in the ctor constructor

MediaStreamContext.cpp
```diff
@@ -741,7 +747,8 @@ void MediaStreamContext::OnFragmentDownloadSuccess(DownloadInfoPtr dlInfo)
 			// In chunk mode, media segments are added to the chunk cache in the SSL callback, but init segments are added here
 			if (aamp->GetLLDashChunkMode())
 			{
-				CacheTsbFragment(fragmentToTsbSessionMgr);
+				auto fragmentForChunkCache = std::make_shared<CachedFragment>(*fragmentToTsbSessionMgr);
```
DomSyna, line 750: same as above

MediaStreamContext.cpp
DomSyna: maybe consider restructuring to make the chunk cache and the TSB writer share an immutable payload (shared pointer to const uint8_t or const Cachedfragment) ?

test/utests/tests/MediaStreamContextTests/FragmentDownloadTests.cpp
```diff
@@ -759,6 +759,52 @@ TEST_F(FragmentDownloadTests, OnFragmentDownloadSuccess_CheckEos_PausedDueToUnde
...
+// ---------------------------------------------------------------------------
+
+/**
+ * @brief Verify that CacheTsbFragment uses move semantics to transfer
+ *        fragment data into the ring buffer slot, avoiding a deep copy.
+ *
+ * After the call the destination (ring buffer slot) must own the data
+ * and the source fragment must be left in a moved-from (empty) state.
+ */
+TEST_F(FragmentDownloadTests, CacheTsbFragment_MoveSemantics_TransfersOwnership)
```
DomSyna, line 778: I don't think this test is testing what you want to test. You are testing that that ringBufferSlot holds what you expect and not that move happened.

If you remove the move from the call to CacheTsbFragment() then the move inside that method will clear the contents of sourceFragment so at the end of the test you can validate by checking params like
    EXPECT_TRUE(sourceFragment->fragment.empty());
    EXPECT_DOUBLE_EQ(sourceFragment->duration, 0.0);
...

Though if someone changes the method to move the sourceFragment into a temporary var, and then does a copy the test will still pas as a move happened on the test data. Hope that makes some sense ;)

MediaStreamContext.cpp
DomSyna: ok, this wasn't to save the deep copy it was to preserve the cached fragment from the move inside cacheTsbFragment

MediaStreamContext.cpp
DomSyna: Since everything now does a move to call this, and internally in now consumes the buffer would it make sense to change the interface to 
bool CacheTsbFragment(std::shared_ptr<CachedFragment>&& fragment);
to enforce it is called with move ?

test/utests/tests/MediaStreamContextTests/FragmentDownloadTests.cpp
pstroffolino: working on this...

MediaStreamContext.cpp
pstroffolino: good idea - I am applying this recommendation

PR #1363, opened 2026-04-27
VPLAY-13208 [VIPA] [Linear]- Progress Bar Point jumping to the End Wh…

anjali-syna:
…ile performing Trick Play/Seek

Reason for change: Progress Bar Point jumping to the end and playing
 from live play position while performing Trick Play/Seek in Linear channels.
 This issue is seen only in one platform. In the problematic scenario,
 aamp is receiving many underflows callbacks - GstPlayer_OnGstBufferUnderflowCb,
 from rialtomsevideosink0, due to which EOS flow is triggered from
 VideoDecoderPtsCheckerForEOS, as 'video-pts' property is not supported in
 RialtoMSEVideoSink.

Changes:
* Modifying aamp's underflow EOS implementation to fallback to gst_element_query_position, if 'video-pts' property is not supported.Both VideoDecoderPtsCheckerForEOS and GstPlayer_OnGstBufferUnderflowCb compare real playback positions in the absence of video-pts property, so a mid-underflow position change prevents NotifyEOS() from firing. Also, modified the code to allow pts check for EOS only during normal play, and not for forward trick play.

Risks: low


middleware/vendor/realtek/RealtekSocInterface.cpp
```diff
@@ -242,6 +242,13 @@ long long RealtekSocInterface::GetVideoPts(GstElement *video_sink, GstElement *v
 
 	if (element)
 	{
+		if(!IsVideoPtsPropertySupported(element))
+		{
+			/* The 'video-pts' property is not exposed on this platform.
+			 * Signal to the caller so it can fall back to
+			 * gst_element_query_position. */
+			return -1;
```
anshephe: Can we avoid the early return, same with SocInterface

middleware/vendor/SocInterface.cpp
```diff
@@ -229,6 +229,13 @@ long long SocInterface::GetVideoPts(GstElement *video_sink, GstElement *video_de
 	gint64 currentPTS = 0;
 	if(video_dec)
 	{
+		if(!IsVideoPtsPropertySupported(video_dec))
+		{
+			/* The 'video-pts' property is not exposed on this platform.
+			 * Signal to the caller so it can fall back to
```
anshephe: I'd remove this bit of the comment, as it assumes the caller's intention

middleware/vendor/SocInterface.cpp
DomSyna: Did you consider std::optional for the method instead or returning -1 ? it would avoid the need of the magic -1

Might be too bigger change in the various call sites

middleware/vendor/SocInterface.h
```diff
@@ -90,6 +90,13 @@ class SocInterface
 	
 	/*config to indicate platforms using westeros sink*/
 	bool mUsingWesterosSink = false;
+
+	/* Cached result of the one-time 'video-pts' GObject property probe.
+	 * Once probed against a non-NULL video sink/decoder, the result is
+	 * reused for the lifetime of this SocInterface instance to avoid
+	 * repeated g_object_class_find_property() calls on hot paths. */
+	bool mVideoPtsPropertyChecked = false;
```
DomSyna: Do these need to be protected for thread safety ? Is it possible that the GST signal handling thread, and the AAMP player thread could access the method at same time ?

maybe:
std::once_flag mVideoPtsPropertyChecked 
std::atomic<bool> mVideoPtsPropertySupported

middleware/vendor/SocInterface.cpp
```diff
@@ -238,6 +245,38 @@ long long SocInterface::GetVideoPts(GstElement *video_sink, GstElement *video_de
...
+ * @brief Check whether the 'video-pts' GObject property is supported by
+ *        the supplied GStreamer element.
+ *
+ * @param element The GStreamer element to probe (may be NULL).
+ *
+ * @return true if 'video-pts' is exposed on the element, false otherwise.
+ */
+bool SocInterface::IsVideoPtsPropertySupported(GstElement *element)
+{
+	if(!mVideoPtsPropertyChecked)
```
DomSyna: with std::once_flag you could remove this check, and inside the scope of element 
std::call_once(mVideoPtsPropertyChecked, lambda fn to find the property, set the supported flag, and log

middleware/vendor/SocInterface.cpp
```diff
@@ -229,6 +229,13 @@ long long SocInterface::GetVideoPts(GstElement *video_sink, GstElement *video_de
 	gint64 currentPTS = 0;
 	if(video_dec)
 	{
+		if(!IsVideoPtsPropertySupported(video_dec))
```
DomSyna: This sequence is duplicated in both cpp files
If(!IsVideo..
g_object_get()

Is it worth moving into a protected helper in SicInterface.h
std::optional<gint64> ReadVideoPts(GstElement *element); // null when no PTS

middleware/InterfacePlayerRDK.cpp
```diff
@@ -3961,9 +3968,65 @@ long long InterfacePlayerRDK::GetVideoPTS(void)
 {
 	gint64 currentPTS = 0;
 	currentPTS = interfacePlayerPriv->socInterface->GetVideoPts(interfacePlayerPriv->gstPrivateContext->video_sink, interfacePlayerPriv->gstPrivateContext->video_dec, interfacePlayerPriv->gstPrivateContext->using_westerossink);
+	if(currentPTS == -1)
+	{
+		/* The 'video-pts' property is not supported on this platform.
+		 * Fall back to gst_element_query_position and convert ms to 90 kHz
+		 * ticks (1 ms = 90 ticks) so that callers relying on PTS units
+		 * (90 kHz) keep working. */
+		MW_LOG_INFO("InterfacePlayerRDK - 'video-pts' property is not supported on this platform; Fall back to gst_element_query_position");
```
DomSyna: Could this flood the logs on a platform where video-pts isn't supported ?

middleware/InterfacePlayerRDK.cpp
```diff
@@ -3961,9 +3968,65 @@ long long InterfacePlayerRDK::GetVideoPTS(void)
...
+		currentPTS = MPEG_PTS_TICKS_PER_MS * GetVideoPosition();
+	}
 	return (long long)currentPTS;
```
DomSyna: cpp not C
static_cast<long long>

middleware/InterfacePlayerRDK.cpp
```diff
@@ -3961,9 +3968,65 @@ long long InterfacePlayerRDK::GetVideoPTS(void)
...
+		}
+	}
+	else
+	{
+		MW_LOG_WARN("InterfacePlayerRDK: cannot query video position; pipeline=%p videoSinkbin=%p",
+					pipeline, videoSinkbin);
+	}
+
+	MW_LOG_INFO("InterfacePlayerRDK: position(ms) = %" G_GINT64_FORMAT , GST_TIME_AS_MSECONDS(position));
+	return (long long)(GST_TIME_AS_MSECONDS(position));
```
DomSyna: as above cpp not C

middleware/InterfacePlayerRDK.cpp
```diff
@@ -66,6 +66,13 @@ static const char* GstPluginNameVMX = "verimatrixdecryptor";
 #include <assert.h>
 #define GST_NORMAL_PLAY_RATE		1
 
+/**
+ * @brief PTS ticks per millisecond on the MPEG / 'video-pts' 90 kHz clock.
+ *        Used to convert a millisecond playback position into 90 kHz ticks
+ *        when the 'video-pts' property is not exposed by the platform.
+ */
+static constexpr long long MPEG_PTS_TICKS_PER_MS = 90LL;
```
DomSyna: I'd be surprised if there isn't already a define for this somewhere ?
Would 
inline constexpr 
be better ?

middleware/vendor/SocInterface.h
```diff
@@ -375,6 +382,20 @@ class SocInterface
...
+	 *
+	 * The result is cached for the lifetime of this SocInterface instance,
+	 * so the underlying g_object_class_find_property() call runs at most
+	 * once.
+	 *
+	 * @param element The GStreamer element to probe (may be NULL).
+	 *
+	 * @return true if 'video-pts' is exposed on the element, false otherwise.
+	 */
+	virtual bool IsVideoPtsPropertySupported(GstElement *element);
```
DomSyna: Should this alsobe marked as nodiscard ?

middleware/test/utests/fakes/FakeSocInterface.cpp
```diff
@@ -188,13 +188,31 @@ void SocInterface::SetWesterosSinkState(bool status)
 {
 	mUsingWesterosSink = true;
 }
+bool SocInterface::IsVideoPtsPropertySupported(GstElement *element)
+{
+	if(!mVideoPtsPropertyChecked)
```
jfagunde: Do we need all this logic in the fake? The same comment applies to SocInterface::GetVideoPts() fake. Ideally the fakes would contain no logic at all, although there is already logic in the existing ones :(

middleware/InterfacePlayerRDK.cpp
anjali-syna: existing code; modified now.

middleware/InterfacePlayerRDK.cpp
anjali-syna: modified

middleware/vendor/realtek/RealtekSocInterface.cpp
anjali-syna: done

middleware/vendor/SocInterface.cpp
anjali-syna: done

middleware/vendor/SocInterface.h
anjali-syna: done

middleware/InterfacePlayerRDK.cpp
anjali-syna: Modified it as TRACE.

middleware/InterfacePlayerRDK.cpp
anjali-syna: There is no define as per ai. moved it to middleware/InterfacePlayerRDK.h as inline constexpr

middleware/test/utests/fakes/FakeSocInterface.cpp
anjali-syna: Leaving it as we discussed.

jfagunde: This is following the current implementation of all other fakes in this file. They implement simple logic instead of calling mocks. Let's leave this as it is.

middleware/vendor/SocInterface.h
anjali-syna: as discussed, kept 'mVideoPtsPropertySupported; as it is and changed mVideoPtsPropertyChecked to std::once_flag and renamed it to mVideoPtsProbeOnce

middleware/vendor/SocInterface.cpp
anjali-syna: done

middleware/vendor/SocInterface.cpp
anjali-syna: done

middleware/vendor/SocInterface.cpp
anjali-syna: I tried this, but llama compilation is failing with std::optional. So reverted the change.

anshephe (approved):

dnfaulkner (commented):
Looks good to me

test/utests/fakes/FakeSocInterface.cpp
```diff
@@ -165,21 +165,34 @@ void SocInterface::SetDecodeError(GstObject* src)
         g_object_set(src, "report_decode_errors", TRUE, NULL);
 }
 
-long long SocInterface::GetVideoPts(GstElement *video_sink, GstElement *video_dec, bool isWesteros)
+long long SocInterface::ReadVideoPts(GstElement *element)
```
Vinish100: Shouldn't fakes be empty and call mocks if any values needs to be retured

middleware/test/utests/fakes/FakeSocInterface.cpp
```diff
@@ -188,20 +188,49 @@ void SocInterface::SetWesterosSinkState(bool status)
 {
 	mUsingWesterosSink = true;
 }
+void SocInterface::CheckVideoPtsPropertySupport(GstElement *element)
+{
+	if (element)
```
Vinish100: Shouldn't fakes be empty?

middleware/InterfacePlayerRDK.cpp
```diff
@@ -3966,9 +3966,65 @@ bool InterfacePlayerRDK::CreatePipeline(const char *pipelineName, int PipelinePr
  */
 long long InterfacePlayerRDK::GetVideoPTS(void)
 {
-	gint64 currentPTS = 0;
+	long long currentPTS = 0;
 	currentPTS = interfacePlayerPriv->socInterface->GetVideoPts(interfacePlayerPriv->gstPrivateContext->video_sink, interfacePlayerPriv->gstPrivateContext->video_dec, interfacePlayerPriv->gstPrivateContext->using_westerossink);
-	return (long long)currentPTS;
+	if (currentPTS == -1)
+	{
+		/* The 'video-pts' property is not supported on this platform.
```
pstroffolino, line 3973: if fallback works well, why even bother first attempting to use video-pts property, which may or may not even work?

would strictly using the fallback code be safe and simpler?

test/utests/fakes/FakeSocInterface.cpp
anjali-syna: done

middleware/test/utests/fakes/FakeSocInterface.cpp
anjali-syna: done

test/utests/fakes/FakeSocInterface.cpp
```diff
@@ -165,21 +165,34 @@ void SocInterface::SetDecodeError(GstObject* src)
...
 {
-        gint64 currentPTS = 0;
-        GstElement *element;
+	long long result = 0;
+	if (element)
+	{
```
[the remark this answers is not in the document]
anjali-syna: done

middleware/InterfacePlayerRDK.cpp
```diff
@@ -3966,9 +3966,65 @@ bool InterfacePlayerRDK::CreatePipeline(const char *pipelineName, int PipelinePr
...
+		}
+	}
+	else
+	{
+		MW_LOG_WARN("InterfacePlayerRDK: cannot query video position; pipeline=%p videoSinkbin=%p",
+					pipeline, videoSinkbin);
+	}
+
+	MW_LOG_INFO("InterfacePlayerRDK: position(ms) = %" G_GINT64_FORMAT , GST_TIME_AS_MSECONDS(position));
+	return static_cast<long long>( (GST_TIME_AS_MSECONDS(position)) );
```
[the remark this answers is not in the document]
anjali-syna: done

middleware/InterfacePlayerRDK.cpp
anjali-syna: As @dnfaulkner  mentioned in slack, We don't know the full history as to why the video pts was originally used and thought that it was lower risk to solve just the specific use case and problem with Rialto rather than change the way all other use cases/platforms currently work. If you're happy that using the position instead of video-pts will be OK then we can remove the use of / dependency on video-pts.
The plan is that Rialto will eventually support the video-pts property. In which case, the fall-back will not be used until another platform comes along that doesn't support vide-pts.
For Direct Rialto, the assumption is that when it adds support for this property then there will be an IMediaPipeline API method to obtain this property.

PR #1366, opened 2026-04-27
Fix getBestMatchedProfileIndexByBandWidth sorted-order assumption

benmesander:
The old implementation iterated mProfiles linearly, assuming profiles were sorted by ascending bandwidth. Because mProfiles is populated in manifest order (not necessarily sorted), the closest-match logic could return an incorrect profile index.

This assumption is largely correct for a MediaKind transcoder, CLO/Pillar linear segmenter, super8 packager origin stack, however the DASH specification does not require profiles to be in sorted order so this is a fragile assumption, and there may be cases where Super8 reorders things in the manifest that could cause subtle bugs here.

Replace the linear scan with a lookup into mSortedBWProfileList (a std::map keyed by bandwidth, already maintained in sorted order). Use std::map::lower_bound to find the closest bandwidth match, checking both the lower_bound entry and its predecessor to pick the nearest by absolute difference.

Applied to both ABR implementations:
- abr/abr.cpp (new ABR)
- support/aampabr/ABRManager.cpp (legacy ABR)

Added #include <climits> for LONG_MAX in both files.

L1 Tests added:
- GetBestMatchedProfile_ExactMatch
- GetBestMatchedProfile_ClosestMatch_UnsortedProfiles
- GetBestMatchedProfile_SkipsIframeTracks (new ABR only)
- GetBestMatchedProfile_ExactMatch_UnsortedProfiles (legacy)
- GetBestMatchedProfile_ClosestMatch_UnsortedProfiles (legacy)

### DIFF
--- a/abr/abr.cpp
+++ b/abr/abr.cpp
@@ -349,7 +349,7 @@ void ABRManager::updateProfile()
  */
 int ABRManager::getBestMatchedProfileIndexByBandWidth(int bandwidth)
 {
-	int desiredProfileIndex = 0;
+	int desiredProfileIndex = INVALID_PROFILE;
 	std::lock_guard<std::mutex> lock(mProfileLock);
 	long bestDiff = LONG_MAX;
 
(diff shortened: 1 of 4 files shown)


abr/abr.cpp
```diff
@@ -349,31 +350,50 @@ void ABRManager::updateProfile()
...
-				// fragment file name bandwidth doesn't match the profile bandwidth, will be always less
-				if (static_cast<size_t>(i + 1) == profileCount) {
-					desiredProfileIndex = i;
-					break;
-				}
-				else
-					desiredProfileIndex = (i + 1);
+	long bestDiff = LONG_MAX;
+
+	for (const auto& periodEntry : mSortedBWProfileList)
```
Vinish100: Should we follow the same as other API's like getClosestProfileIndexByBandwidth() and use the first period map rather than checking through all the period maps?

Expectation is that,  there will be only one single period map at a time, should we log it if there are multiple period maps for discovering any edge cases?

PR #1368, opened 2026-04-27
VPAAMP-207: Fix Ethan log double newline in VIPA logging

shripadbpersonal:
Reason for change:
This change resolves blank lines seen in VIPA logs when Ethan logging is enabled. The log formatter was adding a trailing newline, and Ethan logging was also applying line termination, resulting in double spacing between entries. The fix strips the trailing newline only on the Ethan redirection path, while keeping CLI and journal logging behavior unchanged.

Test Procedure:

Priority: P2

Risks:Low


middleware/playerLogManager/PlayerLogManager.cpp
```diff
@@ -157,6 +157,7 @@ void logprintf(MW_LogLevel logLevelIndex, const char* func, int line, const char
 					    ethanLogLevel = ETHAN_LOG_MILESTONE;
 					    break;
 			    }
+			    format_ptr[format_bytes-1] = 0x00; // Ethan logger appends line termination
```
pstroffolino: here we're replacing a not-needed newline by writing 0x00, since Ethern Logger is automatically applying newline on our behalf and we don't want double newline, right?

If so, comment should be updated to be more clear.  Perhaps:

// strip explicit newline, since Ethen logger will add one and we don't want it doubled

middleware/playerLogManager/PlayerLogManager.cpp
jfagunde, line 112: Does it not make more sense to simply delete the new line ('\n') at the end of this string? E.g.
```
format_bytes = snprintf(format_ptr, format_bytes,
    "%s[PLAYER_IF][%03u][%s][%zx][%s][%d]%s",
(...)
```

aamplogging.cpp
jfagunde, line 119: Same comment

middleware/playerLogManager/PlayerLogManager.cpp
shripadbpersonal: The `\n` in the format string must remain because the CLI path (`disableLogRedirection`) uses `vprintf` directly, which does not append any line termination — removing it would break aampcli log output.

The journal path (`sd_journal_printv`) and Ethan path (`vethanlog`) both add their own line termination, so for those two paths we strip the `\n` from the already-formatted buffer before handing it off. This is why the strip is done per-path at call-site rather than removing it from the shared format string.

PR #1446, opened 2026-05-14
VPAAMP-45 Create prompting guidelines

p-bond:
Attempting to create a reasonably comprehensive set of guidelines for both prompt creation and overall usage for Github Copilot.

These guidelines attempt to take account of the new billing system, trying to avoid excessive token use as well as minimising computing use (so fewer endless prompts, for example).

Bound to be edge cases that haven't been covered yet.


nebutler963 (approved):
Looks good

pstroffolino (approved):

PR #1462, opened 2026-05-18
VPAAMP-363: L1 tests for parallel init segment ordering regression

pstroffolino:
Adds L1 regression tests covering the parallel init-segment download ordering hazards introduced by PR 114108 (RDKAAMP-4072), and also includes an additional regression test plus a production-code fix for a related SegmentBase profileChanged infinite-loop bug.

Changes:
- New FetcherLoopTests regression test exercising the SegmentBase branch of FetchAndInjectInitialization when WaitForFreeFragmentAvailable(0) fails.

- Code fix in fragmentcollector_mpd.cpp adding an else branch that clears profileChanged when the ring buffer is full, preventing repeated re-entry from OnFragmentDownloadComplete.

- Three new AampTrackWorker L1 tests covering high-priority init ordering, the low-priority hazard on the DetectDiscontinuityAndFetchInit path, and the stopped-worker silent-drop behavior.

InitJob_HighPriority_ExecutesBeforeEnqueuedMediaJobs
  Verifies that SubmitJob(..., highPriority=true) places the init segment
  job at the front of the per-track worker queue, executing before any
  already-queued media jobs.  This is the protection mechanism that
  FetchFragment relies on when profileChanged=true.

InitJob_LowPriority_ExecutesAfterEnqueuedMediaJobs
  Documents the ordering hazard on the DetectDiscontinuityAndFetchInit
  path where profileChanged has already been cleared (highPriority=false).
  The init job goes to push_back and executes after queued media jobs.
  A code fix must ensure init always carries high priority on this path.

StoppedWorker_SubmitInitJob_FutureIsInvalid
  Verifies that a job submitted to a stopped worker is silently dropped
  (returned future is invalid).  Documents the stop/retune race: if
  StopWorker() wins, the init segment is lost and the decoder is never
  initialised for that tune attempt.

### DIFF
--- a/fragmentcollector_mpd.cpp
+++ b/fragmentcollector_mpd.cpp
@@ -2638,7 +2638,6 @@ double StreamAbstractionAAMP_MPD::SkipFragments( MediaStreamContext *pMediaStrea
 		ISegmentBase *segmentBase = pMediaStreamContext->representation->GetSegmentBase();
 		if (segmentBase)
 		{ // single-segment
-			// Disable parallel fragment download for segment base streams as there is a sidx box dependency for live contents
 			//SETCONFIGVALUE(AAMP_STREAM_SETTING, eAAMPConfig_DashParallelFragDownload, false);
 			std::string range = segmentBase->GetIndexRange();
 			if (pMediaStreamContext->IDX.empty())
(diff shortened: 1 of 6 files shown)


fragmentcollector_mpd.cpp
```diff
@@ -8535,8 +8535,13 @@ void StreamAbstractionAAMP_MPD::FetchAndInjectInitialization(int trackIdx, bool
...
 								AAMPLOG_TRACE("StreamAbstractionAAMP_MPD: did not cache fragmentUrl %s fragmentTime %f", fragmentUrl.c_str(), pMediaStreamContext->fragmentTime);
 							}
-							pMediaStreamContext->profileChanged = false;
 						}
+						// Clear profileChanged regardless of whether the ring buffer had space.
+						// This prevents FetcherLoop from re-triggering this init-fetch path on
+						// every subsequent OnFragmentDownloadComplete callback.  If the buffer
+						// was full the init will be re-fetched on the next profileChanged cycle
+						// once space is available.
+						pMediaStreamContext->profileChanged = false;
```
nu641001, line 8682: If we can mock WaitForFreeFragmentAvailable to control whether the cache is available before proceeding with downloads like below:
`EXPECT_CALL(*g_mockMediaTrack, WaitForFreeFragmentAvailable(_))
		.WillOnce(Return(true/false));`
We can avoid this change.


Otherwise, we should unify this change for all type of manifests irrespective of SegmentBase/Template/List

nu641001: But I can already see this under FakeMediaTrack, so moving this doesn't matter much in our L1 test suite.

```
bool MediaTrack::WaitForFreeFragmentAvailable(int timeoutMs)
{
    return true;
}
```

nu641001 (changes requested):

fragmentcollector_mpd.cpp
pstroffolino: The assignment of profileChanged = false was moved from inside the "if (WaitForFreeFragmentAvailable(0))" block to outside it. Previously, if the ring buffer was full and WaitForFreeFragmentAvailable returned false, profileChanged remained true. This caused OnFragmentDownloadComplete to re-invoke FetchAndInjectInitialization on every subsequent media segment completion — an infinite silent-skip loop — meaning the decoder never received the init segment and AAMP_EVENT_TUNED never fired

PR #1466, opened 2026-05-19
RDKEMW-16484 : [Rack][RDKE][Llama G1 UK/Cello]: Screen got stuck during VOD series & Disney content playback

rekhap2kandhavelan:
The issue is a race condition between the buffering_timeout callback and the "seek with keepPaused" flow.

Here's the sequence that causes the stuck screen:

App initiates aseek with keepPaused=true(e.g., during VOD series/Disney content playback)
The seek triggersFlush()→ which callsConfigurePipeline()to restart buffering
ConfigurePipelinesets the pipeline toPAUSEDand starts thebuffering_timeouttimer
Beforethe app'sPause(true)call arrives,buffering_timeoutfires and transitions the pipeline toPLAYING
Now whenPause(true)finally arrives, there's a state mismatch — the pipeline is already inPLAYINGbut the middleware believes it should be paused
This inconsistency causes the screen to get stuck — no frames are rendered because the seek hasn't completed cleanly, yet the pipeline is in PLAYING state
The core problem is that there was no mechanism to signal between Flush() and the various PLAYING-transition paths that a keepPaused seek is in progress. The seekPausedState flag fills that gap.

There are multiple code paths that can independently set the pipeline to PLAYING:
ConfigurePipeline(non-buffering path) buffering_timeoutcallback bus_message(CLOCK_LOST handler) NotifyFragmentCachingComplete
SetPlayBackRate
All of these must respect the "don't go to PLAYING yet" contract during a keepPaused seek — otherwise any one of them can trigger the same race. The flag is intentionally checked at each transition point as a guard.


middleware/InterfacePlayerRDK.cpp
```diff
@@ -3,7 +3,23 @@
  * following copyright and licenses apply:
  *
  * Copyright 2024 RDK Management
- *
+	// If playback rate change is a no-op (ret may be true/false depending on soc implementation),
```
mhughesacn: Looks like this additional code is in the wrong place

mhughesacn (changes requested):

mhughesacn (approved):
@rekhap2kandhavelan : Thank you

middleware/InterfacePlayerRDK.cpp
```diff
@@ -146,7 +146,7 @@ firstFrameCallbackIdleTaskId(GST_TASK_ID_INVALID), firstFrameCallbackIdleTaskPen
 using_westerossink(false), usingRialtoSink(false), usingClosedCaptionsControl(false), pauseOnStartPlayback(false), eosSignalled(false),
 buffering_enabled(FALSE), buffering_in_progress(FALSE), buffering_timeout_cnt(0),
 buffering_target_state(GST_STATE_NULL),
-lastKnownPTS(0), ptsUpdatedTimeMS(0), ptsCheckForEosOnUnderflowIdleTaskId(GST_TASK_ID_INVALID),
+seekPausedState(0),lastKnownPTS(0), ptsUpdatedTimeMS(0), ptsCheckForEosOnUnderflowIdleTaskId(GST_TASK_ID_INVALID),
```
Vinish100: Initialize with true/false for bool variables

middleware/InterfacePlayerRDK.cpp
```diff
@@ -1667,8 +1710,22 @@ bool InterfacePlayerRDK::Flush(double position, int rate, bool shouldTearDown, b
 	{
 		if ((interfacePlayerPriv->socInterface->IsSimulatorSink() || interfacePlayerPriv->gstPrivateContext->usingRialtoSink) && rate != GST_NORMAL_PLAY_RATE)
 		{
-			MW_LOG_INFO("Resetting seek position to zero");
-			position = 0;
+			const bool isTrickplay = (rate != GST_NORMAL_PLAY_RATE);
+			const bool isRialtoOrSimulator = (interfacePlayerPriv->socInterface->IsSimulatorSink() || interfacePlayerPriv->gstPrivateContext->usingRialtoSink);
```
Vinish100: Why isRialtoOrSimulator is required? This is already checked in the prior if statement and only enters when its true

middleware/InterfacePlayerRDK.cpp
```diff
@@ -1667,8 +1710,22 @@ bool InterfacePlayerRDK::Flush(double position, int rate, bool shouldTearDown, b
...
+			const bool isLiveMedia = (static_cast<GstMediaFormat>(m_gstConfigParam->media) == eGST_MEDIAFORMAT_OTA);
+			
+			if (isRialtoOrSimulator && isTrickplay)
+			{
+				if (isLiveMedia)
```
Vinish100: Why an OTA check here?

middleware/InterfacePlayerRDK.cpp
```diff
@@ -3332,17 +3395,43 @@ bool InterfacePlayerRDK::Pause(bool pause , bool forceStopGstreamerPreBuffering)
 			 * and the resume play will be handled from StopBuffering once after getting enough buffer/frames.
 			 */
 			interfacePlayerPriv->gstPrivateContext->buffering_in_progress = false;
-		}
+		}	
 
+		MW_LOG_ERR("InterfacePlayerRDK::Pause requested nextState=%s forceStopGstreamerPreBuffering=%d", gst_element_state_get_name(nextState), forceStopGstreamerPreBuffering);
```
Vinish100: Why ERR logging here?

middleware/InterfacePlayerRDK.cpp
rekhap2kandhavelan: For OTA (live) streams, resetting the seek position to 0 during trickplay is incorrect because live content doesn't have a meaningful "position 0" — the stream is continuously moving forward. Seeking to 0 on a live stream results in an invalid or stale position, which causes playback failures or a stuck screen.

For VOD content, resetting to 0 is valid (it represents the start of the asset), so we preserve the original behavior there.

The OTA check ensures we only skip the position reset for live media where it's harmful, while keeping existing VOD trickplay behavior unchanged.

middleware/InterfacePlayerRDK.cpp
```diff
@@ -1584,14 +1609,32 @@ bool InterfacePlayerRDK::Flush(double position, int rate, bool shouldTearDown, b
...
 		interfacePlayerPriv->gstPrivateContext->bufferingTimeoutTimerId = PLAYER_TASK_ID_INVALID;
+		// Reset buffering state to prevent stale timeout_cnt from triggering error after seek
+		interfacePlayerPriv->gstPrivateContext->buffering_in_progress = false;
+		interfacePlayerPriv->gstPrivateContext->buffering_timeout_cnt = DEFAULT_BUFFERING_MAX_CNT;
 
-	}
 
+	}
+	// If rate indicates playback (not paused seek), clear seekPausedState
+	if (rate > 0 && !interfacePlayerPriv->gstPrivateContext->paused)
```
Vinish100: Even when paused, rate is going to be 1, so this check needs to be refactored
So shouldn't we just check for !interfacePlayerPriv->gstPrivateContext->paused

middleware/InterfacePlayerRDK.cpp
Vinish100: If contentType is OTA means AAMP is not handling the playback. Its handling through MediaRite engine using satellite or QAM tuner. I guess AAMP will not be configuring sink as well

middleware/InterfacePlayerRDK.cpp
```diff
@@ -4377,7 +4463,17 @@ static gboolean bus_message(GstBus * bus, GstMessage * msg, InterfacePlayerRDK *
...
 				SetStateWithWarnings(privatePlayer->gstPrivateContext->pipeline, GST_STATE_PAUSED);
-				SetStateWithWarnings(privatePlayer->gstPrivateContext->pipeline, GST_STATE_PLAYING);
+				/* Avoid forcing PLAYING if a seek-with-keepPaused is active */
+				if (!privatePlayer->gstPrivateContext->seekPausedState)
+				{
+					SetStateWithWarnings(privatePlayer->gstPrivateContext->pipeline, GST_STATE_PLAYING);
+				}
+				else
+				{
+					MW_LOG_WARN("GST_MESSAGE_CLOCK_LOST: seekPausedState active - skipping PLAYING");
```
Vinish100, line 4457: Is the buffering timer running at this time?. Is this validated to be working as expected?

middleware/InterfacePlayerRDK.cpp
```diff
@@ -4443,8 +4539,34 @@ bool InterfacePlayerRDK::SetPlayBackRate(double rate)
 			sources.push_back(interfacePlayerPriv->gstPrivateContext->stream[iTrack].source);
 		}
 	}
-	ret = interfacePlayerPriv->socInterface->SetPlaybackRate(sources, interfacePlayerPriv->gstPrivateContext->pipeline, rate, interfacePlayerPriv->gstPrivateContext->video_dec,interfacePlayerPriv->gstPrivateContext->audio_dec);
-	return ret;
+		ret = interfacePlayerPriv->socInterface->SetPlaybackRate(sources, interfacePlayerPriv->gstPrivateContext->pipeline, rate, interfacePlayerPriv->gstPrivateContext->video_dec,interfacePlayerPriv->gstPrivateContext->audio_dec);
+
+		/* If application requested resume via rate change but middleware's
```
Vinish100: SetPlaybackRate is only to employ rate correction. For Pause, you have to explicitly set the pipeline to paused state and not via socInterface->SetPlaybackRate which sends a rate change event. 
You can add a check to avoid processing when rate == 0.0 or return false when pipeline is paused

middleware/InterfacePlayerRDK.cpp
```diff
@@ -4563,6 +4685,20 @@ static gboolean buffering_timeout (gpointer data)
...
+				 * but BEFORE the Pause(1) from keepPaused logic arrives — causing a race. */
+				if (privatePlayer->gstPrivateContext->seekPausedState)
+				{
+					MW_LOG_WARN("buffering_timeout: skipping PLAYING — seekPausedState active (cnt %u, frames %d)", privatePlayer->gstPrivateContext->buffering_timeout_cnt, frames);
+					if (privatePlayer->gstPrivateContext->buffering_timeout_cnt == 0)
+					{
+						MW_LOG_ERR("buffering_timeout: seekPausedState still active after timeout exhausted — clearing to unblock");
+						privatePlayer->gstPrivateContext->seekPausedState = false;
+					}
+					return privatePlayer->gstPrivateContext->buffering_in_progress;
```
Vinish100: Shouldn't we set buffering_in_progress=false and return false here to prevent the buffering timer from running again and preserving the pipeline in PAUSED state

middleware/InterfacePlayerRDK.cpp
rekhap2kandhavelan: "Yes — in the keepPaused seek flow, Flush() → ConfigurePipeline() restarts buffering and sets buffering_in_progress = true, which starts the buffering_timeout GSource timer. The seekPausedState guard was validated on Llama G1 UK/Cello racks with VOD series content — confirmed the timer fires, hits the guard, and the pipeline remains PAUSED until explicit resume."

middleware/InterfacePlayerRDK.cpp
```diff
@@ -1584,14 +1609,32 @@ bool InterfacePlayerRDK::Flush(double position, int rate, bool shouldTearDown, b
...
 		g_source_remove(interfacePlayerPriv->gstPrivateContext->bufferingTimeoutTimerId);
 		interfacePlayerPriv->gstPrivateContext->bufferingTimeoutTimerId = PLAYER_TASK_ID_INVALID;
+		// Reset buffering state to prevent stale timeout_cnt from triggering error after seek
+		interfacePlayerPriv->gstPrivateContext->buffering_in_progress = false;
+		interfacePlayerPriv->gstPrivateContext->buffering_timeout_cnt = DEFAULT_BUFFERING_MAX_CNT;
 
-	}
 
+	}
+	// If rate indicates playback (not paused seek), clear seekPausedState
```
Vinish100: rate is not checked here. Refactor the comment

Vinish100 (commented):
Looks good to me 👍

middleware/InterfacePlayerRDK.cpp
```diff
@@ -1584,14 +1606,32 @@ bool InterfacePlayerRDK::Flush(double position, int rate, bool shouldTearDown, b
 		interfacePlayerPriv->gstPrivateContext->ptsCheckForEosOnUnderflowIdleTaskId = PLAYER_TASK_ID_INVALID;
 
 	}
+	
+	/* If pipeline is paused (seek with keepPaused), mark seekPausedState
+	 * so that when ConfigurePipeline restarts buffering, the buffering_timeout callback
+	 * won't race to set PLAYING before Pause(1) arrives */
+	if (interfacePlayerPriv->gstPrivateContext->paused)
```
jameslofthouse-RED, line 1613: Do we know this will work on all platforms? For seek-while-apused to work it will eed the SOC layer to display the first iframe on PAUSE (originally this was not the case, hence the pause on first frame dispalyed)

jameslofthouse-RED (approved):
Tested on llama x3 and xione

middleware/InterfacePlayerRDK.cpp
rekhap2kandhavelan: The seekPausedState guard only blocks race-driven PLAYING transitions (buffering_timeout, CLOCK_LOST, NotifyFragmentCachingComplete). It does not change how GStreamer internally handles the seek preroll — GStreamer should still push the first I-frame to the sink during PAUSED state after a flush-seek.

PR #1473, opened 2026-05-19
VPAAMP-365: change default preferredDrm to wide vine

Reshma-JO07:
Reason for change: Set default drm as widevine
as part of VPAAMP-31
Risks: Medium


Vinish100 (commented):
Looks good to me 👍

PR #1478, opened 2026-05-20
VPAAMP-325: Enforce MAX_CACHED_FRAGMENTS_PER_TRACK as a hard upper bound for the fragment cache

DomSyna:
Before this change, the mCachedFragment ring buffer array in MediaTrack was sized by DEFAULT_LLD_CACHED_FRAGMENTS_PER_TRACK, but neither AampConfig nor SetCachedFragmentSize enforced that value as an upper bound. An integrator calling PlayerInstanceAAMP::SetDownloadBufferSize(N) with N > DEFAULT_LLD_CACHED_FRAGMENTS_PER_TRACK would not be rejected by the config layer, and a subsequent SetCachedFragmentSize(N) call would silently reject the resize (due to the size <= maxLLDCachedFragmentsPerTrack guard) while the array remained at its original capacity. This left a silent config/behaviour mismatch.

This PR introduces MAX_CACHED_FRAGMENTS_PER_TRACK as a named compile-time hard cap, aligns the config range validation with it, and resizes the array to that cap so valid values in the full accepted range can actually be used.


AampConfig.cpp
```diff
@@ -397,7 +399,7 @@ static const ConfigLookupEntryInt mConfigLookupTableInt[AAMPCONFIG_INT_COUNT+CON
 	{DEFAULT_ABR_SKIP_DURATION,"abrSkipDuration",eAAMPConfig_ABRSkipDuration,false},
 	{DEFAULT_ABR_NW_CONSISTENCY_CNT,"abrNwConsistency",eAAMPConfig_ABRNWConsistency,false},
 	{DEFAULT_AAMP_ABR_THRESHOLD_SIZE,"thresholdSizeABR",eAAMPConfig_ABRThresholdSize,false},
-	{DEFAULT_CACHED_FRAGMENTS_PER_TRACK,"downloadBuffer",eAAMPConfig_MaxFragmentCached,false},
+	{DEFAULT_CACHED_FRAGMENTS_PER_TRACK,"downloadBuffer",eAAMPConfig_MaxFragmentCached,false,eCONFIG_RANGE_FRAGMENT_CACHE},
```
cpc005, line 402: Indentation

jfagunde (approved):

PR #1481, opened 2026-05-20
VPAAMP-18 Add review instructions and tie into existing structure

p-bond:
New instructions to try and make the automated reviews more helpful/less noisy.
Hampered by the opaque nature of which LLM is being used, its lack of memory etc.


nebutler963 (approved):
looks good

pstroffolino (approved):

PR #1498, opened 2026-05-21
VPAAMP-406 suppress noisy logging

pstroffolino:
Reason for Change: tame noisy aampcli logging

aampcli is logging every manifest refresh.
This is excessive - there are events for this, that include manifest size, but logging every few seconds is excessive.  This should be moved to trace level or removed.
We also have warning level logs present for segment and manifest downloads - this appears to be accidentally included as part of the network persona support in aamp.
As warn level logs, it'll show up by default, spewing noise into logs.
Lastly, we have a longstanding behavior where aamp by default logs all info level log messages, and 'info' log level choice is only honored once tune is complete.  This is normally fine, but we have tests involving suppressDecode where the tune never succeeds, and we don't necessarily care about constant spew of info level logs.
Here, propose new semantic where log=WARN will force suppression of info level logs, even at tune time

### DIFF
--- a/test/aampcli/Aampcli.cpp
+++ b/test/aampcli/Aampcli.cpp
@@ -670,7 +670,7 @@ void MyAAMPEventListener::Event(const AAMPEventPtr& e)
 		case AAMP_EVENT_MANIFEST_REFRESH_NOTIFY:
 		{
 			ManifestRefreshEventPtr ev = std::dynamic_pointer_cast<ManifestRefreshEvent>(e);
-			AAMPCLI_PRINTF("[AAMPCLI] AAMP_EVENT_MANIFEST_REFRESH_NOTIFY received Dur[%u]:NoPeriods[%u]:PubTime[%u] manifestType[%s]\n",ev->getManifestDuration(),ev->getNoOfPeriods(),ev->getManifestPublishedTime(),ev->getManifestType().c_str());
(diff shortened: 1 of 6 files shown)


test/aampcli/Aampcli.cpp
```diff
@@ -670,7 +670,7 @@ void MyAAMPEventListener::Event(const AAMPEventPtr& e)
 		case AAMP_EVENT_MANIFEST_REFRESH_NOTIFY:
 		{
 			ManifestRefreshEventPtr ev = std::dynamic_pointer_cast<ManifestRefreshEvent>(e);
-			AAMPCLI_PRINTF("[AAMPCLI] AAMP_EVENT_MANIFEST_REFRESH_NOTIFY received Dur[%u]:NoPeriods[%u]:PubTime[%u] manifestType[%s]\n",ev->getManifestDuration(),ev->getNoOfPeriods(),ev->getManifestPublishedTime(),ev->getManifestType().c_str());
+			//AAMPCLI_PRINTF("[AAMPCLI] AAMP_EVENT_MANIFEST_REFRESH_NOTIFY received Dur[%u]:NoPeriods[%u]:PubTime[%u] manifestType[%s]\n",ev->getManifestDuration(),ev->getNoOfPeriods(),ev->getManifestPublishedTime(),ev->getManifestType().c_str());
```
Vinish100: l2 1020, 2047 and 2004 depends on this log line

Vinish100 (changes requested):

test/aampcli/Aampcli.cpp
pstroffolino: we could leave this default enabled, but give an aampcli command to suppress

PR #1507, opened 2026-05-25
VPAAMP-396 Modify Mp4Demux to ignore unknown MP4 box types

jfagunde:
Reason for Change: Be able to play a stream with an unknown MP4 box
Summary of Changes:
- Ignore MP4 boxes of unknown type
- A WARN is printed when a MP4 box of unknown type is found
- Add what() text to the ERR printed when there is an Mp4Demux exception
- Add aditional log lines to Mp4Demux
- Verify this change in L1 tests

Test Procedure: EsdsReadLenTests L1 tests should pass

Priority: P2

Risks: Low

### DIFF
--- a/mp4demux/MP4Demux.cpp
+++ b/mp4demux/MP4Demux.cpp
@@ -195,7 +195,7 @@ Mp4Demux::~Mp4Demux()
 	LogMetrics();
 }
 
-void Mp4Demux::setParseError( Mp4ParseError err )
+void Mp4Demux::setParseError( Mp4ParseError err, const char* what )
 {
 	parseError = err;
 	const char *text = nullptr;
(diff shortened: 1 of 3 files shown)


mp4demux/MP4Demux.cpp
```diff
@@ -1404,6 +1414,7 @@ bool Mp4Demux::Parse(std::shared_ptr<std::vector<uint8_t>>&& segment)
 		}
 	} catch (const Mp4ParseException& ex) {
 		setParseError(ex.code());
+		MP4_LOG_ERR("%s", ex.what());
```
jfagunde: Is this needed or is it enough with the error printed inside setParseError()?

jfagunde: Replaced with an optional error detail string parameter to setParseError()

PR #1510, opened 2026-05-26
VPAAMP 256 HLS/TS subtitle restamping breaks after SSAI ad break

DomSyna:
After an SSAI ad break in an HLS/TS stream with `eAAMPConfig_HlsTsEnablePTSReStamp` enabled, subtitles failed to display on return to live content. Root causes: flawed `RestampSubtitle()` that shifted cue timestamps (sometimes resulting in invalid -ve ones) instead of the MPEGTS header value; an incorrect PTSOffsetSec normalization that corrupted per-discontinuity offsets; and a stale rollover flag in the TS demuxer that persisted across discontinuities.


priv_aamp.cpp
```diff
@@ -300,6 +301,23 @@ static constexpr const char *mMediaFormatName[] =
...
+static void LogUrlChunked(const char* tag, std::string_view url)
+{
+	constexpr size_t kChunk = 400;
+	for (size_t off = 0; off < url.size(); off += kChunk)
+	{
+		auto chunk = url.substr(off, kChunk);
+		AAMPLOG_MIL("%s[%zu]: %.*s", tag, off,
+			static_cast<int>(chunk.size()), chunk.data());
+	}
+}
```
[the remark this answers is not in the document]
DomSyna: Thats the point to allow us to use identical URL to retrieve the manifest via CURL etc.

streamabstraction.cpp
```diff
@@ -1200,79 +1203,109 @@ static bool isWebVttSegment( const char *buffer, size_t bufferLen )
...
+	//   X-TIMESTAMP-MAP=LOCAL:<time>,MPEGTS:<integer>
+	// MPEGTS is the last field on that line so its value is terminated by '\n'
+	// or end-of-header — std::from_chars stops cleanly at the first non-digit.
+	int64_t mpegtsIn{0};
+	if (auto mpegtsField{extractField("MPEGTS:")}; mpegtsField.has_value())
 	{
-		str = std::string(buffer,bufferLen);
+		std::from_chars(mpegtsField->data(),
+			mpegtsField->data() + mpegtsField->size(), mpegtsIn);
 	}
```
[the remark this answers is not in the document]
DomSyna: RFC 8216 §E.4 is strict — MPEGTS is always a decimal integer on that line.

streamabstraction.cpp
```diff
@@ -1200,79 +1203,109 @@ static bool isWebVttSegment( const char *buffer, size_t bufferLen )
...
+	double position,
+	double duration,
+	double pts_offset_s)
 {
-	long long pts_offset_ms = pts_offset_s*1000;
-	std::string str;
-	if( ISCONFIGSET(eAAMPConfig_HlsTsEnablePTSReStamp) && isWebVttSegment(buffer,bufferLen) )
+	AAMPLOG_INFO("[RESTAMP_SUB] IN: pos=%.3f dur=%.3f pts_offset_s=%.6f in_len=%zu",
+		position, duration, pts_offset_s, bufferLen);
+
```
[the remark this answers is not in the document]
DomSyna: Typically every 6 sec so not frequent.

streamabstraction.cpp
```diff
@@ -1353,13 +1386,14 @@ void MediaTrack::ProcessAndInjectFragment(CachedFragment *cachedFragment, bool f
 				{
 					if( pContext->mPtsOffsetMap.count(cachedFragment->discontinuityIndex)==0 )
 					{
-						AAMPLOG_WARN( "blocking subtitle track injection\n" );
+						AAMPLOG_WARN( "blocking subtitle track injection waiting for pts_offset[%" PRIu64 "]\n", cachedFragment->discontinuityIndex );
```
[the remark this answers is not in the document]
DomSyna: done

streamabstraction.cpp
```diff
@@ -1200,79 +1203,109 @@ static bool isWebVttSegment( const char *buffer, size_t bufferLen )
...
-	else
+
+	// Locate the blank line separating the VTT header block from the cue blocks.
+	const std::string_view bufView{buffer, bufferLen};
+	const auto headerEndPos{bufView.find("\n\n")};
+	if (headerEndPos == std::string_view::npos)
+	{
+		AAMPLOG_WARN("[RESTAMP_SUB] malformed VTT: no header separator, returning unchanged");
+		return {buffer, bufferLen};
+	}
```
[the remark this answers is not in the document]
DomSyna: done

fragmentcollector_hls.cpp
```diff
@@ -5486,9 +5486,20 @@ void StreamAbstractionAAMP_HLS::NotifyFirstVideoPTS(unsigned long long pts, unsi
...
 	{
+		uint64_t ptsOffsetSecs = mFirstPTS.inSeconds();
+		if (ISCONFIGSET(eAAMPConfig_HlsTsEnablePTSReStamp))
+		{
+			// When PTS restamping is active the subtitle MPEGTS values are set to
+			// m_total × 90000 (session-relative, starting at 0). The Rialto subtitle
+			// renderer subtracts the pts-offset from media_PTS to produce its display
+			// time, so the offset must also be in session-relative space. At session
+			// start m_total = 0 always (m_totalDurationForPtsRestamping is reset to
+			// 0.0 in the VideoTrack constructor), so pass 0 unconditionally.
```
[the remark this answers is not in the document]
DomSyna: done

priv_aamp.cpp
```diff
@@ -300,6 +301,23 @@ static constexpr const char *mMediaFormatName[] =
...
+
 /**
```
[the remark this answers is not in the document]
DomSyna: These will appear in the log lines immediately preceding the URL. The entire point of separating these out was to reduce the line length and avoid truncation.

streamabstraction.cpp
```diff
@@ -1233,79 +1236,109 @@ static bool isWebVttSegment( const char *buffer, size_t bufferLen )
...
-			{ // trailing
-				str += std::string(prev,fin-prev);
-				prev = fin;
-			}
-		}
+		return {buffer, bufferLen};
 	}
-	else
+
+	// Locate the blank line separating the VTT header block from the cue blocks.
```
dnfaulkner, line 1265: Why do you need to do this header separator validation? I would have thought that we only care about the presence of the MPEGTS field in the X-TIMESTAMP-MAP line

streamabstraction.cpp
```diff
@@ -1233,79 +1236,109 @@ static bool isWebVttSegment( const char *buffer, size_t bufferLen )
...
+		return headerView.substr(pos + key.size());
+	}};
+
+	// Extract input MPEGTS from the VTT header.
+	// Per RFC 8216 §E.4 the format is:
+	//   X-TIMESTAMP-MAP=LOCAL:<time>,MPEGTS:<integer>
+	// MPEGTS is the last field on that line so its value is terminated by '\n'
+	// or end-of-header — std::from_chars stops cleanly at the first non-digit.
+	int64_t mpegtsIn{0};
+	if (auto mpegtsField{extractField("MPEGTS:")}; mpegtsField.has_value())
```
dnfaulkner, line 1290: Shouldn't you check that the line contains "X-TIMESTAMP-MAP" before looking for "MPEGTS:" field?

streamabstraction.cpp
```diff
@@ -1233,79 +1236,109 @@ static bool isWebVttSegment( const char *buffer, size_t bufferLen )
...
+	// For the subtitle to align, media_PTS must encode P_rel + m_total_before_N:
+	//   mpegtsOut = mpegtsIn + ptsOffset × 90000
+	//             = (firstPts_N + P_rel) × 90000 + (m_total_before_N - firstPts_N) × 90000
+	//             = (P_rel + m_total_before_N) × 90000
+	// So: display_time = cue_t + (P_rel + m_total_before_N) ✓ matches video.
+	//
+	// For MPEGTS == 0 (SSAI proxy-stripped, RFC 8216: media_PTS = cue_t):
+	// cue times are already absolute presentation times — leave as 0.
+	int64_t mpegtsOut{0};
+	if (mpegtsIn != 0)
```
dnfaulkner, line 1315: I'm not sure why you treat 0 differently - it's still a valid PTS that needs restamping

streamabstraction.cpp
```diff
@@ -1233,79 +1236,109 @@ static bool isWebVttSegment( const char *buffer, size_t bufferLen )
...
+	{
+		mpegtsOut = mpegtsIn + std::llround(pts_offset_s * static_cast<double>(kTicksPerSecond));
```
dnfaulkner, line 1320: What if this comes out negative - (if mpegtsIn rolls over)? Maybe just put a comment that we expect a discontinuity tag when PTS rolls over so we assume that it will never be negative.
Plus we don't expect to stream continuously for more than 26.5 hours

streamabstraction.cpp
```diff
@@ -1233,79 +1236,109 @@ static bool isWebVttSegment( const char *buffer, size_t bufferLen )
...
+	const auto extractField{[&](std::string_view key) -> std::optional<std::string_view>
+	{
+		const auto pos{headerView.find(key)};
+		if (pos == std::string_view::npos) return std::nullopt;
+		return headerView.substr(pos + key.size());
+	}};
+
+	// Extract input MPEGTS from the VTT header.
+	// Per RFC 8216 §E.4 the format is:
+	//   X-TIMESTAMP-MAP=LOCAL:<time>,MPEGTS:<integer>
```
dnfaulkner, line 1286: Interesting that LOCAL and MPEGTS are transposed in a capture that I've seen!
X-TIMESTAMP-MAP=MPEGTS:5515126080,LOCAL:00:00:00.000

dnfaulkner (commented):
Looks OK in principle. Just a couple of comments

streamabstraction.cpp
DomSyna: Was that from the logs when I had it backwards ;)

streamabstraction.cpp
DomSyna: No need if its not valid and doesn't contain the MPEGTS bit we treat it as =0 and let it fall through without changing it

streamabstraction.cpp
DomSyna: We discussed this. The spec indirectly states it will have a DISCONTINUITY on rollover

streamabstraction.cpp
DomSyna: The separator position is needed for two things: (1) `extractField` searches only
 within `headerView` (header-only, bounded by `headerEndPos`), so it cannot
 accidentally match `MPEGTS:` or `LOCAL:` inside a cue block's text; (2)
 `headerEndPos` is the splice point at which the rebuilt header ends and the original
 cue blocks are appended verbatim (`out.append(buffer + headerEndPos, …)`).
 Without it we cannot safely reconstruct the output file.

streamabstraction.cpp
DomSyna: Additionally, `extractField` searches only within `headerView` (bounded to the header block by the `\n\n` separator). 
The only header line that can contain `MPEGTS:` is `X-TIMESTAMP-MAP`, so an explicit prefix check is redundant. If `MPEGTS:` is absent for any reason, `mpegtsIn` stays 0 and restamping is skipped — a safe no-op."

streamabstraction.cpp
DomSyna: "When an SSAI proxy strips the `X-TIMESTAMP-MAP` MPEGTS value to 0, RFC 8216 §E.4 treats this as a signal that cue timestamps are already absolute presentation times (offsets from the start of the presentation), not relative to a 90 kHz broadcast PTS counter. Applying `pts_offset_s` to cue times in this case would incorrectly shift all subtitles. The `mpegtsIn == 0` branch is precisely the SSAI-stripped path that was the original bug — these segments should pass through unchanged."

streamabstraction.cpp
DomSyna: The implementation uses `extractField` (substring `find`) for both fields so field
order is immaterial — `MPEGTS:…,LOCAL:…` and `LOCAL:…,MPEGTS:…` are both handled
correctly. The RFC §E.4 comment just shows the canonical form.

streamabstraction.cpp
DomSyna: Can add a comment if you really want one, but I think anyone changing this code should be looking at the spec

streamabstraction.cpp
dnfaulkner: OK. I was expecting that you would only look for the MPEGTS field in the line containing X-TIMESTAMP-MAP but I guess if you didn't restrict your search to the header, you could still accidentally get a match inside a cue block's text.

streamabstraction.cpp
dnfaulkner: No, it was from inspecting a harvest of a live stream.
No prob.

streamabstraction.cpp
dnfaulkner: Hmmm. OK, if you're near 100% confident of that. That may be the case today.

streamabstraction.cpp
dnfaulkner: I don't see how that will work. We'll chat!

streamabstraction.cpp
dnfaulkner: I think that it would be helpful to add a comment.

fragmentcollector_hls.cpp
```diff
@@ -5485,9 +5485,21 @@ void StreamAbstractionAAMP_HLS::NotifyFirstVideoPTS(unsigned long long pts, unsi
 	StreamSink *sink = AampStreamSinkManager::GetInstance().GetStreamSink(aamp);
 	if (sink)
 	{
-		// The pts_offset is expected to be in seconds for RialtoSink, so we convert it to GstClockTime (nanoseconds).
-				// For non-Rialto sinks, we need to convert the pts_offset to milliseconds to maintain consistency.
-		sink->SetSubtitlePtsOffset(mFirstPTS.inSeconds());
+		uint64_t ptsOffsetSecs = mFirstPTS.inSeconds();
+		if (ISCONFIGSET(eAAMPConfig_HlsTsEnablePTSReStamp))
```
cpc005, line 5487: Could do the other way, save calling inSeconds but does it really matter?

uint64_t ptsOffsetSecs{}
if ! ISCONFIGSET(eAAMPConfig_HlsTsEnablePTSReStamp)
{
ptsOffsetSecs = mFirstPTS.inSeconds();
}

streamabstraction.cpp
```diff
@@ -1233,79 +1236,112 @@ static bool isWebVttSegment( const char *buffer, size_t bufferLen )
...
+	// MPEGTS is the last field on that line so its value is terminated by '\n'
+	// or end-of-header — std::from_chars stops cleanly at the first non-digit.
+	int64_t mpegtsIn{0};
+	if (auto mpegtsField{extractField("MPEGTS:")}; mpegtsField.has_value())
 	{
-		str = std::string(buffer,bufferLen);
+		std::from_chars(mpegtsField->data(),
+			mpegtsField->data() + mpegtsField->size(), mpegtsIn);
```
cpc005, line 1293: We are not giving any log warning if we cannot get a number from this field I.E malformed manifest

streamabstraction.cpp
dnfaulkner: Having discussed and considered the impact of this being wrong versus the time to re-test plus the fact that we're imminently going to remove this once we implement pts offset in the injection API, we concluded that we'll leave it as it is. It just means that if ads have subtitles (which they current don't appear to have), they will not be seen for the duration of the first segment of each ad.

dnfaulkner (approved):
Looks good but we've agreed to let [one point](https://github.com/rdkcentral/aamp/pull/1510/changes#diff-b7beb31ccdab37057840c13c517e4ad6d803f7532572a58d0760c9c21ca8e4b5R1305) slide due to the re-testing time overhead:
Having discussed and considered the impact not restamping if MPEGTS is 0 being wrong versus the time to re-test plus the fact that we're imminently going to remove this once we implement pts offset in the injection API, we concluded that we'll leave it as it is. It just means that if ads have subtitles (which they current don't appear to have), they will not be seen for the duration of the first segment of each ad.

streamabstraction.cpp
DomSyna: No we are not validating the manifest, but assuming it is valid. In the invalid case it will safely not get restamped.

fragmentcollector_hls.cpp
DomSyna: I think its easier to read like this, and no real performance difference

PR #1514, opened 2026-05-26
Feature/vpaamp 20 review copilot instructions

p-bond:
Harmonising, sanity checking existing instruction structure.
See ticket for full details.

### DIFF
--- a/.github/instructions/README.md
+++ b/.github/instructions/README.md
@@ -1,35 +1,54 @@
 # Copilot Instructions
 
-This directory contains specialized GitHub Copilot instructions organized by topic and language.
+This directory contains specialized GitHub Copilot instruction files,
+organized by topic and language. Each file activates automatically via
+its `applyTo` glob; load others manually when relevant.
 
-## Structure
+## Files
 
-- **`aamp.instructions.md`** - Descriptive context of the AAMP project's current architecture, key components, and existing patterns (the "AS-IS" state).
-- **`cpp.instructions.md`** - Comprehensive C++ guidelines focused on modern C++ and embedded systems
-- **`legacy-cpp-patterns.instructions.md`** - Specific guidance for modernizing complex legacy C++ code
-- **`testing.instructions.md`** - Unit testing patterns for all languages
+### Architecture & language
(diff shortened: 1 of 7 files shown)


.github/instructions/cpp.instructions.md
```diff
@@ -250,9 +287,29 @@ private:
...
+
+Guidelines:
+- Match the error-handling style of the surrounding module. Do not mix
+  paradigms within a single component.
+- Prefer explicit, predictable failure handling (return codes,
+  `std::optional`, status enums) on runtime/playback paths.
+- Reserve exceptions for genuinely exceptional construction-time or
+  configuration-time failures where the caller has no recovery path and
+  the surrounding code already uses exceptions.
+- Never throw across a C ABI boundary (GStreamer callbacks, `extern "C"`
```
jfagunde, line 306: API boundary?

.github/instructions/README.md
```diff
@@ -1,35 +1,54 @@
...
-- Better organization for the complex C++ codebase
-- Specific guidance for legacy code modernization
-- Clear separation between primary (C++) and secondary (Python/JS) language concerns
+### C++ development
+The C++ files activate on `**/*.{cpp,h,hpp,cxx,hxx}`. For most edits
+this is all you need. When refactoring legacy code, also keep
+`legacy-cpp-patterns.instructions.md` in mind.
+
+### L1 unit test work
+All five `l1-*.instructions.md` files activate on `test/utests/**`. For
```
jfagunde, line 42: Could the L1 specific instructions be converted to a skill, so they are only loaded if needed?

.github/instructions/cpp.instructions.md
p-bond: No, really ABI.

.github/instructions/README.md
p-bond: The architecture of the instructions/skills partition is correct.
The relevant instruction files should only be loaded into context when the LLM infers they are relevant to the operation.

Or at least, that's the case now.

jfagunde (approved):

PR #1519, opened 2026-05-28
VPAAMP-422 update isobmff

p-bond:
Modernise & clean up ISOBMFF code so it is more idiomatic to C++17.

Remove the macros, remove the raw pointer use and protect the API.

(NB - does not remove duplicate code from middleware, that's the next ticket)


isobmff/isobmffbox.cpp
```diff
@@ -187,77 +194,86 @@ Box* Box::constructBox(uint8_t *hdr, uint32_t maxSz, bool correctBoxSize, int ne
 		}
 		else
 		{
-#ifdef AAMP_DEBUG_BOX_CONSTRUCT
-			AAMPLOG_WARN("Box[%s] Size error:size[%u] > maxSz[%u]",type, size, maxSz);
-#endif
+			AAMPLOG_WARN("Box[%s] Size error:size[%u] > maxSz[%u]",
+				type, size, maxSz);
+			return std::make_unique<Box>(maxSz, (const char *)type);
```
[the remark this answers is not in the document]
p-bond: Agreed; done.

isobmff/isobmffbox.cpp
```diff
@@ -774,66 +797,109 @@ uint32_t EmsgBox::getMessageLen()
...
+			if (schemeIdValueLen > 0 && remainingSize >= static_cast<uint32_t>(schemeIdValueLen))
 			{
 				schemeIdValue = (uint8_t*) malloc(sizeof(uint8_t)*schemeIdValueLen);
 				READ_U8(schemeIdValue, ptr, schemeIdValueLen);
 				remainingSize -= (sizeof(uint8_t) * schemeIdValueLen);
 			}
 		}
+		else
+		{
+			AAMPLOG_WARN("Malformed emsg v0: invalid schemeIdUri field");
```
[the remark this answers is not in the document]
p-bond: Agreed; done.

isobmff/isobmffbuffer.h
```diff
@@ -407,30 +412,30 @@ class IsoBmffBuffer
...
 	Box* getChunkedfBox() const;
 
 	/**
 	 * @fn getParsedBoxes
 	 *
-	 * @return Box handle list if Box found at index given. NULL otherwise
+	 * @return Box handle list if Box found at index given. nullptr otherwise
 	 */
-	std::vector<Box*> *getParsedBoxes();
+	std::vector<std::unique_ptr<Box>> *getParsedBoxes();
```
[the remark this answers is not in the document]
p-bond: Agreed; done.

isobmff/isobmffbox.cpp
```diff
@@ -150,114 +150,134 @@ const char *Box::getType() const
...
-		return new Box(maxSz, (const char *)"UKWN");
+		return std::make_unique<Box>(maxSz, (const char *)"UKWN");
 	}
 	else
 	{
 		size = READ_U32(hdr);
 		READ_U8(type, hdr, 4);
 		type[4] = '\0';
+		for (int i = 0; i < 4; i++)
+			safeType[i] = (type[i] >= 0x20 && type[i] <= 0x7e) ? static_cast<char>(type[i]) : '.';
```
jfagunde: Block should be within brackets {}. I.e.
```
for (int i = 0; i < 4; i++)
{
	safeType[i] = (type[i] >= 0x20 && type[i] <= 0x7e) ? static_cast<char>(type[i]) : '.';
}
```

isobmff/isobmffbox.cpp
```diff
@@ -150,114 +150,134 @@ const char *Box::getType() const
...
+		safeType[4] = '\0';
```
jfagunde: [Existing code] How about using a define instead of magical number 4?

isobmff/isobmffbox.cpp
jfagunde: If the box is corrupted, wouldn't it be better to print an error message and discard it?
Inside this function safeType does not have the non-printable characters, but the function is returning type, with non-printable characters, and that is printed outside this function.
Also, is there a L1 test for a corrupted box with non-printable characters in the type?

isobmff/isobmffbox.cpp
```diff
@@ -774,65 +801,128 @@ uint32_t EmsgBox::getMessageLen()
...
+		}
+		return true;
+	};
 
 	/*
 	 * Extraction is done as per https://aomediacodec.github.io/id3-emsg/
 	 */
 	if (1 == version)
 	{
+		if (!hasRemaining((sizeof(uint32_t) * 3) + sizeof(uint64_t),
```
jfagunde, line 862: How about a comment explaining what are the fixed fields?

isobmff/isobmffbox.h
```diff
@@ -30,39 +30,119 @@
...
+{
+	uint16_t val = (static_cast<uint16_t>(buf[0]) << 8) |
+		static_cast<uint16_t>(buf[1]);
+	buf += 2;
+	return val;
+}
 
-#define READ_U32(buf) \
-	((uint32_t)buf[0] << 24) | ((uint32_t)buf[1] << 16) | ((uint32_t)buf[2] << 8) | buf[3]; buf+=4;
+inline uint16_t READ_U16(const uint8_t *&buf)
```
jfagunde: Is there a way of not duplicating this code? The same comment applies to other functions here.

isobmff/isobmffbox.h
```diff
@@ -30,39 +30,119 @@
...
 // Size of the size and tag fields in IsoBmff
 #define SIZEOF_SIZE_AND_TAG    (8)
 
-#define READ_U16(buf) \
-	(buf[0] << 8) | buf[1]; buf+=2;
+inline uint16_t READ_U16(uint8_t *&buf)
+{
+	uint16_t val = (static_cast<uint16_t>(buf[0]) << 8) |
+		static_cast<uint16_t>(buf[1]);
+	buf += 2;
```
jfagunde: How about using `sizeof(uint16_t)` to make it clear where does this value come from?

isobmff/isobmffbox.h
```diff
@@ -30,39 +30,119 @@
...
 
-#define WRITE_U64(buf, val) \
-	buf[0]= val>>56; buf[1]= val>>48; buf[2]= val>>40; buf[3]= val>>32; buf[4]= val>>24; buf[5]= val>>16; buf[6]= val>>8; buf[7]= val;
+inline uint32_t READ_U32(uint8_t *&buf)
+{
+	uint32_t val = (static_cast<uint32_t>(buf[0]) << 24) |
+		(static_cast<uint32_t>(buf[1]) << 16) |
+		(static_cast<uint32_t>(buf[2]) << 8) |
+		static_cast<uint32_t>(buf[3]);
+	buf += 4;
```
jfagunde: How about using `sizeof(uint32_t)` to make it clear where does this value come from?

isobmff/isobmffbox.h
```diff
@@ -80,11 +160,43 @@ void WriteUint64(uint8_t *dst, uint64_t val);
  */
 int ReadCStringLen(const uint8_t* buffer, uint32_t bufferLen);
 
-#define READ_U64(buf) \
-		ReadUint64(buf); buf+=8;
+inline uint64_t READ_U64(uint8_t *&buf)
+{
+	uint64_t val = ReadUint64(buf);
+	buf += 8;
```
jfagunde: How about using `sizeof(uint64_t)` to make it clear where does this value come from?

isobmff/isobmffbuffer.cpp
```diff
@@ -646,7 +736,11 @@ void IsoBmffBuffer::printMdatBoxes()
 {
 	std::vector<Box*> mdatBoxes;
 	(void)getBoxesInternal(&boxes ,Box::MDAT, &mdatBoxes);
-	printBoxesInternal(&mdatBoxes);
+	for (auto box : mdatBoxes)
+	{
+		AAMPLOG_WARN("Offset[%u] Type[%s] Size[%u]", box->getOffset(),
```
jfagunde: I'd make this a MIL. It is not a warning.

isobmff/isobmffbuffer.cpp
```diff
@@ -504,11 +587,12 @@ bool IsoBmffBuffer::getTimeScale(uint32_t &timeScale)
...
  */
-void IsoBmffBuffer::printBoxesInternal(const std::vector<Box*> *boxes)
+void IsoBmffBuffer::printBoxesInternal(
+	const std::vector<std::unique_ptr<Box>> *boxes)
 {
 	for (size_t i = 0; i < boxes->size(); i++)
 	{
-		Box *box = boxes->at(i);
+		Box *box = boxes->at(i).get();
 		AAMPLOG_WARN("Offset[%u] Type[%s] Size[%u]", box->getOffset(), box->getType(), box->getSize());
```
jfagunde: [Existing code] This should be a MIL. The same applies to the other WARN in this function.

isobmff/isobmffbox.cpp
cpc005: We are generating safeType even though it may never get used. We do not want to add extra processing if not needed.

Better to only make safe if we are going to output an error

 makeSafe(type)
AAMPLOG_WARN("Box[%s] has invalid small size[%u]", type, size);

isobmff/isobmffbox.cpp
```diff
@@ -774,65 +801,128 @@ uint32_t EmsgBox::getMessageLen()
...
 	{
+		bool malformedValueField{false};
 		int schemeIdLen = ReadCStringLen(ptr, remainingSize);
-		if(schemeIdLen > 0)
+		if(schemeIdLen > 0 && remainingSize >= static_cast<uint32_t>(schemeIdLen))
 		{
 			schemeId = (char*) malloc(sizeof(char)*schemeIdLen);
 			READ_U8(schemeId, ptr, schemeIdLen);
 			remainingSize -= (sizeof(uint8_t) * schemeIdLen);
 			int schemeIdValueLen = ReadCStringLen(ptr, remainingSize);
```
cpc005, line 901: Simpler if ReadCStringLen() returns uint32_t to avoid subsequent casting.

No reason for  ReadCStringLen() to return -1 as it does now, it can return 0 instead

isobmff/isobmffbox.h
```diff
@@ -80,11 +160,43 @@ void WriteUint64(uint8_t *dst, uint64_t val);
...
+}
+
+inline uint64_t READ_U64(const uint8_t *&buf)
+{
+	uint64_t val = ReadUint64(buf);
+	buf += 8;
+	return val;
+}
+
+inline bool IS_TYPE(const uint8_t *value, const char *type)
```
cpc005: Better if the code stuck to one type for the type field. Currently it uses both char and uint8_t to hold the field.

IS_TYPE(const char *a, const char *b) return !std::strncmp(a,b,4);

isobmff/isobmffbox.h
```diff
@@ -30,39 +30,119 @@
...
-#define READ_FLAGS(buf) \
-		(buf[0] << 16) | (buf[1] << 8) | buf[2]; buf+=3;
+inline uint8_t READ_VERSION(const uint8_t *&buf)
+{
+	uint8_t version = buf[0];
+	buf++;
+	return version;
+}
+
+inline uint32_t READ_FLAGS(uint8_t *&buf)
```
cpc005: It does not make sense to duplicate these functions because 1 takes const and one does not. It is creating 2x the amount of code to maintain.

isobmff/isobmffbox.cpp
p-bond: Done.

isobmff/isobmffbuffer.cpp
p-bond: Done.

isobmff/isobmffbuffer.cpp
p-bond: As this was already a WARN, so the other one was set to WARN. 
Changed both.

isobmff/isobmffbox.h
p-bond: Done.  Same with the other types.

isobmff/isobmffbox.h
p-bond: Done.

isobmff/isobmffbox.h
p-bond: Done.

isobmff/isobmffbox.h
p-bond: Templated it.

isobmff/isobmffbox.cpp
p-bond: Done

isobmff/isobmffbox.cpp
p-bond: Out of scope

isobmff/isobmffbox.h
p-bond: Templated.

isobmff/isobmffbox.h
p-bond: Done.

isobmff/isobmffbox.cpp
p-bond: Done.

isobmff/isobmffbox.cpp
p-bond: Only interested in robustness, not changing code flow.
L1 test created.

test/utests/tests/IsoBmffBufferTests/IsoBmffBufferTests.cpp
```diff
@@ -190,6 +190,59 @@ TEST_F(IsoBmffBufferTests, parseBufferTwiceTest)
...
 
+TEST_F(IsoBmffBufferTests, readOnlyBufferDisallowsMutation)
+{
+	std::string file_path = std::string(TESTS_DIR) + "/" + "initSegmentTests/vInit.mp4";
+	auto result = readFile(file_path.c_str());
+	std::vector<uint8_t> vInitSeg;
+	if (!result.first.empty())
+	{
+		vInitSeg = result.first;
+	}
```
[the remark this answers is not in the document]
p-bond: Done.

isobmff/isobmffbuffer.cpp
```diff
@@ -134,17 +125,43 @@ bool IsoBmffBuffer::ParseChunkData(const char* name, uint8_t* &unParsedBuffer, u
  */
 bool IsoBmffBuffer::parseBuffer(bool correctBoxSize, int newTrackId)
 {
+	constexpr size_t minHeaderSize = sizeof(uint32_t) + sizeof(uint32_t);
 	size_t curOffset = 0;
 	while (curOffset < bufSize)
 	{
```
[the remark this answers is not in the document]
p-bond: Done.

isobmff/isobmffbuffer.cpp
```diff
@@ -134,17 +125,43 @@ bool IsoBmffBuffer::ParseChunkData(const char* name, uint8_t* &unParsedBuffer, u
...
+			AAMPLOG_WARN("Invalid box size[%u] at offset %zu (remaining %zu); stopping parse",
+				boxSize, curOffset, remaining);
+			break;
+		}
 		if( ((bufSize - curOffset) < 4) || ( (bufSize - curOffset) < box->getSize()) )
 		{
-			chunkedBox = box;
+			chunkedBox = box.get();
 		}
 		box->setOffset((uint32_t)curOffset);
```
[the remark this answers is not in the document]
p-bond: Done

nebutler963 (approved):
Looks Good

PR #1525, opened 2026-06-01
VPAAMP-435 DirectRialto - Position Reporting

jfagunde:
Reason for Change: Comply with Position Reporting AC

Test Procedure: Configure and check ProgressReport

Priority: P1

Risks: Low

### DIFF
--- a/direct-rialto/AampRialtoPlayer.cpp
+++ b/direct-rialto/AampRialtoPlayer.cpp
@@ -319,8 +319,6 @@ void AampRialtoPlayer::Configure(
 						OnBufferUnderflow(sid);
 					});
 
-				StartProgressTimer();
-
 				// Advance state machine: pipeline is now created and loaded.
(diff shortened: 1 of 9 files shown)


direct-rialto/AampRialtoPlayer.cpp
```diff
@@ -314,6 +319,8 @@ void AampRialtoPlayer::Configure(
 						OnBufferUnderflow(sid);
 					});
 
+				StartProgressTimer();
```
anshephe: Is this where we would want to start the timer?
Wouldn't it start on playback?

jfagunde: You are right. I'll change the implementation to match what is done for AampGstPlayer (start the progress timer when the pipeline transitions to PLAYING)

jfagunde: Done

anshephe (approved):

PR #1530, opened 2026-06-02
VPAAMP-304: Audio Drops & freeze for 1-2 Seconds on Change of linear …

psiva01:
…channel

Reason for change: Added retry logic for curl 55 error to avoid this AV glitch


Vinish100 (commented):
Looks good to me 👍

PR #1531, opened 2026-06-02
VPAAMP-477: a single curl 56 on manifest refresh causes manifest refresh attempts to stop

pstroffolino:
Reason for Change: After a download failure mMPDData->mIsLiveManifest is false (only set during successful parse), so refreshNeeded was never set and the downloader thread exited on the first refresh error.

Add an else-if branch below the existing timeout/COULDNT_CONNECT fast-retry block to catch all other non-success error codes (including CURLE_RECV_ERROR / curl 56) when past the first download. Sets a 500 ms retry interval and keeps the refresh loop alive so the next manifest fetch is attempted rather than killing the downloader thread.

The !firstDownload guard ensures tune-time failures still exit the loop immediately, preserving existing behaviour for initial tune failures.

### DIFF
--- a/AampMPDDownloader.cpp
+++ b/AampMPDDownloader.cpp
@@ -347,6 +347,7 @@ void AampMPDDownloader::downloadMPDThread1()
 	bool refreshNeeded = false;
 	std::string tuneUrl = mMPDDnldCfg->mTuneUrl;
 	bool firstDownload	=	true;
+	bool downloadFailed = false; // set when a refresh attempt fails; cleared on recovery to log the transition
 	ManifestDownloadResponsePtr cachedBackupData = nullptr;
 	do
 	{
@@ -418,9 +419,11 @@ void AampMPDDownloader::downloadMPDThread1()
 					doPush = readMPDData(mMPDData);
 				}
 				AAMPLOG_INFO("Successfully parsed Manifest ...IsLive[%d]",mMPDData->mIsLiveManifest);
-
-				// Update the effective url , so that next refresh uses the effective url
-				tuneUrl = mMPDData->mMPDDownloadResponse->sEffectiveUrl;
+				if(downloadFailed)
+				{
+					AAMPLOG_WARN("Manifest refresh recovered after previous download failure.");
+					downloadFailed = false;
+				}
 
 				// first time download complete . Do what need to be done . ....
 				if(firstDownload && mMPDData->mIsLiveManifest)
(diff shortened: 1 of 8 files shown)


AampMPDDownloader.cpp
```diff
@@ -516,12 +519,24 @@ void AampMPDDownloader::downloadMPDThread1()
...
+		else if(!firstDownload && !IS_HTTP_SUCCESS(mMPDData->mMPDDownloadResponse->iHttpRetValue) && !mReleaseCalled)
+		{
+			// Other download failures (e.g. curl 56 CURLE_RECV_ERROR, transient HTTP errors)
+			// during an established live session: keep the refresh loop alive so the next
+			// manifest fetch is attempted rather than killing the downloader thread.
+			// mRefreshInterval retains the MPD-derived value from the last successful parse
+			// so no override is needed; the stream's own minimumUpdatePeriod is respected.
+			AAMPLOG_WARN("Manifest download error [%d], will retry after %ums.", mMPDData->mMPDDownloadResponse->iHttpRetValue, mRefreshInterval);
+			downloadFailed = true;
 			refreshNeeded = waitForRefreshInterval(mRefreshInterval);
```
Vinish100: This would have an impact in LLD use-cases.
Waiting for minUpdateDuration(1.92s) will cause buffer drop of same amount and could also lead to buffering. Any non-HTTP error should also follow the 500ms refresh path to have a less impact on buffering

AampMPDDownloader.cpp
```diff
@@ -418,9 +419,11 @@ void AampMPDDownloader::downloadMPDThread1()
 					doPush = readMPDData(mMPDData);
 				}
 				AAMPLOG_INFO("Successfully parsed Manifest ...IsLive[%d]",mMPDData->mIsLiveManifest);
-
-				// Update the effective url , so that next refresh uses the effective url
-				tuneUrl = mMPDData->mMPDDownloadResponse->sEffectiveUrl;
+				if(downloadFailed)
```
Vinish100, line 426: Please bring back tuneUrl = mMPDData->mMPDDownloadResponse->sEffectiveUrl;

AampMPDDownloader.cpp
```diff
@@ -516,12 +523,24 @@ void AampMPDDownloader::downloadMPDThread1()
...
+		{
+			// Other download failures (e.g. curl 56 CURLE_RECV_ERROR, transient HTTP errors)
+			// during an established live session: keep the refresh loop alive so the next
+			// manifest fetch is attempted rather than killing the downloader thread.
+			// mRefreshInterval retains the MPD-derived value from the last successful parse
+			// so no override is needed; the stream's own minimumUpdatePeriod is respected.
+			AAMPLOG_WARN("Manifest download error [%d], will retry after %ums.", mMPDData->mMPDDownloadResponse->iHttpRetValue, mRefreshInterval);
+			downloadFailed = true;
 			refreshNeeded = waitForRefreshInterval(mRefreshInterval);
 		}
```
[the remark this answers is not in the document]
pstroffolino: The comment is stale/should be marked outdated. Test AampMPDDownloader_LiveRefreshRetriesWhenFailureIsCurlRecvError was added in commit 76e647c9 and does exactly what the comment requested: injects CURLE_RECV_ERROR after the first successful live manifest fetch, then asserts the downloader continues producing subsequent refresh attempts (second and third responses observed within the 500ms fast-retry window).

PR #1542, opened 2026-06-04
Feature/vpaamp 457 missing web vtt subtitles

DomSyna:
When tuning to certain channels i.e NBCNewsNOW subtitles aren't shown.

To fix the issue the pts offset value is calculated and passed through the stack to gstreamer / subtec so it can apply the offset directly as is currently done for DASH


fragmentcollector_hls.cpp
```diff
@@ -1807,12 +1807,23 @@ void TrackState::InjectFragmentInternal(CachedFragment* cachedFragment, bool &fr
 	else
 	{
 		fragmentDiscarded = false;
-		aamp->SendStreamCopy(
+		// Route every non-demuxed fragment via SendStreamTransfer so
+		// cachedFragment->PTSOffsetSec, initFragment and discontinuity
+		// are forwarded to the gstreamer sink. Mirrors the DASH
```
dnfaulkner: Change this (since direct Rialto does not use gstreamer)
gstreamer -> Stream

fragmentcollector_hls.cpp
```diff
@@ -1807,12 +1807,23 @@ void TrackState::InjectFragmentInternal(CachedFragment* cachedFragment, bool &fr
...
+		// MediaStreamContext::InjectFragmentInternal else branch. For
+		// HLS bare elementary audio (AAC/AC3/EC3) PTSOffsetSec is 0
```
dnfaulkner: For HLS-TS audio and video, the PTS has already been re stamped (if restamping enabled) so PTSOffsetSec is 0, initFragment is always false and the discontinuity is false if PTS restamping is enabled.

dnfaulkner (commented):
A couple of small comments about the code comments

jfagunde (approved):

pstroffolino (approved):

PR #1545, opened 2026-06-04
VPAAMP-490: Prioritize latencyParams present in manifest over init config

nu641001:
Reason for change: When latencyParams are present in the manifest, they should take precedence over the values provided in the initialization configuration. This change ensures that the player uses the most accurate and up-to-date latency parameters as specified by the content provider, leading to improved performance and user experience. Risks: Low
Test Procedure: Test with LLD streams
Priority: P1

Change-Id: I6a1efdee4c654012377604c31e9a221cd70aea68


Vinish100 (commented):
Looks good to me 👍

PR #1553, opened 2026-06-05
VPAAMP-473:Timeshift DAI Place CDAI ads for cold CDVR and iVOD

varshnie:
Reason for change:cdai support for cold cdvr
Priority: P1

### DIFF
--- a/admanager_mpd.cpp
+++ b/admanager_mpd.cpp
@@ -97,7 +97,7 @@ bool CDAIObjectMPD::IsAdPlaying()
  */
 PrivateCDAIObjectMPD::PrivateCDAIObjectMPD(PrivateInstanceAAMP* aamp) : mAamp(aamp),mDaiMtx(), mIsFogTSB(false), mAdBreaks(), mPeriodMap(), mCurPlayingBreakId(), mAdObjThreadID(), mCurAds(nullptr),
 					mCurAdIdx(-1), mContentSeekOffset(0), mAdState(AdState::OUTSIDE_ADBREAK),mPlacementObj(), mAdFulfillObj(),currentAdPeriodClosed(false),mAdtoInsertInNextBreakVec(),
-					mAdBrkVecMtx(), mAdFulfillMtx(), mAdFulfillCV(), mAdFulfillQ(), mExitFulfillAdLoop(false), mAdPlacementMtx(), mAdPlacementCV()
+					mAdBrkVecMtx(), mAdFulfillMtx(), mAdFulfillCV(), mAdFulfillQ(), mExitFulfillAdLoop(false), mAdPlacementMtx(), mAdPlacementCV(), mCachedMPDParseHelper(nullptr)
 {
 	StartFulfillAdLoop();
 	mAamp->CurlInit(eCURLINSTANCE_DAI,1,mAamp->GetNetworkProxy());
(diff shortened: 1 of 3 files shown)


admanager_mpd.cpp
```diff
@@ -1318,6 +1318,15 @@ bool PrivateCDAIObjectMPD::FulFillAdObject()
 						}
 						node.url = mAdFulfillObj.url;
 						node.resolved = true;
+						// For cold CDVR the manifest is static — no periodic PlaceAds() calls happen.
+						// Trigger immediate placement now that all node fields (mpd, duration, url) are
+						// fully populated, so PlaceAds reads the correct ad duration during stitching.
+						if(adStatus && !mAamp->IsLive() && mAamp->IsCDVRContent() && mCachedMPDParseHelper)
```
Vinish100: Can remove the IsCDVRContent check

admanager_mpd.h
```diff
@@ -392,6 +392,7 @@ class PrivateCDAIObjectMPD
 	std::mutex                                     mAdPlacementMtx;       /**< Mutex protecting Ad placement */
 	std::condition_variable                        mAdPlacementCV;        /**< Condition variable for Ad placement */
 	uint64_t                                       mWaitForManifestUpdate;/**< segment position in manifest at end of Ad */
+	AampMPDParseHelperPtr                          mCachedMPDParseHelper; /**< Cached MPD parse helper for cold CDVR immediate ad placement */
```
Vinish100: instead of cold cdvr, it should be static manifests

fragmentcollector_mpd.cpp
```diff
@@ -3056,6 +3056,13 @@ AAMPStatusType StreamAbstractionAAMP_MPD::GetMPDFromManifest( ManifestDownloadRe
 		{
 			mCdaiObject->PlaceAds(mMPDParseHelper);
 		}
+		// For cold CDVR, the manifest is static so PlaceAds() will never be called again
```
Vinish100: if its mIsLiveStream => Call PlaceAds
if its !mIsLiveStream => Call a new function which will cache mMPDParseHelper and then also place the remaining ads. This condition will come into play when CDVR moves from Hot to Cold and pending placements should be resolved as there will be no further refresh

fragmentcollector_mpd.cpp
```diff
@@ -5328,12 +5335,14 @@ bool StreamAbstractionAAMP_MPD::ProcessEventStream(uint64_t startMS, int64_t sta
 					AAMPLOG_INFO("SCTEDBG adjust start time %" PRIu64 " -> %" PRIu64 " (duration %d)", eventInfo.presentationTime, eventStartTime, eventInfo.duration);
 				}
 
-				//for livestream send the timedMetadata only., because at init, control does not come here
-				if(mIsLiveManifest && ! ISCONFIGSET(eAAMPConfig_BulkTimedMetaReportLive))
+				//for livestream or cold CDVR, route through CDAI pipeline via FoundEventBreak
+				bool isColdCdvr = !mIsLiveManifest && aamp->IsCDVRContent();
+				if(isColdCdvr || (mIsLiveManifest && ! ISCONFIGSET(eAAMPConfig_BulkTimedMetaReportLive)))
```
Vinish100: This needs rework. The logic should not be cdvr specific. What happens when we test iVOD

PR #1582, opened 2026-06-09
VPAAMP-529 [AAMP][SKY][XiOne-DE] HD VOD playback fails after ad start with "Please wait until the program is loaded" error

pstroffolino:
Reason for Change: reworked with simpler implementation


Vinish100 (approved):
Looks good to me 👍

PR #1604, opened 2026-06-12
Feature/vpaamp 411 debug

nebutler963:
This fixes a race condition, where the clean up of the pipeline occasionally  before the pipeline was successfully stopped.  This lead to a crash.
https://ccp.sys.comcast.net/browse/VPAAMP-411


p-bond (approved):
Looks good.
Intermittent L1 failures not affected by this.

PR #1625, opened 2026-06-18
RDKEMW-20076: Fix for WPEWebProcess during deepsleep wakeup scenario

dp0000:
Update done to fix the crash that happens in deep sleep wakeup scenario , where the webprocess is unresponsive.

Once the HDMI hotplug event arrived, Instead of blocking the IARM thread with SetHDMIStatus() → dsDisplayInit → broken D-Bus, the callback would have returned immediately.
The spawned worker thread would have blocked on the broken IARM Bus, but this would NOT have caused the WPE watchdog to trigger SIGFPE, because the WebProcess main thread would have remained responsive.
This is the most impactful single fix for preventing the crash scenario observed.


Vinish100 (approved):
Looks good to me 👍

PR #1653, opened 2026-06-24
VPAAMP-394:[VIPA - TS DAI][cDVR on XiOne/Llama] Ads are not played on cdvr asset and getting "Error: Ad fetch failed"

srikanthreddybijjam-comcast:
Reason for change: Invoke FoundEventBreak for both static and dynamic manifests when CDAI is enabled to support CDAI on cold CDVR. Avoid calling onAdEvent from init() to prevent processing DAI events during the tune-time period for a new tune.Added L1s.
Test Procedure: Refer jira ticket VPAAMP-394
Priority: P1


Vinish100 (commented):
Looks good to me 👍

PR #1654, opened 2026-06-24
VPAAMP-565: Deprecate ceaFormat from AAMP Config.

Gnanesha:
Reason for Change: Deprecated ceaFormat from AAMP

Test Procedure: see ticket
Priority: P1
Risks: Low

### DIFF
--- a/jsbindings/jsbindings.cpp
+++ b/jsbindings/jsbindings.cpp
@@ -498,7 +498,6 @@ static bool AAMP_setProperty_preferredCEAFormat(JSContextRef context, JSObjectRe
 		return false;
 	}
 	preferredCEAFormat = (int)JSValueToNumber(context, value, exception);
-		LOG_WARN(pAAMP, "preferredCEAFormat is deprecated; ignoring (context=%p value=%d exception=%p)", context, preferredCEAFormat, exception);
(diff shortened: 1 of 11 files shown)


jsbindings/jsbindings.cpp
```diff
@@ -498,8 +498,7 @@ static bool AAMP_setProperty_preferredCEAFormat(JSContextRef context, JSObjectRe
 		return false;
 	}
 	preferredCEAFormat = (int)JSValueToNumber(context, value, exception);
-        LOG_WARN(pAAMP,"_aamp->SetCEAFormat context=%p  value=%d exception=%p ",context, preferredCEAFormat, exception);
-	pAAMP->_aamp->SetCEAFormat(preferredCEAFormat);
+		LOG_WARN(pAAMP, "preferredCEAFormat is deprecated; ignoring (context=%p value=%d exception=%p)", context, preferredCEAFormat, exception);
```
pstroffolino: do we need an explicit deprecation warning, or could we truly gut all references and let a generic "api doesn't exist" warning to be logged?

Gnanesha: getting rid of this as well

PR #1674, opened 2026-07-01
VPAAMP-651: Trickplay skips incorrectly at cdai ad boundaries

varshnie:
Reason for change:Init segments at ad boundaries carry discontinuity flag but yield no samples,leaving state machine in STEADY phase. Next ad data segment computes incorrect PTS delta, causing catastrophic position jump to stream end.
Risks: p1

Signed-off-by: varshnie <varshniblue14@gmail.com>


Vinish100 (commented):
Looks good to me 👍

pstroffolino (approved):

PR #1704, opened 2026-07-08
Fix SaizBox truncate sizing for default sample info entries

p-bond:
Fix SaizBox truncate sizing for default sample info layouts

SaizBox::truncate assumed every SAIZ box had a per-sample info table and always reduced the box size by sample_count - 1 when truncating to one sample. That is only valid when default_sample_info_size is zero. For boxes that use a non-zero default sample info size, there is no per-sample table to shrink, so the old logic could write an invalid smaller size and corrupt the rewritten buffer. 

This change preserves whether a per-sample table is present when parsing SAIZ and only shrinks the box when that table exists. It also returns early for sample counts of 0 or 1 to avoid invalid size arithmetic. 

The L1 IsoBmffBoxTests suite now covers the table-present path, the default-size path where size must stay unchanged, and the low-sample guard behavior.


isobmff/isobmffbox.cpp
```diff
@@ -1635,13 +1637,14 @@ SaizBox* SaizBox::constructSaizBox(uint32_t sz, uint8_t *ptr)
 
 	uint8_t *sample_count_loc{ptr};
 	uint32_t sample_count = READ_U32(ptr);
+	bool has_per_sample_info_table{0 == default_sample_info_size};
 
 	uint8_t sample_info_size = (0 == default_sample_info_size) ? *ptr : default_sample_info_size;
```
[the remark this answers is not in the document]
p-bond: Done.

isobmff/isobmffbox.h
```diff
@@ -1129,7 +1130,7 @@ class SaizBox : public FullBox
 	 * @param[in] numSamples - number of samples
 	 * @param[in] sample_info_size - Size for the first auxiliary sample information entry
 	 */
-	SaizBox(FullBox &fbox, uint8_t *sampleCountLoc, uint32_t numSamples, uint32_t sample_info_size);
+	SaizBox(FullBox &fbox, uint8_t *sampleCountLoc, uint32_t numSamples, uint32_t sample_info_size, bool hasPerSampleInfoTable);
```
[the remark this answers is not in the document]
p-bond: Done.

test/utests/tests/IsoBmffBoxTests/IsoBmffBoxTests.cpp
```diff
@@ -139,10 +139,73 @@ TEST_F(IsoBmffBoxTests, saizTests)
 	saiz->truncate();
 
 	// First data set has only one sample, so truncate does nothing.
+	EXPECT_EQ(0, memcmp(buffer, saizSingleSample_expected, sizeof(saizSingleSample_expected)));
+
+	// Table-present layout (default_sample_info_size == 0) with enough room to insert skip.
```
jfagunde: How about making each scenario a separate test?

test/utests/tests/IsoBmffBoxTests/IsoBmffBoxTests.cpp
```diff
@@ -139,10 +139,73 @@ TEST_F(IsoBmffBoxTests, saizTests)
...
+	const uint8_t saizTablePresent[] = {
```
jfagunde: I would move these arrays to IsoBMFFTestData.h, to be consistent with existing tests.

test/utests/tests/IsoBmffBoxTests/IsoBmffBoxTests.cpp
p-bond: Done

test/utests/tests/IsoBmffBoxTests/IsoBmffBoxTests.cpp
p-bond: Done

jfagunde (approved):

pstroffolino (approved):

PR #1705, opened 2026-07-08
VPAAMP-733 GetStreamFormat() wrong for HLS-TS when useMp4Demux=true

jfagunde:
Reason for Change: GetStreamFormat() returns wrong A/V format for HLS-TS
                   when useMp4Demux=true

Summary of Changes:
- For HLS, only set format to UNKNOWN if Mp4Demux is used for fMP4
- Add L1 tests to verify the change

Test Procedure: Run L1 tests

Priority: P1

Risks: Low

### DIFF
--- a/priv_aamp.cpp
+++ b/priv_aamp.cpp
@@ -14956,14 +14956,6 @@ void PrivateInstanceAAMP::GetStreamFormat(StreamOutputFormat &primaryOutputForma
 {
 	mpStreamAbstractionAAMP->GetStreamFormat(primaryOutputFormat, audioOutputFormat, subtitleOutputFormat);
 
-	if (ISCONFIGSET_PRIV(eAAMPConfig_UseMp4Demux) &&
-		(mMediaFormat != eMEDIAFORMAT_HLS))
-	{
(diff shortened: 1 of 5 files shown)


priv_aamp.cpp
```diff
@@ -14956,6 +14956,14 @@ void PrivateInstanceAAMP::GetStreamFormat(StreamOutputFormat &primaryOutputForma
 {
 	mpStreamAbstractionAAMP->GetStreamFormat(primaryOutputFormat, audioOutputFormat, subtitleOutputFormat);
 
+	if (ISCONFIGSET_PRIV(eAAMPConfig_UseMp4Demux) &&
```
cpc005: Can you take a similar block of code out of dash: GetStreamFormat() and then remove (mMediaFormat != eMEDIAFORMAT_HLS) 

Is there a reason why dash and hls GetStreamFormat() need to be different?

anshephe (approved):

priv_aamp.cpp
jfagunde: I moved this back to what it was. Setting the format to UNKNOWN is back to StreamAbstractionAAMP_HLS, so this code is common for DASH and HLS.

PR #1712, opened 2026-07-10
VPAAMP-602: LLD video freeze

Reshma-JO07:
Reason for change: Rampdown when curl 56 occurs
Risks: Medium


MediaStreamContext.cpp
```diff
@@ -951,6 +952,20 @@ void MediaStreamContext::OnFragmentDownloadFailed(DownloadInfoPtr dlInfo)
 				}
 			}
 		}
+		else if ((eTRACK_VIDEO == type) && !dlInfo->isInitSegment &&
```
Vinish100: Retry should be done internally in GetFile with a 100-200ms delay

priv_aamp.cpp
```diff
@@ -4671,6 +4710,16 @@ bool PrivateInstanceAAMP::GetFile( std::string remoteUrl, AampMediaType mediaTyp
 			{
 				// Reset context values specific to each download attempt
 				context.ResetForNewDownload();
+				if (resumeAttemptPending)
```
Vinish100: This should be a separate feature. Not part of this

priv_aamp.cpp
```diff
@@ -4990,10 +5042,17 @@ bool PrivateInstanceAAMP::GetFile( std::string remoteUrl, AampMediaType mediaTyp
 						print_headerResponse(context.allResponseHeaders, mediaType);
 
 					}
+						bool isRetryableDnsFailure =
+							(res == CURLE_COULDNT_RESOLVE_HOST) &&
```
Vinish100: Retry on DNS resolve failure, never succeeds. That's the data we have so far.

priv_aamp.cpp
```diff
@@ -5204,6 +5263,31 @@ bool PrivateInstanceAAMP::GetFile( std::string remoteUrl, AampMediaType mediaTyp
...
+						resumeOffsetBytes = buffer.size();
+						resumeAttemptPending = true;
+					}
+					else
+					{
+						resumeOffsetBytes = 0;
+						resumeAttemptPending = false;
+					}
+
+					int retryWaitMs = CalculateCurlRetryBackoffMs(res, downloadAttempt);
```
Vinish100: Lets keep it simple. Fixed retry delay for now. There is only 1 retry, so there is no need for calculation here

streamabstraction.cpp
```diff
@@ -2660,7 +2660,8 @@ bool StreamAbstractionAAMP::CheckForRampDownProfile(int http_error)
 
 		if (http_error == 404 || http_error == 403 ||
```
Vinish100, line 2661: the plan is to remove these conditions going forward

MediaStreamContext.cpp
Reshma-JO07: done

priv_aamp.cpp
Reshma-JO07: Removed

priv_aamp.cpp
Reshma-JO07: Done

streamabstraction.cpp
Reshma-JO07: ok.

streamabstraction.cpp
```diff
@@ -2676,7 +2676,7 @@ bool StreamAbstractionAAMP::CheckForRampDownProfile(int http_error)
 		// to 480p), UpdateProfileBasedOnFragmentCache would ramp UP instead of down, causing an
 		// infinite 480p-stall → ramp-up-to-1080p → 1080p-stall → ramp-down → 480p-stall loop.
 		// FragmentfailureRampdown already performs multi-step ramp-downs for timeout scenarios.
-		else if (IsCurlTimeoutFailure (http_error))
+		else if (IsCurlTimeoutFailure (http_error) || CURLE_RECV_ERROR == http_error )
```
Vinish100: How different its from the if block? Do we need separate if-else to call the same function the same way and the retValue=true.

Reshma-JO07: That we can make a single line, doesnot have any difference in Rampdown functionality wrt current code, because in RampDownProfile function currently we are not checking http_error for any decision making other than that for logging purpose.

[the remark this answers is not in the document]
Reshma-JO07: Done

streamabstraction.cpp
Vinish100: Lose the comment, which is not pointing to any code now

Reshma-JO07: Done

test/utests/tests/AampCurlDownloader/FunctionalTests.cpp
```diff
@@ -590,6 +628,44 @@ TEST_F(FunctionalTests, AampCurlDownloader_Retry_SendError)
 	EXPECT_FALSE(mAampCurlDownloader->IsDownloadActive());
 }
 
+TEST_F(FunctionalTests, AampCurlDownloader_Retry_RecvError)
```
Vinish100, line 638: Add some comments on what this test validates. It will be easier for the next person to debug the test in case it fails

Reshma-JO07: Done

PR #1764, opened 2026-07-22
VPAAMP-827 seeking can lead to frozen AV with pts restamping enabled

pstroffolino:
Video can freeze when seeking to period boundaries.


test/utests/tests/StreamAbstractionAAMP_MPD/ManualTestScenarios.cpp
```diff
@@ -0,0 +1,50 @@
+/**
```
DomSyna: Not sure on the point of this test file given the other file appears to do the tests ?
copy notice etc ?

test/utests/tests/StreamAbstractionAAMP_MPD/ManualTestScenarios.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * Replicates manual test:
+ * - Load VPAAMP-785_ads_only.mpd
+ * - sleep 2000ms (allow initial playback)
+ * - seek 5s (seek within first period near boundary)
+ * - sleep 5000ms (verify playback continues)
+ * - stop
+ * 
+ * Expected: Seek completes successfully, playback continues from 5s position
+ */
+TEST_F(ManualTestScenarios, SeekWithinPeriodNearBoundary_VPAAMP785) {
```
DomSyna: Didn't think we were supposed to be referencing JIRA tickets in the public Repo ?

test/utests/tests/StreamAbstractionAAMP_MPD/ManualTestScenarios.cpp
```diff
@@ -0,0 +1,50 @@
...
+    // TODO: Implement test that:
```
DomSyna: Would be good to implement the test scenarios and prove the fix addresses them before delivering.

test/utests/tests/StreamAbstractionAAMP_MPD/PeriodBoundarySeekTests.cpp
```diff
@@ -0,0 +1,207 @@
+/*
+ * If not stated otherwise in this file or this component's license file the
+ * following copyright and licenses apply:
+ *
+ * Copyright 2024 RDK Management
```
DomSyna: 2026 ?

test/utests/tests/StreamAbstractionAAMP_MPD/PeriodBoundarySeekTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+    GTEST_SKIP() << "Test infrastructure needs to be set up";
+}
+
+/**
+ * @brief Simplified unit test for SkipFragments logic
+ * 
+ * This test directly exercises the SkipFragments calculation without
+ * the full MPD infrastructure.
+ */
+TEST_F(PeriodBoundarySeekTest, SkipFragments_Calculation_PeriodBoundary)
```
DomSyna: Are these tests actually testing any code ?

Looks to be just testing A<B ?

fragmentcollector_mpd.cpp
```diff
@@ -2190,6 +2190,45 @@ void StreamAbstractionAAMP_MPD::SeekInPeriod( double seekPositionSeconds, bool s
 		}
 
 	}
+	
+	// After SkipFragments, check if tracks selected fragments in different periods
+	// This can happen when seeking near a period boundary where audio/video fragments have
+	// slightly different end times. Ensure both tracks start in the same period.
+	bool videoEOS = false, audioEOS = false;
+	double videoFragTime = 0, audioFragTime = 0;
```
DomSyna: Is this used ? I can't spot it

fragmentcollector_mpd.cpp
```diff
@@ -2190,6 +2190,45 @@ void StreamAbstractionAAMP_MPD::SeekInPeriod( double seekPositionSeconds, bool s
...
+	
+	if (mMediaStreamContext[eMEDIATYPE_VIDEO] && mMediaStreamContext[eMEDIATYPE_VIDEO]->enabled)
+	{
+		videoEOS = mMediaStreamContext[eMEDIATYPE_VIDEO]->eos;
+		videoFragTime = mMediaStreamContext[eMEDIATYPE_VIDEO]->fragmentTime;
```
DomSyna: videoFragTime  doesn't seem to be used later

fragmentcollector_mpd.cpp
```diff
@@ -2190,6 +2190,45 @@ void StreamAbstractionAAMP_MPD::SeekInPeriod( double seekPositionSeconds, bool s
...
+	}
+	if (mMediaStreamContext[eMEDIATYPE_AUDIO] && mMediaStreamContext[eMEDIATYPE_AUDIO]->enabled)
+	{
+		audioEOS = mMediaStreamContext[eMEDIATYPE_AUDIO]->eos;
+		audioFragTime = mMediaStreamContext[eMEDIATYPE_AUDIO]->fragmentTime;
```
DomSyna: audioFragTime  doesn't seem to be used after here

fragmentcollector_mpd.cpp
```diff
@@ -2190,6 +2190,45 @@ void StreamAbstractionAAMP_MPD::SeekInPeriod( double seekPositionSeconds, bool s
...
+	}
+	
+	// If any track hit EOS but the original seek position is within the current period,
+	// this is a spurious EOS caused by seeking near the period boundary or tracks having
+	// slightly different durations. Clear the EOS flags to prevent unwanted period transition.
+	// For live streams, use < to allow seeks at period boundary to trigger manifest refresh.
+	// For VOD, use <= to handle exact boundary seeks (e.g., 28.8s in 28.8s period).
+	// If seeking BEYOND the period (e.g., 12s in 10s period), allow the EOS and
+	// period transition to occur with the remaining seek value.
+	double periodDurationSeconds = mPeriodDuration / 1000.0;
```
DomSyna: isn't mPeriodDuration 0 of live streams with open ended current period ? I may be wrong but elsewhere we check (0 != mPeriodDuration)

I guess this is OK as the seekWithinPeriod will always be false skipping the EOS reset check

fragmentcollector_mpd.cpp
DomSyna: compiler ?

fragmentcollector_mpd.cpp
```diff
@@ -2360,15 +2399,23 @@ double StreamAbstractionAAMP_MPD::SkipFragments( MediaStreamContext *pMediaStrea
 				pMediaStreamContext->fragmentTime -= offset;
 			}
 		}
+		int loopCount = 0;
 		do
 		{
+			loopCount++;
```
DomSyna: This looks like purely defensive AI code and I can't see how it serves any purpose for valid content. I think you would need a malformed manifest with  d=0 and a large r value for this to make sense which would be a very wrong.

fragmentcollector_mpd.cpp
```diff
@@ -2484,6 +2531,29 @@ double StreamAbstractionAAMP_MPD::SkipFragments( MediaStreamContext *pMediaStrea
...
+								pMediaStreamContext->type, skipTimeAfterFragment);
+							if (updateFirstPTS && pMediaStreamContext->type == eTRACK_VIDEO)
+							{
+								if ((mFirstPTS == 0) || (firstPTS < mFirstPTS))
+								{
+									AAMPLOG_INFO("[%s] mFirstPTS %f -> %f ", pMediaStreamContext->name, mFirstPTS, firstPTS);
+									mFirstPTS = firstPTS;
+								}
+							}
+							skipTime = 0;
```
DomSyna: Do we not also need to set mIsFinalFirstPTS, and mVideoPosRemainder here ?

fragmentcollector_mpd.cpp
pstroffolino: GetPeriodDuration() can return zero when the current live period duration cannot be determined, although it often derives a nonzero duration from the available segment timeline. With the current comparison, a normal nonnegative live seek yields seekWithinPeriod == false, so EOS is not cleared. This is intentional and conservative because, without a known period boundary, the EOS may be needed to trigger manifest refresh/period handling. I will add an explicit mPeriodDuration > 0 guard to document that intent and avoid relying on the comparison with zero.

